PR #2, opened 2018-03-28
Rename project to EVMC

axic:
Fixes https://github.com/ethereum/evmjit/issues/169.


axic:
Should also rename the file from `evm.h` to `evmc.h`

PR #4, opened 2018-03-28
Fix documentation

chfast:


chfast:
Work but waiting for Docker Hub, and "docs" branch to be removed from whitelist.

include/evmc.h
```diff
@@ -1,4 +1,4 @@
-/// EVM-C -- C interface to Ethereum Virtual Machine
+/// EVMC -- Ethereum Client-VM Connector API
```
axic: I felt like merging #2 was a bit rushed :)

circle.yml
```diff
@@ -14,8 +14,50 @@ jobs:
...
+      - checkout
+      - run:
+          name: "Generate documentation"
+          command: doxygen docs/Doxyfile
+      - run:
+          name: "Upload documentation"
+          command: |
+            git config user.email "docs-bot@ethereum.org"
+            git config user.name "Documentation Bot"
+            git add docs
```
axic: Actually can you move it to the root instead so that https://ethereum.github.io/evmc works.

chfast: I will try.

axic: Both work for me equally.

include/evmc.h
```diff
@@ -273,8 +273,8 @@ struct evmc_result {
 /// @see evmc_get_optional_data(), evmc_get_const_optional_data().
 union evmc_result_optional_data
```
axic: Is this actually still used?!

chfast: Not sure. I will check it.

PR #8, opened 2018-04-09
CMake: Support multiple evmc project instances

chfast:


chfast:
It just avoids declaring evmc library twice.

PR #11, opened 2018-04-10
Add Cable as git subtree

chfast:


axic:
I wonder how will the update work, I guess subtree will find that meta commit and extract information off it, just need to make sure nobody makes changes to the `cable` directory.

chfast:
I'm still not sure it's more than source code copy-paste. I've read you can make changes in cable dir, but make sure the commits modifies only files there. Then you should be able to push these changes up.

PR #12, opened 2018-04-10
VM Tester tool

chfast:
Ready, but some git history cleanup might be needed.

Depends on #11.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,7 @@ target_include_directories(evmc INTERFACE include)
 
 option(EVMC_BUILD_TESTS "Build EVMC tests and test tools" OFF)
 if(EVMC_BUILD_TESTS)
+    enable_testing()
     add_subdirectory(test)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,25 @@
+# EVMC -- Ethereum Client-VM Connector API
+# Copyright 2018 Pawel Bylica.
+# Licensed under the MIT License. See the LICENSE file.
+
 cmake_minimum_required(VERSION 3.5)
 
 if(TARGET evmc)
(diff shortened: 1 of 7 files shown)


axic:
Can hunter and boost be disabled if the project pulling this in only uses the header file?

chfast:
1. Boost and GTest are going to be downloaded only if `EVMC_BUILD_TEST=ON`.
2. Hunter can be additionally disabled with `HUNTER_ENABLED=OFF`.

chfast:
BTW I have to improve command line options handling in vmtester.

Also, how about naming everything with `evmc-` prefix? I.e. `evmc-vmtester` and `evmc-examplevm`?

axic:
> Also, how about naming everything with evmc- prefix? I.e. evmc-vmtester and evmc-examplevm?

Sounds good.

chfast:
Ready?

axic:
> Boost and GTest are going to be downloaded only if EVMC_BUILD_TEST=ON.

I really hope that's true, because with my limited understanding of cmake it seems to show the contrary. The `HunterGate` command seems to be outside that `if`.

axic:
Can you document that in the README for other implementers?

chfast:
How about I make also Hunter to be off by default add install rule for the evmc.h header and document CMake options in another PR?

chfast:
I did a bit more experiments with CMake. `HunterGate` is only a declaration and does not init Huter. Hunter is initialized when used first time, what is only when building with `-DEVMC_BUILD_TESTS=ON`. I added a comment about it.

axic:
> Document what? CMake options?

Yep, how to turn off hunter and tests.

chfast:
> > Document what? CMake options?
> 
> Yep, how to turn off hunter and tests.

They are off by default.

PR #14, opened 2018-04-11
Add clang-format config

chfast:
@axic are you fine with using clang-format config from cpp-ethereum / ethminer?

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,42 @@
+---
+Language:        Cpp
+BasedOnStyle:  Chromium
+AccessModifierOffset: -4
+AlignAfterOpenBracket: Align
+BinPackParameters: false
+BraceWrapping:
+  AfterClass:      true
+  AfterControlStatement: true
+  AfterEnum:       true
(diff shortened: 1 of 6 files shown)


axic:
I think it looks horrible in `cpp-ethereum`, but hopefully it doesn't distort the header too much?

Can you run a reformatting here? I actually very much do not like the partial reformatting done in cpp-ethereum and since this project is tiny, I'd just reformat it at in this PR.

chfast:
What exactly looks horrible?
We can stick to 80 column limit if preferred.

axic:
The way parameters and arguments are laid out, especially that the closing parenthesis is not on a new line :)

axic:
Also need to reformat the example files.

chfast:
> The way parameters and arguments are laid out, especially that the closing parenthesis is not on a new line :)

clang-format does not support this, but you can send a patch to clang-format.

chfast:
Updated.

chfast:
Can I merge it?

PR #15, opened 2018-04-11
EVMC Example VM

chfast:
This refactors evmc-examplevm. On CI examplevm is build as a shared library and tested with vmtester.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,7 @@ if(TARGET evmc)
 endif()
 
 include(cmake/cable/bootstrap.cmake)
+include(CableCompilerSettings)
 include(HunterGate)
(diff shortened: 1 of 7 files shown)


axic:
This PR does a lot of reformatting before merging clang-format? 😉

chfast:
> This PR does a lot of reformatting before merging clang-format? :wink:
Yes. I reformatted examplevm.cpp.

chfast:
Is this a problem?

axic:
I guess there will be no change to the clang-format anyway (😢), but still looks better to have the formatting on its own.

chfast:
We consider some changes, but you have to propose them as a .clang-format change. Not in natural language.

PR #17, opened 2018-04-12
Precompile failure

### LINKED ISSUE #16: New status code `EVM_EC_ERROR`
This test: https://github.com/ethereum/tests/blob/develop/BlockchainTests/GeneralStateTests/stZeroKnowledge2/ecadd_0-0_1-3_21000_128_d0g0v0.json

throws exceptions in KEVM during some elliptic curve math. Currently, I am reporting elliptic curve errors as `EVM_INTERNAL_ERROR`, which does not count as an "exceptional" error but a VM failure. However, the test indicates that it should be treated as an "exceptional" error. See here (search for `EVM_INTERNAL_ERROR`): https://github.com/kframework/evm-semantics/blob/typed-exceptions/evm.md#precompiled-contracts

When running this using `./kevm run tests/ethereum-tests/BlockchainTests/GeneralStateTests/stZeroKnowledge2/ecadd_0-0_1-3_21000_128_d0g0v0.json`, it indicates that it reaches:

```
<statusCode> EVM_INTERNAL_ERROR </statusCode>
<k> #exception ~> #return ... </k>
```

which basically means "if it's an exceptional error, handle it correctly", but because it's being reported as an `EVM_INTERNAL_ERROR`, execution halts and the test now fails.

I propose adding `EVM_EC_ERROR`, which indicates errors in the elliptic curve crypto functions.

ehildenb:
Fixes #16.

Actually addresses two things:

1. adding EVMC_PRECOMPILE_FAILURE
2. renaming EVMC_STATIC_MODE_ERROR

### DIFF
--- a/include/evmc.h
+++ b/include/evmc.h
@@ -151,9 +151,7 @@ enum evmc_status_code {
     EVMC_REVERT = 7,                ///< Execution terminated with REVERT opcode.
 
     /// Tried to execute an operation which is restricted in static mode.
-    ///
-    /// @todo Avoid _ERROR suffix that suggests fatal error.
-    EVMC_STATIC_MODE_ERROR = 8,
+    EVMC_STATIC_MODE_VIOLATION = 8,
 
     /// The dedicated INVALID instruction was hit.
     EVMC_INVALID_INSTRUCTION = 9,
--- a/include/evmc.h
+++ b/include/evmc.h
@@ -163,6 +163,11 @@ enum evmc_status_code {
     /// An example is RETURNDATACOPY reading past the available buffer.
     EVMC_INVALID_MEMORY_ACCESS = 10,
 
+    /// Exceptions produced by precompiles/system contracts
+    ///
+    /// An example: elliptic curve functions handed invalid EC points
+    EVMC_PRECOMPILE_FAILURE = 11,
+
     /// The EVM rejected the execution of the given code or message.
     ///
     /// This error SHOULD be used to signal that the EVM is not able to or


include/evmc.h
```diff
@@ -151,9 +151,7 @@ enum evmc_status_code {
     EVMC_REVERT = 7,                ///< Execution terminated with REVERT opcode.
 
     /// Tried to execute an operation which is restricted in static mode.
-    ///
-    /// @todo Avoid _ERROR suffix that suggests fatal error.
-    EVMC_STATIC_MODE_ERROR = 8,
+    EVMC_STATIC_MODE_VIOLATION = 8,
```
axic: This sounds good!

include/evmc.h
```diff
@@ -163,6 +161,9 @@ enum evmc_status_code {
     /// An example is RETURNDATACOPY reading past the available buffer.
     EVMC_INVALID_MEMORY_ACCESS = 10,
 
+    /// Some error throw by precompiles (eg., elliptic curve handed invalid EC points)
```
chfast: thrown

ehildenb: Should be better now, check again.

include/evmc.h
```diff
@@ -163,6 +161,11 @@ enum evmc_status_code {
     /// An example is RETURNDATACOPY reading past the available buffer.
     EVMC_INVALID_MEMORY_ACCESS = 10,
 
+    /// Exceptions produced by precompiles/system contracts
+    ///
+    /// An example: elliptic curve functions handed invalid EC points
+    EVMC_PRECOMPILE_FAILURE = 11,
```
axic: I'm not sure what will return this, because the engines cannot and the clients don't at the moment.

I have a different idea how to solve this in the long term, but this might be an ok short term solution.

PR #18, opened 2018-04-17
EVMC ABI 1

### LINKED ISSUE #10: Entry point naming convention
It is required that VM implementations have different names for "create" function.
At the moment we recommend naming them as `<vm-name>_create`.
I'd like to change the convention to `evmc_create_<vm-name>`.

This name would be more distinctive and will contain `evmc_` prefix to easier recognize that it is for.
Also, when loading dynamic library I'm considering iterating over all exported symbols searching for `evmc_create_` prefix.

chfast:
1. It introduces "name" and "version" fields in evmc_instance. This should be useful for dynamic loaded VM.
2. The naming convention for "create" function changed to be `evmc_create_` prefixed.
3. ABI version bumped to 1.

Fixes #10.

### DIFF
--- a/examples/examplevm/examplevm.h
+++ b/examples/examplevm/examplevm.h
@@ -1,0 +1,13 @@
+/* EVMC: Ethereum Client-VM Connector API.
+ * Copyright 2018 Pawel Bylica.
+ * Licensed under the MIT License. See the LICENSE file.
+ */
+
+#pragma once
+
+#include <evmc.h>
+
+/**
+ * Creates EVMC Example VM.
+ */
+struct evmc_instance* evmc_create_examplevm(void);
--- a/include/evmc.h
+++ b/include/evmc.h
@@ -31,9 +31,10 @@ extern "C" {
 
 // BEGIN Python CFFI declarations
 
-enum {
+enum
+{
     /// The EVMC ABI version number of the interface declared in this file.
-    EVMC_ABI_VERSION = 0
+    EVMC_ABI_VERSION = 1
 };
 
 /// Big-endian 256-bit integer.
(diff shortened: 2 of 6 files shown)


include/evmc.h
```diff
@@ -549,6 +549,10 @@ struct evmc_instance
     /// represented by this file is in ::EVMC_ABI_VERSION.
     const int abi_version;
 
+    const char* name;
+
+    const char* version;
```
axic: Should have some comments (such as cannot be nullptr or empty).

examples/examplevm/examplevm.h
```diff
@@ -0,0 +1,13 @@
+/* EVMC: Ethereum Client-VM Connector API.
+ * Copyright 2018 Pawel Bylica.
+ * Licensed under the MIT License. See the LICENSE file.
+ */
+
+#pragma once
```
axic: Is this a C or C++ header? :)

chfast: C. I will have to fix comments syntax in evmc.h some day. GCC -pedantic is complaining.

include/evmc.h
chfast: Yes, CI with doxygen has reported this.

examples/examplevm/examplevm.h
axic: For me `#pragma once` is C++-ism.

chfast: Maybe, but none of the compilers complains about it.

axic: I think in pedantic mode GCC would complain, since it is not part of <= C99 IIRC.

gcc and clang are too liberal allowing C++-ism in C by default.

PR #23, opened 2018-04-18
Reorder status codes

chfast:


include/evmc.h
```diff
@@ -140,48 +140,97 @@ typedef void (*evmc_get_block_hash_fn)(struct evmc_uint256be* result,
...
+
+    /** Tried to execute an operation which is restricted in static mode. */
+    EVMC_STATIC_MODE_VIOLATION = 9,
+
+    /**
+     * Tried to read outside memory bounds.
+     *
+     * An example is RETURNDATACOPY reading past the available buffer.
+     */
     EVMC_INVALID_MEMORY_ACCESS = 10,
```
axic: I think this should come before `static mode`.

include/evmc.h
```diff
@@ -140,48 +140,97 @@ typedef void (*evmc_get_block_hash_fn)(struct evmc_uint256be* result,
...
-    /// Tried to read outside memory bounds.
-    ///
-    /// An example is RETURNDATACOPY reading past the available buffer.
+/**
+ * The execution status code.
+ *
+ * Successful execution is represented by ::EVMC_SUCCESS having value 0.
+ * Positive values represent failures defined by VM specifications with generic
+ * ::EVMC_FAILURE code of value 1.
+ * Negative values represent VM internal errors and might be not recoverable.
```
axic: Should mention the special `REJECTED` here I think.

include/evmc.h
```diff
@@ -140,48 +140,97 @@ typedef void (*evmc_get_block_hash_fn)(struct evmc_uint256be* result,
...
+     * by the EVM implementation.
+     *
+     * This error SHOULD be used to signal that the EVM is not able to or
+     * willing to execute the given code type or message.
+     * If an EVM returns the ::EVMC_REJECTED status code,
+     * the Client MAY try to execute it in other EVM implementation.
+     * For example, the Client tries running a code in the EVM 1.5. If the
+     * code is not supported there, the execution falls back to the EVM 1.0.
+     */
+    EVMC_REJECTED = -2,
```
axic: What is the reason for swapping these two?

include/evmc.h
chfast: Why?

include/evmc.h
chfast: The generic first, the specific after.

include/evmc.h
axic: I think a logical order would be:
- invalid/unknown instructions
- invalid number of arguments for instructions
- invalid arguments (bad jump, invalid memory access)
- invalid state (static mode violation)

include/evmc.h
axic: I a bit torn on this, since `REJECTED` is a special one and I am not sure it is an internal error. It is certainly recoverable.

chfast: It's internal in sense it should not be passed back to the caller.
Yes, it's recoverable if the client supports it and has a backup VM. I will extend the description of `evmc_status_code` to mention this as suggested.

include/evmc.h
chfast: Mentioned.

include/evmc.h
axic: No comment on this nice ordered list? 😉

include/evmc.h
```diff
@@ -140,51 +140,105 @@ typedef void (*evmc_get_block_hash_fn)(struct evmc_uint256be* result,
...
+ * Positive values represent failures defined by VM specifications with generic
+ * ::EVMC_FAILURE code of value 1.
+ *
+ * Status codes with negative values represent VM internal errors
+ * not provided by EVM specifications. These errors MUST not be passed back
+ * to the caller. They MAY be handled by the Client in predefined manner
+ * (see e.g. ::EVMC_REJECTED), otherwise internal errors are not recoverable.
+ * The generic representant of errors is ::EVMC_INTERNAL_ERROR but
+ * an EVM implementation MAY return negative status codes that are not defined
+ * in the EVMC documentation.
```
axic: Should probably leave a comment here along the lines of "should you need more error codes, please create an issue on the evmc repo to discuss" to facilitate discussion.

include/evmc.h
chfast: I did it this way...

include/evmc.h
chfast: Added.

ehildenb:
I might be good to start a convention along the lines of:

-   `0*`: normal statuses (`EVMC_SUCCESS`, `EVMC_REVERT`, etc ...)
-   `1*`: exceptional statuses (oog, oom, etc...)
-   `2*`: validity statuses (bad jump structure, etc ...)
-   `3*`: error modes (internal error, etc ...)

this way we don't have to re-organize the entire error-code space every time we add codes. Perhaps we make them hexadecimal numbers to allow for at least 16 codes in each category.

Notice that the KEVM formalization of this: https://github.com/kframework/evm-semantics/blob/typed-exceptions/network.md
does not include numbers, it simply bins them by sort. So I have a sort `ExceptionalStatusCode` for execeptions, and `EndStatusCode` for ones that end execution, then just "the rest". Also note that I have `EVMC_BALANCE_UNDERFLOW` and `EVMC_ACCOUNT_ALREADY_EXISTS` as status codes which are not present here (and are used in KEVM).

chfast:
I don't mind categories, but I'm capable of doing good clusterization myself. If that will follow external reference I'm ok if having this here.

Personally, I believe it will just look nice here, with not benefits for implementations.

I'd like to keep internal errors with negative values. This way you can specify that `status_code >= 0` must be passed to the caller and `status_code < 0` must not be passed to the caller.

I would also avoid hex, because the number would like different in this header (e.g. `0x11`) and different in logs (e.g. `Failure code 17`) so you cannot quickly look them up here.
If we need more space than 10, use 100.

What do you think @axic?

axic:
> I would also avoid hex, because the number would like different in this header (e.g. 0x11) and different in logs (e.g. Failure code 17) so you cannot quickly look them up here.

This seems to be a reasonable idea.

I wouldn't bother too much about this, I think 0.2.0 will come soon after all these different implementers experience the first problems and we can fix it there.

include/evmc.h
axic: I don't think so :)

chfast:
```
EVMC_INVALID_MEMORY_ACCESS = 9,
EVMC_CALL_DEPTH_EXCEDED = 10,
EVMC_STATIC_MODE_VIOLATION = 11,
```

What am I missing?

chfast:
> I wouldn't bother too much about this, I think 0.2.0 will come soon after all these different implementers experience the first problems and we can fix it there.

I'm a bit inclined not to overdesign it now.

include/evmc/evmc.h
```diff
@@ -140,55 +140,116 @@ typedef void (*evmc_get_block_hash_fn)(struct evmc_uint256be* result,
...
+
+    /**
+     * Tried to read outside memory bounds.
+     *
+     * An example is RETURNDATACOPY reading past the available buffer.
+     */
+    EVMC_INVALID_MEMORY_ACCESS = 9,
+
+    /** Call depth has exceeded the limit (if any) */
+    EVMC_CALL_DEPTH_EXCEDED = 10,
```
axic: This still has the typo (`exceded` vs. `exceeded`).

chfast: Oops, I only fixed the comment.

include/evmc.h
axic: This part:
- invalid/unknown instructions
- invalid number of arguments for instructions
- invalid arguments (bad jump)

The order now is argument count, bad jump, invalid/unknown, etc.

chfast: Reordered again then.

include/evmc/evmc.h
```diff
@@ -140,55 +140,112 @@ typedef void (*evmc_get_block_hash_fn)(struct evmc_uint256be* result,
...
+     * in case this INVALID instruction has been encountered.
+     */
+    EVMC_INVALID_INSTRUCTION = 4,
+
+    /** An undefined instruction has been encountered. */
+    EVMC_UNDEFINED_INSTRUCTION = 5,
+
+    /**
+     * The execution has attempted to put more items on the EVM stack
+     * than the specified limit.
```
tcsiwula: Was the EVM stack limit 1024? If it is not going to change in the future, then it may be of use to include that number here 😁 @chfast

include/evmc/evmc.h
```diff
@@ -140,55 +140,112 @@ typedef void (*evmc_get_block_hash_fn)(struct evmc_uint256be* result,
...
+    EVMC_BAD_JUMP_DESTINATION = 8,
+
+    /**
+     * Tried to read outside memory bounds.
+     *
+     * An example is RETURNDATACOPY reading past the available buffer.
+     */
+    EVMC_INVALID_MEMORY_ACCESS = 9,
+
+    /** Call depth has exceeded the limit (if any) */
```
tcsiwula: Was the call depth limit, also 1024?

include/evmc/evmc.h
chfast: Both limits for stack and call depth are 1024 and have never changed.

include/evmc/evmc.h
```diff
@@ -140,55 +140,112 @@ typedef void (*evmc_get_block_hash_fn)(struct evmc_uint256be* result,
...
+     * by the EVM implementation.
+     *
+     * This error SHOULD be used to signal that the EVM is not able to or
+     * willing to execute the given code type or message.
+     * If an EVM returns the ::EVMC_REJECTED status code,
+     * the Client MAY try to execute it in other EVM implementation.
+     * For example, the Client tries running a code in the EVM 1.5. If the
+     * code is not supported there, the execution falls back to the EVM 1.0.
+     */
+    EVMC_REJECTED = -2,
```
tcsiwula: In my opinion (for what it is worth) a more descriptive name would be `EVMC_INCOMPATIBILE_VERSION` or `EVMC_INVALID_VERSION`

chfast: The provided example is not the only case we are using it even now. You can configure EVMJIT to jit only popular codes (measured by hit count). It will return EVMC_REJECTED if the hit count for requested code is below the threshold.

include/evmc/evmc.h
axic: It may make sense including in the description but not in the error name. We can consider it when a hardfork is introduced changing those :)

chfast: I will leave it for another time...

PR #25, opened 2018-04-19
CMake install

chfast:
Adds install rule for evmc.h header.
Creates an archive package (see Circle CI artifacts).


CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ HunterGate(
 )
 
 project(evmc)
+set(PROJECT_VERSION "0.1.0.dev0")
```
axic: What is `dev0`? Isn't `alpha0` more common?

chfast: It's "in development" version. `.dev < alpha (a) < beta (b) < rc < ""`.

chfast: So I'm not sure about versioning syntax, but when I don't I use PEP 440 .

PR #26, opened 2018-04-19
Add EVMC_CONTRACT_VALIDATION_FAILED error

axic:


include/evmc/evmc.h
```diff
@@ -170,6 +170,10 @@ enum evmc_status_code {
     /// Call depth exceded (if there is a call depth limit)
     EVMC_CALL_DEPTH_EXCEDED = 12,
 
+    /// Contract validation failed (such as EVM 1.5 jump validity,
+    /// Casper's purity checker or ewasm contract rules).
+    EVMC_CONTRACT_VALIDATION_FAILED = 13,
```
axic: Not sure about the name. Is `INVALID_CONTRACT` better?

chfast: In long form it should be changed to `EVMC_CONTRACT_VALIDATION_FAILURE`.

For me "invalid contract" or "contract invalid" is also good enough, but it slightly suggests the invalid contract exists in the wild, not that the new one was rejected.

axic: Updated. Can we merge this before the reorder? I'll rebase that.

Issue #19, opened 2018-04-17
Add compilation tests for C90 C99 and C11


chfast:
See example compilation tests in [ethash](https://github.com/chfast/ethash): https://github.com/chfast/ethash/commit/9551b5973280df7b18fb8798a1bd014dc7557b46.

axic:
Also: https://github.com/ethereum/evmc/pull/18#discussion_r182003696

PR #30, opened 2018-04-24
Make header compilable in C99 and C11 pedantic mode

chfast:
Resolves https://github.com/ethereum/evmc/issues/19.

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -1,141 +1,177 @@
-/// EVMC -- Ethereum Client-VM Connector API
-///
-/// ## High level design rules
-///
-/// 1. Pass function arguments and results by value.
-///    This rule comes from modern C++ and tries to avoid costly alias analysis
-///    needed for optimization. As the result we have a lots of complex structs
-///    and unions. And variable sized arrays of bytes cannot be passed by copy.
-/// 2. The EVM operates on integers so it prefers values to be host-endian.
-///    On the other hand, LLVM can generate good code for byte swaping.
-///    The interface also tries to match host application "natural" endianess.
-///    I would like to know what endianess you use and where.
-///
-/// ## Terms
-///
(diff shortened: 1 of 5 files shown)


include/evmc/evmc.h
```diff
@@ -245,411 +281,469 @@ enum evmc_status_code
...
+ *  @param code_size   The length of the bytecode.
+ *  @return            All execution results.
+ */
 typedef struct evmc_result (*evmc_execute_fn)(struct evmc_instance* instance,
-                                             struct evmc_context* context,
-                                             enum evmc_revision rev,
-                                             const struct evmc_message* msg,
-                                             uint8_t const* code,
-                                             size_t code_size);
+                                              struct evmc_context* context,
```
chfast: I kept this param packing. How do you like it? I think it's useful for documentation purposes.

axic:
It's really hard to see the changes. Is it only C++-style to C-style comments?

chfast:
Only comments, some `,` removed. The file is mostly comments.

include/evmc/evmc.h
```diff
@@ -1,141 +1,177 @@
...
-/// 2. Host -- an entity controlling the EVM. The Host requests code execution
-///            and responses to EVM queries by callback functions.
-///
-/// @defgroup EVMC EVMC
-/// @{
+/**
+ * EVMC: Ethereum Client-VM Connector API
+ *
+ * @copyright
+ * Copyright 2018 Pawel Bylica.
```
axic: Btw, if there's a copyright here then I guess all the authors should be listed.

Probably way better is to have an `AUTHORS` file to list everyone, while the header can list the major contributors.

chfast: I share the same opinion, but getting a clarification would be nice. See https://github.com/ethereum/evmc/pull/1.

PR #32, opened 2018-05-11
Tracing API

### LINKED ISSUE #22: Support VM tracing
Original thread: https://github.com/ethereum/evmjit/issues/85

chfast:
Fixes #22.

### DIFF
--- a/examples/examplevm/examplevm.c
+++ b/examples/examplevm/examplevm.c
@@ -9,6 +9,8 @@ struct examplevm
 {
     struct evmc_instance instance;
     int verbose;
+    evmc_trace_callback trace_callback;
+    struct evmc_tracer_context* tracer_context;
 };
 
 static void destroy(struct evmc_instance* evm)
@@ -113,6 +115,15 @@ static struct evmc_result execute(struct evmc_instance* instance,
     return ret;
 }
 
+static void set_tracer(struct evmc_instance* instance,
+                       evmc_trace_callback callback,
+                       struct evmc_tracer_context* context)
+{
+    struct examplevm* vm = (struct examplevm*)instance;
+    vm->trace_callback = callback;
+    vm->tracer_context = context;
+}
+
 struct evmc_instance* evmc_create_examplevm()
 {
     struct evmc_instance init = {
@@ -122,6 +133,7 @@ struct evmc_instance* evmc_create_examplevm()
         .destroy = destroy,
         .execute = execute,
         .set_option = set_option,
+        .set_tracer = set_tracer,
     };
     struct examplevm* vm = calloc(1, sizeof(struct examplevm));
     struct evmc_instance* interface = &vm->instance;
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -718,12 +718,22 @@ struct evmc_tracer_context;
  *                               execution. This value is null when the instruction does not push
  *                               anything to the stack.
  * @param memory_size            The size of the EVM memory after the instruction execution.
- * @param changed_memory_offset  The beginning of the memory area modified as the result of
- *                               the instruction execution.
+ * @param changed_memory_offset  The offset in number of bytes of the beginning of the memory area
+ *                               modified as the result of the instruction execution.
+ *                               The Client MAY use this information together with
+ *                               @p changed_memory_size and @p changed_memory to incrementally
+ *                               update the copy of the full VM's memory.
  * @param changed_memory_size    The size of the memory area modified as the result of
  *                               the instruction execution.
  * @param changed_memory         The pointer to the memory area modified as the result of
  *                               the instruction execution.
+ *                               The Client MAY access the pointed memory area
+ *                               (limited by the @p changed_memory_size) only during the current
+ *                               execution of the evmc_trace_callback().
+ *                               The pointer MUST NOT be stored by the Client.
+ *                               The Client MUST NOT assume that
+ *                               `changed_memory - changed_memory_offset` is a valid base pointer
+ *                               of the VM memory.
  */
 typedef void (*evmc_trace_callback)(struct evmc_tracer_context* context,
                                     size_t code_offset,
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -675,18 +675,18 @@ enum evmc_revision
 
 
 /**
- * Generates and executes machine code for given EVM bytecode.
+ * Executes the given EVM bytecode using the input in the message
  *
- *  All the fun is here. This function actually does something useful.
+ * This function MAY be invoked multiple times for a single EVM instance.
  *
- *  @param instance    A EVM instance.
- *  @param context     The pointer to the Host execution context to be passed
- *                     to callback functions. @see ::evmc_context.
- *  @param rev         Requested EVM specification revision.
- *  @param msg         Call parameters. @see ::evmc_message.
- *  @param code        Reference to the bytecode to be executed.
- *  @param code_size   The length of the bytecode.
- *  @return            All execution results.
+ * @param instance   The EVM instance.
+ * @param context    The pointer to the Client execution context to be passed
+ *                   to the callback functions. @see ::evmc_context.
+ * @param rev        Requested EVM specification revision.
+ * @param msg        Call parameters. @see ::evmc_message.
+ * @param code       Reference to the bytecode to be executed.
+ * @param code_size  The length of the bytecode.
+ * @return           All execution results.
  */
 typedef struct evmc_result (*evmc_execute_fn)(struct evmc_instance* instance,
                                               struct evmc_context* context,
@@ -703,22 +703,24 @@ struct evmc_tracer_context;
  * The callback to trace instructions execution in an EVM.
  *
  * This function informs the Client what instruction has been executed in the EVM implementation
- * and what are the results of the execution this particular instruction.
+ * and what are the results of executing this particular instruction.
  * The message level information (like call depth, destination address, etc.) are not provided here.
  * This piece of information can be acquired by inspecting messages being sent to the EVM in
  * ::evmc_execute_fn and the results of the messages execution.
  *
  * @param context                The pointer to the Client-side tracing context. This allows to
  *                               implement the tracer in OOP manner.
- * @param step                   The instruction counter for the current message.
- * @param code_offset            The instruction possition in the code.
+ * @param step                   The instruction counter: number of instructions executed.
+ *                               This counter starts from 0 for every message passed to
+ *                               ::evmc_execute_fn.
+ * @param code_offset            The current instruction position in the code.
  * @param status_code            The status code of the instruction execution.
  * @param gas_left               The amount of the gas left after the instruction execution.
- * @param stack_num_items        The current EVM stack height.
+ * @param stack_num_items        The current EVM stack height after the instruction execution.
  * @param pushed_stack_item      The top EVM stack item pushed as the result of the instruction
  *                               execution. This value is null when the instruction does not push
  *                               anything to the stack.
- * @param memory_size            The current size of the EVM memory.
+ * @param memory_size            The size of the EVM memory after the instruction execution.
  * @param changed_memory_offset  The beginning of the memory area modified as the result of
  *                               the instruction execution.
  * @param changed_memory_size    The size of the memory area modified as the result of
@@ -741,13 +743,18 @@ typedef void (*evmc_trace_callback)(struct evmc_tracer_context* context,
 /**
  * Sets the EVM instruction tracer.
  *
- * When the tracer is set in the EVM instance, the EVM SHOULD callback the tracer with information
+ * When the tracer is set in the EVM instance, the EVM SHOULD call back the tracer with information
  * about instructions execution in the EVM.
  * @see ::evmc_trace_callback.
  *
+ * This will overwrite the previous settings (the callback and the context).
+ *
  * @param instance    The EVM instance.
- * @param callback    The tracer callback function.
- * @param context     The Client-side tracer context.
+ * @param callback    The tracer callback function. This argument MAY be NULL to disable previously
+ *                    set tracer.
+ * @param context     The Client-side tracer context. This argument MAY be NULL in case the tracer
+ *                    does not require any context. This argument MUST be NULL if the callback
+ *                    argument is NULL.
  */
 typedef void (*evmc_set_tracer_fn)(struct evmc_instance* instance,
                                    evmc_trace_callback callback,
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -710,9 +710,6 @@ struct evmc_tracer_context;
  *
  * @param context                The pointer to the Client-side tracing context. This allows to
  *                               implement the tracer in OOP manner.
- * @param step                   The instruction counter: number of instructions executed.
- *                               This counter starts from 0 for every message passed to
- *                               ::evmc_execute_fn.
  * @param code_offset            The current instruction position in the code.
  * @param status_code            The status code of the instruction execution.
  * @param gas_left               The amount of the gas left after the instruction execution.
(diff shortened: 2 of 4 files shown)


include/evmc/evmc.h
```diff
@@ -685,6 +685,21 @@ typedef struct evmc_result (*evmc_execute_fn)(struct evmc_instance* instance,
                                               size_t code_size);
 
 
+struct evmc_tracer_context;
+
+typedef void (*evmc_trace_callback)(struct evmc_tracer_context* context,
+                                    int depth,
+                                    int step,
+                                    size_t code_offset,
```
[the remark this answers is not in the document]
chfast: The problem with "pc" name I have is not everyone knows exactly that it is exactly. I think `code_offest` === `pc`.

chfast:
> Has a serious lack of comments :)

Intentionally. I expect this to change a lot in the first phase. And I want this to be clear enough without comments.

include/evmc/evmc.h
```diff
@@ -685,6 +685,26 @@ typedef struct evmc_result (*evmc_execute_fn)(struct evmc_instance* instance,
...
+                                    enum evmc_status_code status_code,
+                                    int64_t gas_left,
+                                    size_t stack_num_items,
+                                    const struct evmc_uint256be* pushed_stack_item,
+                                    size_t memory_size,
+                                    size_t changed_memory_offset,
+                                    size_t changed_memory_size,
+                                    const uint8_t* changed_memory);
+
+typedef void (*evmc_set_tracer)(struct evmc_instance* instance,
```
gumb0: Shouldn't this return a pointer to previously set callback?
Why do you need to declare a pointer to function (`evmc_set_tracer`) if it's just to declare one function?

chfast: Yes, it can return previous callback, but it would be hard to also return the context, so you will not be able to restore previous state anyway.

(If I understand the question) The evmc_instance is a set of function pointers. It is easier to read the spec when function pointer types are declared separately. I forgot the `_fn` suffix in this type name.

gumb0: Ah ok, I didn't notice that it's a part of `evmc_instance`, sorry

include/evmc/evmc.h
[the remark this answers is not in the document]
axic: So what is `code_offset` and what is `step` ?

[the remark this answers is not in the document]
chfast: The `step` is the number of executed instructions so far (only increases).
The `code_offset` is the position in the code where we are currently. It's different than the `step` because of jump and push instructions.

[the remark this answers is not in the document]
axic: `execution_step` perhaps? Not sure, `step`, without any documentation gave me the wrong indication about what it is.

Why don't you want to have a documentation above it?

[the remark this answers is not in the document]
chfast: We can also remove this because this can be implemented easily on the client side.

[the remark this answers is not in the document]
chfast: I will add documentation soon. I just need to finish other PRs first.

chfast:
The PoC is here: https://github.com/ethereum/cpp-ethereum/pull/5065

I believe the API is good enough to be merged as is. That would allow me to focus on geth support for EVMC.

include/evmc/evmc.h
```diff
@@ -687,6 +687,64 @@ typedef struct evmc_result (*evmc_execute_fn)(struct evmc_instance* instance,
...
+ * This function informs the Client what instruction has been executed in the EVM implementation
+ * and what are the results of the execution this particular instruction.
+ * The message level information (like call depth, destination address, etc.) are not provided here.
+ * This piece of information can be acquired by inspecting messages being sent to the EVM in
+ * ::evmc_execute_fn and the results of the messages execution.
+ *
+ * @param context                The pointer to the Client-side tracing context. This allows to
+ *                               implement the tracer in OOP manner.
+ * @param step                   The instruction counter for the current message.
+ * @param code_offset            The instruction possition in the code.
```
gumb0: typo: position

include/evmc/evmc.h
```diff
@@ -687,6 +687,64 @@ typedef struct evmc_result (*evmc_execute_fn)(struct evmc_instance* instance,
...
+ *
+ * This function informs the Client what instruction has been executed in the EVM implementation
+ * and what are the results of the execution this particular instruction.
+ * The message level information (like call depth, destination address, etc.) are not provided here.
+ * This piece of information can be acquired by inspecting messages being sent to the EVM in
+ * ::evmc_execute_fn and the results of the messages execution.
+ *
+ * @param context                The pointer to the Client-side tracing context. This allows to
+ *                               implement the tracer in OOP manner.
+ * @param step                   The instruction counter for the current message.
```
gumb0: Maybe to clarify it instead of "current message" call it " The instruction counter after the instruction execution."
And also below add "after the instruction execution" for code_offset, stack_num_items, memory_size

chfast:
I changed the documentation.

include/evmc/evmc.h
```diff
@@ -687,6 +687,66 @@ typedef struct evmc_result (*evmc_execute_fn)(struct evmc_instance* instance,
...
 
 
+/** The opaque type representing a Client-side tracer object. */
+struct evmc_tracer_context;
+
+/**
+ * The callback to trace instructions execution in an EVM.
+ *
+ * This function informs the Client what instruction has been executed in the EVM implementation
+ * and what are the results of the execution this particular instruction.
```
axic: `execution of this` or `of executing this`

include/evmc/evmc.h
```diff
@@ -687,6 +687,66 @@ typedef struct evmc_result (*evmc_execute_fn)(struct evmc_instance* instance,
...
+ * @param context                The pointer to the Client-side tracing context. This allows to
+ *                               implement the tracer in OOP manner.
+ * @param step                   The instruction counter: number of instructions executed.
+ *                               This counter starts from 0 for every message passed to
+ *                               ::evmc_execute_fn.
+ * @param code_offset            The current instruction position in the code.
+ * @param status_code            The status code of the instruction execution.
+ * @param gas_left               The amount of the gas left after the instruction execution.
+ * @param stack_num_items        The current EVM stack height after the instruction execution.
+ * @param pushed_stack_item      The top EVM stack item pushed as the result of the instruction
```
axic: So based on these two the client needs to apply the stack changes?

include/evmc/evmc.h
```diff
@@ -687,6 +687,66 @@ typedef struct evmc_result (*evmc_execute_fn)(struct evmc_instance* instance,
...
+ *                               execution. This value is null when the instruction does not push
+ *                               anything to the stack.
+ * @param memory_size            The size of the EVM memory after the instruction execution.
+ * @param changed_memory_offset  The beginning of the memory area modified as the result of
+ *                               the instruction execution.
+ * @param changed_memory_size    The size of the memory area modified as the result of
+ *                               the instruction execution.
+ * @param changed_memory         The pointer to the memory area modified as the result of
```
axic: Why is this needed if one can calculate that via `changed_memory_offset`?

include/evmc/evmc.h
```diff
@@ -687,6 +687,66 @@ typedef struct evmc_result (*evmc_execute_fn)(struct evmc_instance* instance,
...
+                                    const struct evmc_uint256be* pushed_stack_item,
+                                    size_t memory_size,
+                                    size_t changed_memory_offset,
+                                    size_t changed_memory_size,
+                                    const uint8_t* changed_memory);
+
+/**
+ * Sets the EVM instruction tracer.
+ *
+ * When the tracer is set in the EVM instance, the EVM SHOULD callback the tracer with information
```
axic: `callback` -> `call` or `call back`

include/evmc/evmc.h
```diff
@@ -722,6 +782,13 @@ struct evmc_instance
     /** Pointer to function executing a code by the EVM instance. */
     evmc_execute_fn execute;
 
+    /**
+     * Optional pointer to function setting the EVM instruction tracer.
+     *
+     * If the EVM does not support this feature the pointer can be NULL.
+     */
+    evmc_set_tracer_fn set_tracer;
```
axic: Can this be unset after its being set?

I don't fully remember (and seemingly our documentation isn't clear enough): can an `evmc_instance` reused for multiple executions?

chfast: Good question. Answered in the latest commit.

chfast:
I pushed additional commit with improved documentation. I will squeeze them later (easier to review changes now).

include/evmc/evmc.h
axic: That commit looks good.

include/evmc/evmc.h
axic: Any comments?

include/evmc/evmc.h
axic: Any comments?

chfast: You cannot calculate it via `changed_memory_offset` without having the base memory pointer. We can return the base memory pointer instead of the pointer to the changed area. However that would put higher requirements on EVM memory layout.

include/evmc/evmc.h
chfast: If the tracer want to know what's on stack it has to track the changes. This is even more complicated when you consider SWAP and DUP instructions - the tracer would have to reapply SWAP and DUP semantics.

include/evmc/evmc.h
```diff
@@ -687,6 +687,71 @@ typedef struct evmc_result (*evmc_execute_fn)(struct evmc_instance* instance,
...
+ * This piece of information can be acquired by inspecting messages being sent to the EVM in
+ * ::evmc_execute_fn and the results of the messages execution.
+ *
+ * @param context                The pointer to the Client-side tracing context. This allows to
+ *                               implement the tracer in OOP manner.
+ * @param step                   The instruction counter: number of instructions executed.
+ *                               This counter starts from 0 for every message passed to
+ *                               ::evmc_execute_fn.
+ * @param code_offset            The current instruction position in the code.
+ * @param status_code            The status code of the instruction execution.
```
chfast: I'm not sure this is useful (99% of the time the value is 0). And the Client can get it from evm_result in the end.

include/evmc/evmc.h
gumb0: Are we sure there aren't ever going to be instructions pushing more than one item?

chfast: In theory we have them now in form of DUP and SWAP.
In practice, not likely needed. We can also change the API in future. I will check WASM instructions.

include/evmc/evmc.h
```diff
@@ -687,6 +687,71 @@ typedef struct evmc_result (*evmc_execute_fn)(struct evmc_instance* instance,
...
+ * and what are the results of executing this particular instruction.
+ * The message level information (like call depth, destination address, etc.) are not provided here.
+ * This piece of information can be acquired by inspecting messages being sent to the EVM in
+ * ::evmc_execute_fn and the results of the messages execution.
+ *
+ * @param context                The pointer to the Client-side tracing context. This allows to
+ *                               implement the tracer in OOP manner.
+ * @param step                   The instruction counter: number of instructions executed.
+ *                               This counter starts from 0 for every message passed to
+ *                               ::evmc_execute_fn.
```
ehildenb: Why call this `step` instead of the more conventional `programCounter`? Is this to facilitate ewasm?

include/evmc/evmc.h
ehildenb: I think it's probably best (esp for other implementations) to directly pass back the changed memory. Not all language give you access to memory via pointers, so having an explicit argument present here will make it easier to use as a guide for implementing traces in other languages.

My question is how will you handle it if there are multiple disjoint changed regions of memory? What if they are very spaced out (as is often the case when using the "hash of variable name" scheme for deciding where to place things)?

include/evmc/evmc.h
gumb0: See https://github.com/ethereum/evmc/pull/32#discussion_r188351911

include/evmc/evmc.h
chfast: > directly pass back the changed memory

I'm not sure what you exactly mean by this.

> My question is how will you handle it if there are multiple disjoint changed regions of memory?

You don't. There should be a way to report without information about memory. Something like `changed_memory_size == -1`. I'm leaving it unanswered because I don't have good solution at the moment. I prefer do defer the decision until we have the use case for it.

include/evmc/evmc.h
chfast: This is not the program counter. Program counter points the instruction being executed, so it's `code_offset` in this case. The `step` is the trace entry index. 

I'm ok with renaming any of those if you have better ideas for names.

chfast:
@axic @ehildenb Whould you like to re-review this again (including review comments)?

axic:
Rebased.

include/evmc/evmc.h
axic: How about `tracing_step` or `trace_index` ? Why is it needed?

examples/examplevm/examplevm.c
```diff
@@ -9,6 +9,8 @@ struct examplevm
 {
     struct evmc_instance instance;
     int verbose;
+    evmc_trace_callback trace_callback;
```
jwasinger: it would be useful to have another callback here to allow the EVM to log text back to the client.

jwasinger: instead of a separate callback, actually it could just be added as an additional parameter to `evmc_trace_callback`.

include/evmc/evmc.h
chfast: Precisely I wanted the "instruction counter" per current message (i.e. EVM call). This is not not needed (because the Host can count tracing callback calls on its own) *UNLESS* the VM will not report every instruction. 
I propose to rename it to `instruction_counter` or remove it for now.

examples/examplevm/examplevm.c
chfast: The C++ `std::cerr` works. We could add a logger method to capture VM logs nicely on the Host side. But that is different feature and should be done independently.

include/evmc/evmc.h
axic: I think that's why `trace_counter`/`trace_step`/`trace_index` is the appropriate name.

I'm not sure this is a useful field though.

chfast: I mean this case where VM only traces storage instructions:
```
PUSH 0
SLOAD
PUSH 9
SSTORE
```

When counting the number of traces you should get:
```
(0, SLOAD, ...)
(1, SSTORE, ...)
```

When counting number of instructions you should get:
```
(1, SLOAD, ...)
(3, SSTORE, ...)
```

In the second case the Host can figure out that some executed instructions are missing in the trace. In the first case the counter is redundant as the Host can count it itself.

I'm going to remove the param.

axic: > I mean this case where VM only traces storage instructions:

You mean a specific VM won't support tracing every instruction?

I have no problem keeping it, but `step` seems to be rather confusing. Consider some of the suggestions above.

include/evmc/evmc.h
```diff
@@ -696,6 +696,77 @@ typedef struct evmc_result (*evmc_execute_fn)(struct evmc_instance* instance,
...
+ *                               the instruction execution.
+ * @param changed_memory         The pointer to the memory area modified as the result of
+ *                               the instruction execution.
+ *                               The Client MAY access the pointed memory area
+ *                               (limited by the @p changed_memory_size) only during the current
+ *                               execution of the evmc_trace_callback().
+ *                               The pointer MUST NOT be stored by the Client.
+ *                               The Client MUST NOT assume that
+ *                               `changed_memory - changed_memory_offset` is a valid base pointer
+ *                               of the VM memory.
```
axic: I'm glad we discussed these on Gitter, much more cleaner now.

PR #33, opened 2018-06-07
EVM instructions table

chfast:


lib/instructions/instructions.c
```diff
@@ -0,0 +1,1602 @@
...
+#define ZERO 0
+#define BASE 2
+#define VERYLOW 3
+#define LOW 5
+#define MID 8
+#define HIGH 10
+
+/** @} */
+
+static struct evmc_instruction_metrics constantinople_metrics[256] = {
```
gumb0: This file is 1600 lines, maybe put revisions in separate files? It would be also easier to diff them

include/evmc/instructions.h
```diff
@@ -0,0 +1,209 @@
...
+    enum evmc_revision revision);
+
+/**
+ * Get the table of the EVM 1 instruction names.
+ *
+ * This table is EVM revision independent and contains the superset of the names of the instructions
+ * from all EVM revisions. Use evmc_get_instruction_metrics_table() to know if an instruction
+ * is present in the given EVM revision.
+ *
+ * @return  The pointer to the array of 256 instruction names.
```
gumb0: Maybe a note that it contains some NULLs besides names?

chfast: Added.

lib/instructions/instructions.c
chfast: After thinking about this for 2 day, I'd like to keep it as it is. I hope this is single time job and it should depend on testing, not manual review (some parts already tested, I can add test cases for all HFs).

1. To add a new EVM revision you will have to copy the recent one and apply changes. This should not be very complicated. In case it was in separate files, you will also have to alter the CMake config.

2. Having all this in 2 files make it easier to copy to other projects, even translate to other languages.

3. Having the tables in separate files require to make the symbol default visibility.

include/evmc/instructions.h
```diff
@@ -0,0 +1,213 @@
...
+ *
+ * This table is EVM revision independent and contains the superset of the names of the instructions
+ * from all EVM revisions. Use evmc_get_instruction_metrics_table() to know if an instruction
+ * is present in the given EVM revision.
+ *
+ * The entries for undefined instructions contain null pointers.
+ *
+ * @return  The pointer to the array of 256 instruction names.
+ */
+const char* const* evmc_get_instruction_name_table();
```
axic: Shouldn't this also have `revision` parameter?

chfast: No, it has names for all opcodes, i.e. from the latest revision. I think it's good enough for debugging.

include/evmc/instructions.h
```diff
@@ -0,0 +1,213 @@
...
+ */
+const struct evmc_instruction_metrics* evmc_get_instruction_metrics_table(
+    enum evmc_revision revision);
+
+/**
+ * Get the table of the EVM 1 instruction names.
+ *
+ * This table is EVM revision independent and contains the superset of the names of the instructions
+ * from all EVM revisions. Use evmc_get_instruction_metrics_table() to know if an instruction
+ * is present in the given EVM revision.
```
axic: I thought this was the reason for the decision, but it seems a bit convoluted.

chfast:
Because of the name clash on Windows, I have to rename the enum items.

Which one to use?

- `EVMC_OP_ADDMOD`
- `EVM_ADDMOD`
- `OP_ADDMOD`

axic:
Basically it can clash with macros, correct?

chfast:
Is the "opcode" a more precise name to "instruction"? Everybody seems to be using "instruction" name in Ethereum.

chfast:
> Basically it can clash with macros, correct?

It clashes with `BYTE` typedef somewhere in Windows headers.

axic:
I think the "instruction" term stands for the abstract instruction, while "opcode" term stands for the encoding of it as bytes. Since EVM is rather simple, there's not much going on in terms of encoding :)

axic:
Based on that I think opcode would be the more relevant term here, since you assign the encoded value to it.

include/evmc/instructions.h
axic: Would a good middle ground be a helper which based on the revision generates a table relevant to that revision? It would do that every time, so the callee would need to ensure it stores the result.

chfast: No. I will just add the revision here as well if there is the use case for it.

axic: I do like the idea of having this static const table too.

PR #34, opened 2018-06-11
EVM instructions part 2

chfast:
Changes:

- `evmc_instruction` enum renamed to `evmc_opcode` and items prefixed with `OP_`.
- names tables now also depend on EVM revision argument.

Follow up of #33.


lib/instructions/instruction_names.c
```diff
@@ -261,10 +261,800 @@ static const char* names[256] = {
...
+    NULL,
+    NULL,
+    NULL,
+    NULL,
+    NULL,
+    NULL,
+    NULL,
+    NULL,
+    "INVALID",
+    "SELFDESTRUCT",
```
axic: I think this should be `SUICIDE` in Frontier :)

And technically `INVALID` should only be added at a later hard fork if were are pedantic.

test/unittests/test_instructions.cpp
```diff
@@ -90,7 +90,7 @@ TEST(instructions, byzantium_hard_fork)
 
 TEST(instructions, name_gas_cost_equivalence)
 {
-    for (auto rev = EVMC_FRONTIER; rev <= EVMC_CONSTANTINOPLE;
+    for (auto rev = EVMC_FRONTIER; rev <= EVMC_LATEST_REVISON;
```
axic: `REVISON` -> `REVISION`

include/evmc/evmc.h
```diff
@@ -659,7 +659,9 @@ enum evmc_revision
     EVMC_TANGERINE_WHISTLE = 2,
     EVMC_SPURIOUS_DRAGON = 3,
     EVMC_BYZANTIUM = 4,
-    EVMC_CONSTANTINOPLE = 5
+    EVMC_CONSTANTINOPLE = 5,
+
+    EVMC_LATEST_REVISON = EVMC_CONSTANTINOPLE
```
axic: I wonder if this should be rather `LATEST_STABLE_REVISION` and point to `BYZANTIUM`?

lib/instructions/instruction_names.c
chfast: The `INVALID` EIP does not have the date specified. I think we can assume that this is a kind of change that affects EVM also back in time.
The similar case is for SELFDESTRUCT (EIP-6) which was not the part of Homestead HF officially.

include/evmc/evmc.h
chfast: Not sure, probably yes. Would you use it anywhere in Hera?

Maybe name it `DEFAULT_REVISION`?

lib/instructions/instruction_names.c
```diff
@@ -0,0 +1,1060 @@
+/* EVMC: Ethereum Client-VM Connector API.
+ * Copyright 2018 Pawel Bylica.
+ * Licensed under the MIT License. See the LICENSE file.
+ */
+
+#include <evmc/instructions.h>
+
+static const char* constantinople_names[256] = {
```
gumb0: These lists are hard to maintain.
Maybe at least add delimiting comment lines like
```
   // 0x0
   "STOP",
   "ADD",
...
   // 0x10
...
```

chfast: Added numbers in comments.

PR #35, opened 2018-06-19
CMake package config

chfast:


create_github_release.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python3
+
+import requests
+
+owner = 'ethereum-mining'
```
gumb0: Does this need to be changed?

chfast: Removed.

PR #36, opened 2018-06-20
Entrypoint alias

chfast:
This adds a helper macro that creates the symbol alias "evmc_create" for the given EVMC create function.

### DIFF
--- a/include/evmc/utils.h
+++ b/include/evmc/utils.h
@@ -1,0 +1,19 @@
+/* EVMC: Ethereum Client-VM Connector API.
+ * Copyright 2018 Pawel Bylica.
+ * Licensed under the MIT License. See the LICENSE file.
+ */
+
+#pragma once
+
+#if __cplusplus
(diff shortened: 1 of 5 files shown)


include/evmc/utils.h
```diff
@@ -0,0 +1,19 @@
...
+#pragma once
+
+#if __cplusplus
+#define EVMC_EXTERNC extern "C"
+#else
+#define EVMC_EXTERNC
+#endif
+
+#if _MSC_VER
+#define EVMC_ENTRYPOINT(function_name)
```
chfast: @gumb0 do you want to play with this on Windows? 
See https://stackoverflow.com/questions/2290587/gcc-style-weak-linking-in-visual-studio

axic:
Should be rebased on top of #37.

chfast:
It's hard (or impossible) to get similar feature on Windows. I propose to and search heuristic on the client side based on the file name.

PR #38, opened 2018-06-26
"Create" function name search heuristic

chfast:


test/vmtester/vmtester.cpp
```diff
@@ -35,6 +35,38 @@ evmc_instance* get_vm_instance()
     return vm.get();
 }
 
+std::vector<std::string> get_vm_names(const fs::path path)
```
axic: Why not make this into a helper in the library (like instructions)? It has to have the filename as an input string though, so in the `vmtester` you could have this extra function to deal with `fs::path`.

chfast: Maybe I will do it later to use in aleth. But not useful for geth anyway.

include/evmc/evmc.h
```diff
@@ -732,16 +732,19 @@ struct evmc_instance
...
+#if EVMC_DOCUMENTATION
 /**
  * Example of a function creating an instance of an example EVM implementation.
  *
- *  Each EVM implementation MUST provide a function returning an EVM instance.
- *  The function SHOULD be named `evmc_create_<vm-name>(void)`.
+ * Each EVM implementation MUST provide a function returning an EVM instance.
+ * The function SHOULD be named `evmc_create_<vm-name>(void)`. If the VM name contains hyphens
+ * replaces them with underscores in the function names. For VMs distributed as shared libraries,
+ * the name of the library SHOULD match the VM name.
```
axic: `libhera` -> `evmc_create_libhera` ?

chfast: The "lib" prefix is skipped if present.

chfast: Done.

PR #40, opened 2018-07-05
Add evmc::loader library to support dynamic loading

chfast:


chfast:
Can you review this?

include/evmc/loader.h
```diff
@@ -0,0 +1,24 @@
...
+
+enum evmc_loader_error_code
+{
+    EVMC_ERRC_SUCCESS = 0,
+    EVMC_ERRC_CANNOT_OPEN,
+    EVMC_ERRC_SYMBOL_NOT_FOUND,
+    EVMC_ERRC_INVALID_ARGUMENT,
+};
+
+struct evmc_instance* evmc_load(const char* filename, enum evmc_loader_error_code* error_code);
```
axic: Should have documentation here. Also mention that `error_code` can be `NULL` and in that case it won't be set.

include/evmc/loader.h
```diff
@@ -0,0 +1,24 @@
...
+
+#pragma once
+
+#if __cplusplus
+extern "C" {
+#endif
+
+enum evmc_loader_error_code
+{
+    EVMC_ERRC_SUCCESS = 0,
```
axic: Why not `EVMC_LOADER_SUCCESS`?

include/evmc/loader.h
axic: Okay this was added with a subsequent commit.

include/evmc/loader.h
axic: @chfast this is still outstanding.

include/evmc/loader.h
```diff
@@ -0,0 +1,58 @@
...
+ * If the file does not exists or is not a valid shared library the ::EVMC_ERRC_CANNOT_OPEN error
+ * code is signaled and NULL is returned.
+ *
+ * After the DLL is successfully loaded the function tries to find the EVM create function in the
+ * library. The `filename` is used to guess the EVM name and the name of the create function.
+ * The create function name is constructed as following:
+ * - the filename is taken from the path,
+ * - the "lib" prefix and file extension are stripped from the name,
+ * - all "-" are replaced with "_" to construct _full name_,
+ * - the last stem from the _full name_ separated with "_" is taken to construct the _short name_,
```
axic: Can you explain this?

lib/loader/loader.c
```diff
@@ -0,0 +1,118 @@
...
+    DLL_HANDLE handle = DLL_OPEN(filename);
+    if (!handle)
+    {
+        ec = EVMC_ERRC_CANNOT_OPEN;
+        goto exit;
+    }
+
+    // Create name buffer with the prefix.
+    const char prefix[] = "evmc_create_";
+    const size_t prefix_length = strlen(prefix);
```
axic: Wouldn't `sizeof(prefix)` work here?

chfast: 1. Then I would need `sizeof(prefix) - 1`.
2. The compiler is replacing `strlen(prefix)` with a constant.

include/evmc/loader.h
chfast: At first I wanted to use a subset of  error codes from C++ errc: https://en.cppreference.com/w/cpp/error/errc, but in the end they didn't match.
We can either make it evmc-generic: evmc_loader_error_code -> evmc_error_code, or use `EVMC_LOADER_` prefix as you proposed.

include/evmc/loader.h
chfast: How about?

> The _full name_ is split by "_" char and the last item is taken to form the _short name_.

lib/loader/loader.c
axic: Are you sure it replaces it with a constant?

include/evmc/loader.h
axic: I'd vote for `EVMC_LOADER_`

include/evmc/loader.h
axic: Can you give an example of a filename having both short and full names?

lib/loader/loader.c
chfast: https://godbolt.org/g/1u39cy

chfast: That's not super smart. They have hardcoded knowledge of some C standard library functions.

include/evmc/loader.h
```diff
@@ -0,0 +1,58 @@
...
+enum evmc_loader_error_code
+{
+    EVMC_ERRC_SUCCESS = 0,
+    EVMC_ERRC_CANNOT_OPEN,
+    EVMC_ERRC_SYMBOL_NOT_FOUND,
+    EVMC_ERRC_INVALID_ARGUMENT,
+};
+
+/**
+ * Dynamically loads the shared object and with an EVM implementation.
```
gumb0: Some verb missing after "and"?

include/evmc/loader.h
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * This function tries to open a DLL at the given `filename`. On UNIX-like systems dlopen() function
+ * is used. On Windows LoadLibrary() function is used.
+ *
+ * If the file does not exists or is not a valid shared library the ::EVMC_ERRC_CANNOT_OPEN error
```
gumb0: does not exist

include/evmc/loader.h
gumb0: Yeah, example in the comment would be helpful, like `for "libfull-evm-name.so" the full name "evmc_create_full_evm_name" and "evmc_create_name" are checked in the library`.

gumb0: Also. why do we need these complicated rules, can't we just have one hard-coded name for create function in any evmc library?

lib/loader/loader.c
```diff
@@ -0,0 +1,118 @@
...
+#else
+#include <dlfcn.h>
+#define DLL_HANDLE void*
+#define DLL_OPEN(filename) dlopen(filename, RTLD_LAZY)
+#define DLL_CLOSE(handle) dlclose(handle)
+#define DLL_GET_CREATE_FN(handle, name) (evmc_create_fn)(uintptr_t) dlsym(handle, name)
+#define HAVE_STRCPY_S 0
+#endif
+
+#define PATH_MAX_LENGHT 4096
```
gumb0: length

include/evmc/loader.h
axic: Want to support both static and dynamic linking of the same source.

lib/loader/loader.c
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    // Create name buffer with the prefix.
+    const char prefix[] = "evmc_create_";
+    const size_t prefix_length = strlen(prefix);
+    char name[sizeof(prefix) + PATH_MAX_LENGHT];
+    strcpy_s(name, sizeof(name), prefix);
+
+    // Find filename in the path.
+    const char* sep_pos = strrchr(filename, '/');
```
gumb0: will not work when windows path provided

lib/loader/loader.c
```diff
@@ -0,0 +1,118 @@
...
+    if (!create_fn)
+    {
+        DLL_CLOSE(handle);
+        ec = EVMC_ERRC_SYMBOL_NOT_FOUND;
+    }
+
+exit:
+    if (error_code)
+        *error_code = ec;
+    return create_fn;
```
gumb0: So `handle` stays open forever after sucessful `evmc_load`, is this ok?

lib/loader/loader.c
```diff
@@ -0,0 +1,118 @@
...
+    // Search for the "full name" based function name.
+    create_fn = DLL_GET_CREATE_FN(handle, name);
+    if (!create_fn)
+    {
+        // Try the "short name" based function name.
+        const char* short_name_pos = strrchr(name, '_');
+        if (short_name_pos)
+        {
+            short_name_pos += 1;
+            memmove(name + prefix_length, short_name_pos, strlen(short_name_pos) + 1);
```
gumb0: Does this still work fine when file name ends with `_`? Add a test for it

lib/loader/loader.c
axic: I don't think you can close it if there's a pointer into it?

gumb0: On Windows there is `FreeLibrary` https://msdn.microsoft.com/en-us/library/windows/desktop/ms683152(v=vs.85).aspx

chfast: The `dlclose()` will unload the library from the memory so you cannot use functions from it (similarly on Windows). We are doing so only when we could not find the create function because then the library is not needed any more. In case we use the library, it is never unloaded.

lib/loader/loader.c
chfast: True. I have to add a test for it. What do you think about doing a normalization on Windows before this by replacing `\` with `/`. I believe on Windows you can have your separators mixed.

include/evmc/loader.h
chfast: On POSIX we could use symbol alias in the .so library. On Windows it is much harder. In the end I had to just define `evmc_create()` function that executes `evmc_create_interpreter()` and disable the `evmc_create()` when building a static library. This generally gets messy and requires toolchain specific build flags. And I expect that to be even harder for rust or nim. So I decided to provide this naming convention and keep the implementation of it in this single place.

lib/loader/loader.c
gumb0: I think it should work

lib/loader/loader.c
chfast: It works. For `lib_.so` the full name is "_", and short name is "". I'm adding unit tests for it.

chfast:
Ready. Rebase is needed before merge.

PR #41, opened 2018-07-12
Go bindings

chfast:
Can you help with the review @gballet, @fjl?

Appropriate go-ethereum PR: https://github.com/ethereum/go-ethereum/pull/17050

### DIFF
--- a/bindings/go/evmc/evmc.go
+++ b/bindings/go/evmc/evmc.go
@@ -1,0 +1,273 @@
+// EVMC: Ethereum Client-VM Connector API.
+// Copyright 2018 Pawel Bylica.
+// Licensed under the MIT License. See the LICENSE file.
+
+package evmc
+
+/*
+#cgo CFLAGS:  -I${SRCDIR}/.. -Wall -Wextra
+#cgo !windows LDFLAGS: -ldl
+
+#include <evmc/evmc.h>
+#include <evmc/helpers.h>
+#include <evmc/loader.h>
+
+#include <stdlib.h>
+#include <string.h>
+
+static inline int set_option(struct evmc_instance* instance, char* name, char* value)
+{
+	int ret = evmc_set_option(instance, name, value);
+	free(name);
+	free(value);
+	return ret;
+}
+
+struct extended_context
+{
+	struct evmc_context context;
+	int64_t index;
+};
+
+extern const struct evmc_context_fn_table evmc_go_fn_table;
+
+static struct evmc_result execute_wrapper(struct evmc_instance* instance, int64_t context_index, enum evmc_revision rev,
+	const struct evmc_address* destination, const struct evmc_address* sender, const struct evmc_uint256be* value,
+	const uint8_t* input_data, size_t input_size, const struct evmc_uint256be* code_hash, int64_t gas,
+	int32_t depth, enum evmc_call_kind kind, uint32_t flags, const uint8_t* code, size_t code_size)
+{
+	struct evmc_uint256be create2_salt = {};
+	struct evmc_message msg = {
+		*destination,
+		*sender,
+		*value,
+		input_data,
+		input_size,
+		*code_hash,
+		create2_salt,
+		gas,
+		depth,
+		kind,
+		flags,
+	};
+
+	struct extended_context ctx = {{&evmc_go_fn_table}, context_index};
+	return instance->execute(instance, &ctx.context, rev, &msg, code, code_size);
+}
+*/
+import "C"
+
+import (
+	"errors"
+	"fmt"
+	"sync"
+	"unsafe"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// Static asserts.
+const (
+	_ = uint(common.HashLength - C.sizeof_struct_evmc_uint256be) // The size of evmc_uint256be equals the size of Hash.
+	_ = uint(C.sizeof_struct_evmc_uint256be - common.HashLength)
+	_ = uint(common.AddressLength - C.sizeof_struct_evmc_address) // The size of evmc_address equals the size of Address.
+	_ = uint(C.sizeof_struct_evmc_address - common.AddressLength)
+)
+
+type Error int32
+
+func (err Error) IsInternalError() bool {
+	return err < 0
+}
+
+func (err Error) Error() string {
+	code := C.enum_evmc_status_code(err)
+
+	switch code {
+	case C.EVMC_FAILURE:
+		return "evmc: failure"
+	case C.EVMC_REVERT:
+		return "evmc: revert"
+	case C.EVMC_OUT_OF_GAS:
+		return "evmc: out of gas"
+	case C.EVMC_INVALID_INSTRUCTION:
+		return "evmc: invalid instruction"
+	case C.EVMC_UNDEFINED_INSTRUCTION:
+		return "evmc: undefined instruction"
+	case C.EVMC_STACK_OVERFLOW:
+		return "evmc: stack overflow"
+	case C.EVMC_STACK_UNDERFLOW:
+		return "evmc: stack underflow"
+	case C.EVMC_BAD_JUMP_DESTINATION:
+		return "evmc: bad jump destination"
+	case C.EVMC_INVALID_MEMORY_ACCESS:
+		return "evmc: invalid memory access"
+	case C.EVMC_CALL_DEPTH_EXCEEDED:
+		return "evmc: call depth exceeded"
+	case C.EVMC_STATIC_MODE_VIOLATION:
+		return "evmc: static mode violation"
+	case C.EVMC_PRECOMPILE_FAILURE:
+		return "evmc: precompile failure"
+	case C.EVMC_CONTRACT_VALIDATION_FAILURE:
+		return "evmc: contract validation failure"
+	case C.EVMC_INTERNAL_ERROR:
+		return "evmc: internal error"
+	case C.EVMC_REJECTED:
+		return "evmc: rejected"
+	}
+
+	if code < 0 {
+		return fmt.Sprintf("evmc: unknown internal error (%d)", int32(code))
+	}
+
+	panic(fmt.Sprintf("evmc: unknown status code %d", int32(code)))
+}
+
+const (
+	Failure = Error(C.EVMC_FAILURE)
+	Revert  = Error(C.EVMC_REVERT)
+)
+
+type Revision int32
+
+const (
+	Frontier         Revision = C.EVMC_FRONTIER
+	Homestead        Revision = C.EVMC_HOMESTEAD
+	TangerineWhistle Revision = C.EVMC_TANGERINE_WHISTLE
+	SpuriousDragon   Revision = C.EVMC_SPURIOUS_DRAGON
+	Byzantium        Revision = C.EVMC_BYZANTIUM
+	Constantinople   Revision = C.EVMC_CONSTANTINOPLE
+)
+
+type Instance struct {
+	handle *C.struct_evmc_instance
+}
+
+func Load(filename string) (instance *Instance, err error) {
+	cfilename := C.CString(filename)
+	var loaderErr C.enum_evmc_loader_error_code
+	handle := C.evmc_load_and_create(cfilename, &loaderErr)
+	C.free(unsafe.Pointer(cfilename))
+	switch loaderErr {
+	case C.EVMC_LOADER_SUCCESS:
+		instance = &Instance{handle}
+	case C.EVMC_LOADER_CANNOT_OPEN:
+		err = fmt.Errorf("evmc loader: cannot open %s", filename)
+	case C.EVMC_LOADER_SYMBOL_NOT_FOUND:
+		err = fmt.Errorf("evmc loader: the EVMC create function not found in %s", filename)
+	case C.EVMC_LOADER_INVALID_ARGUMENT:
+		panic("evmc loader: filename argument is invalid")
+	case C.EVMC_LOADER_INSTANCE_CREATION_FAILURE:
+		err = errors.New("evmc loader: VM instance creation failure")
+	case C.EVMC_LOADER_ABI_VERSION_MISMATCH:
+		err = errors.New("evmc loader: ABI version mismatch")
+	default:
+		panic(fmt.Sprintf("evmc loader: unexpected error (%d)", int(loaderErr)))
+	}
+	return instance, err
+}
+
+func (instance *Instance) Destroy() {
+	C.evmc_destroy(instance.handle)
+}
+
+func (instance *Instance) Name() string {
+	return C.GoString(instance.handle.name)
+}
+
+func (instance *Instance) Version() string {
+	return C.GoString(instance.handle.version)
+}
+
+func (instance *Instance) SetOption(name string, value string) (err error) {
+
+	r := C.set_option(instance.handle, C.CString(name), C.CString(value))
+	if r != 1 {
+		err = fmt.Errorf("evmc: option '%s' not accepted", name)
+	}
+	return err
+}
+
+func (instance *Instance) Execute(ctx HostContext, rev Revision,
+	destination common.Address, sender common.Address, value common.Hash, input []byte, codeHash common.Hash, gas int64,
+	depth int, kind CallKind, static bool, code []byte) (output []byte, gasLeft int64, err error) {
+
+	flags := C.uint32_t(0)
+	if static {
+		flags |= C.EVMC_STATIC
+	}
+
+	ctxId := addHostContext(ctx)
+	// FIXME: Clarify passing by pointer vs passing by value.
+	evmcDestination := evmcAddress(destination)
+	evmcSender := evmcAddress(sender)
+	evmcValue := evmcUint256be(value)
+	evmcCodeHash := evmcUint256be(codeHash)
+	result := C.execute_wrapper(instance.handle, C.int64_t(ctxId), uint32(rev), &evmcDestination, &evmcSender, &evmcValue,
+		bytesPtr(input), C.size_t(len(input)), &evmcCodeHash, C.int64_t(gas), C.int32_t(depth), C.enum_evmc_call_kind(kind),
+		flags, bytesPtr(code), C.size_t(len(code)))
+	removeHostContext(ctxId)
+
+	output = C.GoBytes(unsafe.Pointer(result.output_data), C.int(result.output_size))
+	gasLeft = int64(result.gas_left)
+	if result.status_code != C.EVMC_SUCCESS {
+		err = Error(result.status_code)
+	}
+
+	if result.release != nil {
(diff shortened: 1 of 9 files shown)


axic:
Can you rebase this please?

bindings/go/evmc/evmc.go
```diff
@@ -0,0 +1,258 @@
...
+#include <evmc/helpers.h>
+#include <evmc/loader.h>
+
+#include <stdlib.h>
+#include <string.h>
+
+inline int set_option(struct evmc_instance* instance, _GoString_ name, _GoString_ value)
+{
+	if (instance->set_option)
+		return instance->set_option(instance, name.p, value.p);
```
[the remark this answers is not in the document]
chfast: It's easier to use `_GoString_` type here, than use malloc/free to pass the string to the original helper.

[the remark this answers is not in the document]
axic: I mean instead of the if and return here, just wrap `evmc_set_option`?

bindings/go/evmc/evmc.go
```diff
@@ -0,0 +1,258 @@
...
+	return -1;
```
axic: Note: the `evmc_set_option` wrapper returns 0 in this case. Which one is preferred?

chfast: Here it returns 1, 0 or -1. This is not super needed, but I have different warnings for 0 and -1.

bindings/go/evmc/evmc.go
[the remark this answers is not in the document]
chfast: Oh, yes. Good idea.

axic:
@gballet would you mind having a look at the Go code? :)

Also please tag anyone else from the Go team who you think makes sense.

bindings/go/evmc/evmc.go
```diff
@@ -0,0 +1,258 @@
...
+func getHostContext(idx int) HostContext {
+	hostContextMapMu.Lock()
+	ctx := hostContextMap[idx]
+	hostContextMapMu.Unlock()
+	return ctx
+}
+
+func evmcUint256be(in common.Hash) C.struct_evmc_uint256be {
+	out := C.struct_evmc_uint256be{}
+	for i := 0; i < len(in); i++ {
```
axic: Can we assert that `len(in)` equals that of `uint256be`?

bindings/go/evmc/evmc.go
```diff
@@ -0,0 +1,258 @@
...
+		out.bytes[i] = C.uint8_t(in[i])
+	}
+	return out
+}
+
+func evmcAddress(address common.Address) C.struct_evmc_address {
+	r := C.struct_evmc_address{}
+	for i := 0; i < len(address); i++ {
```
axic: Same here.

bindings/go/evmc/host.go
```diff
@@ -0,0 +1,220 @@
...
+		out[i] = byte(in.bytes[i])
+	}
+	return out
+}
+
+func goByteSlice(data *C.uint8_t, size C.size_t) []byte {
+	if size == 0 {
+		return []byte{}
+	}
+	return (*[1 << 30]byte)(unsafe.Pointer(data))[:size:size]
```
axic: What is this magic doing? :)

bindings/go/evmc/host.c
```diff
@@ -0,0 +1,22 @@
+#include "_cgo_export.h"
+
+#include <stdlib.h>
+
+const struct evmc_context_fn_table fn_table = {
```
axic: I'm not sure how linking works and whether this global symbol would pollute any consumer of this go library, but probably it makes sense prefixing these.

bindings/go/evmc/evmc.go
```diff
@@ -0,0 +1,249 @@
...
+#cgo LDFLAGS: -ldl
+
+#include <evmc/evmc.h>
+#include <evmc/helpers.h>
+#include <evmc/loader.h>
+
+#include <stdlib.h>
+#include <string.h>
+
+inline int set_option(struct evmc_instance* instance, _GoString_ name, _GoString_ value)
```
axic: Not sure about linking visibility in cgo, but perhaps this should be static to limit the scope to this file?

chfast: It should.

bindings/go/evmc/host.c
chfast: Yes. I agree. I can also add "hidden" attribute to it.

bindings/go/evmc/host.go
chfast: Simple. You cast the `data` pointer to a pointer to a huge array and then take a slice of size `size` of it. This is a cgo why to treat C array as a Go slice.

bindings/go/evmc/evmc.go
chfast: There is static assert. You can't even check the struct size (without "unsafe" package). But this is not a big problem here, because the array access is safe (will panic if wrong, but I have not checked if the go compiler is able to optimize this).

bindings/go/evmc/evmc.go
```diff
@@ -0,0 +1,258 @@
...
+	msg.input_size = input.n;
+	msg.gas = gas;
+	msg.depth = depth;
+	msg.kind = EVMC_CALL;
+	msg.flags = flags;
+
+	memcpy(msg.destination.bytes, destination.p, sizeof(msg.destination));
+	memcpy(msg.sender.bytes, sender.p, sizeof(msg.sender));
+	memcpy(msg.value.bytes, value.p, sizeof(msg.value));
+	memcpy(msg.code_hash.bytes, code_hash.p, sizeof(msg.code_hash));
```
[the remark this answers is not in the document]
chfast: 1. I'm not sure using a C assert is good idea here.
2. Because only this Go module can use this function we know exactly what size it is (still prone to future mistakes).
3. Because this does not compile with Go 1.9, this might be rewritten anyway.

bindings/go/evmc/evmc.go
```diff
@@ -0,0 +1,247 @@
...
+	int64_t index;
+};
+
+extern const struct evmc_context_fn_table evmc_go_fn_table;
+
+static struct evmc_result execute_wrapper(struct evmc_instance* instance, int64_t context_index, enum evmc_revision rev,
+	_GoBytes_ destination, _GoBytes_ sender, _GoBytes_ value, _GoBytes_ input, _GoBytes_ code_hash, int64_t gas,
+	int32_t depth, uint32_t flags, _GoBytes_ code)
+{
+	struct evmc_message msg;
```
gumb0: Would it be good idea to default-initialize it here first? (` = {};` is it valid in C?)

bindings/go/evmc/evmc.go
```diff
@@ -0,0 +1,247 @@
...
+	C.free(unsafe.Pointer(cfilename))
+	switch loaderErr {
+	case C.EVMC_LOADER_SUCCESS:
+		instance = &Instance{handle}
+	case C.EVMC_LOADER_CANNOT_OPEN:
+		err = fmt.Errorf("evmc loader: cannot open %s", filename)
+	case C.EVMC_LOADER_SYMBOL_NOT_FOUND:
+		err = fmt.Errorf("evmc loader: the EVMC create function not found in %s", filename)
+	case C.EVMC_LOADER_INVALID_ARGUMENT:
+		panic("evmc loader: filename argument is empty")
```
gumb0: also can be too long

axic:
Basically I would like to have some guarantees that memory sizes equal between the C and Go data structures. Is that achievable somehow?

bindings/go/evmc/evmc.go
axic: Why not `filename argument is invalid (empty, too long, etc.)`?

bindings/go/evmc/host.go
```diff
@@ -0,0 +1,220 @@
...
+)
+
+type CallKind int
+
+const (
+	Call         = C.EVMC_CALL
+	DelegateCall = C.EVMC_DELEGATECALL
+	CallCode     = C.EVMC_CALLCODE
+	Create       = C.EVMC_CREATE
+)
```
axic: Need to add `Create2` here.

axic:
Needs to be rebased to pick up latest ABI.

bindings/go/evmc/evmc.go
chfast: Not in the pedantic C, but here it should work.

bindings/go/evmc/evmc.go
```diff
@@ -0,0 +1,255 @@
...
+
+struct extended_context
+{
+	struct evmc_context context;
+	int64_t index;
+};
+
+extern const struct evmc_context_fn_table evmc_go_fn_table;
+
+static struct evmc_result execute_wrapper(struct evmc_instance* instance, int64_t context_index, enum evmc_revision rev,
```
gumb0: Should this now have `create2_salt` parameter?

gumb0: ah it's only for calls...

bindings/go/evmc/evmc.go
```diff
@@ -0,0 +1,255 @@
...
+
+	if code < 0 {
+		return fmt.Sprintf("evmc: unknown internal error (%d)", int32(code))
+	}
+
+	panic(fmt.Sprintf("evmc: unknown status code %d", int32(code)))
+}
+
+const (
+	Failure = Error(C.EVMC_FAILURE)
```
gumb0: Why consts only for two status codes?

bindings/go/evmc/evmc.go
chfast: It should be passed back to the Host interface. I have not done this yet.

bindings/go/evmc/evmc.go
chfast: I'm waiting for a comment from Go devs, because sometimes they predefined errors as consts and sometimes create errors at hoc.

bindings/go/evmc/host.go
```diff
@@ -0,0 +1,221 @@
...
+	EmitLog(addr common.Address, topics []common.Hash, data []byte)
+	Call(kind CallKind,
+		destination common.Address, sender common.Address, value *big.Int, input []byte, gas int64, depth int,
+		static bool) (output []byte, gasLeft int64, createAddr common.Address, err error)
+}
+
+//export accountExists
+func accountExists(pCtx unsafe.Pointer, pAddr *C.struct_evmc_address) C.int {
+	idx := int((*C.struct_extended_context)(pCtx).index)
+	ctx := getHostContext(idx)
```
gumb0: Is it possible that `idx` points to non-existing context? Should we check it here or in `getHostContext` ?

bindings/go/evmc/evmc.go
axic: This has been fixed by the assertion commit.

bindings/go/evmc/evmc.go
axic: This has been fixed by the assertion commit.

bindings/go/evmc/evmc.go
```diff
@@ -0,0 +1,255 @@
...
+func (instance *Instance) SetOption(name string, value string) (err error) {
+	r := C.set_option(instance.handle, name+"\x00", value+"\x00")
+	if r != 1 {
+		err = fmt.Errorf("evmc: option '%s' not accepted", name)
+	}
+	return err
+}
+
+func (instance *Instance) Execute(ctx HostContext, rev Revision,
+	destination common.Address, sender common.Address, value common.Hash, input []byte, codeHash common.Hash, gas int64, depth int, static bool,
```
axic: Why is `value` of `Hash` type?

bindings/go/evmc/evmc.go
```diff
@@ -0,0 +1,255 @@
...
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// Static asserts.
+const (
+	_ = uint(common.HashLength - C.sizeof_struct_evmc_uint256be) // The size of evmc_uint256be equals the size of Hash.
+	_ = uint(C.sizeof_struct_evmc_uint256be - common.HashLength)
+	_ = uint(common.AddressLength - C.sizeof_struct_evmc_address) // The size of evmc_address equals the size of Address.
+	_ = uint(C.sizeof_struct_evmc_address - common.AddressLength)
+)
```
axic: Can you also add "asserts" for:
- `evmc_message.destination == AddressLength`
- `evmc_message.sender == AddressLength`
- `evmc_message.value == HashLength` (?? see above)
- `evmc_message.code_hash == HashLength`

bindings/go/evmc/evmc.go
```diff
@@ -0,0 +1,255 @@
...
+
+type Error int32
+
+func (err Error) IsInternalError() bool {
```
axic: This doesn't seem to be used below (or anywhere?).

bindings/go/evmc/evmc.go
```diff
@@ -0,0 +1,255 @@
...
+	return err < 0
+}
+
+func (err Error) Error() string {
```
axic: Would a name like `ErrorToString` be more descriptive?

bindings/go/evmc/evmc.go
```diff
@@ -0,0 +1,255 @@
...
+static struct evmc_result execute_wrapper(struct evmc_instance* instance, int64_t context_index, enum evmc_revision rev,
+	_GoBytes_ destination, _GoBytes_ sender, _GoBytes_ value, _GoBytes_ input, _GoBytes_ code_hash, int64_t gas,
+	int32_t depth, uint32_t flags, _GoBytes_ code)
+{
+	struct evmc_message msg = {};
+	msg.input_data = input.p;
+	msg.input_size = input.n;
+	msg.gas = gas;
+	msg.depth = depth;
+	msg.kind = EVMC_CALL;
```
axic: This means there's no way to report a CREATE transaction?

axic: Hera does care :)

bindings/go/evmc/evmc.go
axic: Ah ok, so all the wrappers return a standard `Error` object which contain an internal error value and this can be used to turn that into a string.

bindings/go/evmc/evmc.go
```diff
@@ -0,0 +1,255 @@
...
+		return "evmc: internal error"
+	case C.EVMC_REJECTED:
+		return "evmc: rejected"
+	}
+
+	if code < 0 {
+		return fmt.Sprintf("evmc: unknown internal error (%d)", int32(code))
+	}
+
+	panic(fmt.Sprintf("evmc: unknown status code %d", int32(code)))
```
axic: Not sure it is a good idea making this a panic?

bindings/go/evmc/loader.c
```diff
@@ -0,0 +1,6 @@
+/* EVMC: Ethereum Client-VM Connector API.
+ * Copyright 2018 Pawel Bylica.
+ * Licensed under the MIT License. See the LICENSE file.
+ */
+
+#include "../../../lib/loader/loader.c"
```
axic: Why not make this file a symlink instead? Is that not allowed by govendor or package managers?

chfast: Do symlinks work on Windows?

axic: Git allows symlinks, I am sure it resolves it somehow on Windows. @gumb0 do you know by any chance what git is doing in that case?

bindings/go/evmc/evmc.go
chfast: It means the `evmc.Error` type can used as the builtin `error` type.

bindings/go/evmc/evmc_test.go
```diff
@@ -0,0 +1,24 @@
...
+// Licensed under the MIT License. See the LICENSE file.
+
+package evmc
+
+import (
+	"os"
+	"testing"
+)
+
+func TestLoad(t *testing.T) {
```
axic: Where is this used? What file is it loading?

gumb0:
Needs to be udated for EXTCODEHASH support

bindings/go/evmc/evmc.go
chfast: This has been reworked.

bindings/go/evmc/host.go
chfast: This should be always valid. Otherwise Go will raise an exception.

chfast:
> Needs to be udated for EXTCODEHASH support



> Needs to be udated for EXTCODEHASH support

EXTCODEHASH is included.

axic:
@chfast is anything blocking the merging of this?

chfast:
Missing reviews?

I'd go and merge this if no more obvious issues. Maybe I should fix one FIXME comment from go-ethereum.

Also this might be not compatible with the master branch. I base this on v5.0.0. But I can handle this.

The benefit of merging this is to have it tested on CI in new PRs so the Go bindings will match the recent changes in the EVMC API.

axic:
Even if the geth team hasn't reviewed this I'd merge it for the reasons you've said.

bindings/go/evmc/evmc.go
```diff
@@ -15,11 +15,12 @@ package evmc
...
 
-static inline int set_option(struct evmc_instance* instance, _GoString_ name, _GoString_ value)
+static inline int set_option(struct evmc_instance* instance, char* name, char* value)
 {
-	if (instance->set_option)
-		return instance->set_option(instance, name.p, value.p);
-	return 0;
+	int r = evmc_set_option(instance, name, value);
+	free(name);
+	free(value);
```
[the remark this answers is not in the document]
chfast: Yes.
```
// Go string to C string
// The C string is allocated in the C heap using malloc.
// It is the caller's responsibility to arrange for it to be
// freed, such as by calling C.free (be sure to include stdlib.h
// if C.free is needed).
func C.CString(string) *C.char
```

bindings/go/evmc/host.c
```diff
@@ -0,0 +1,28 @@
...
+    (evmc_get_code_hash_fn)getCodeHash,
+    (evmc_copy_code_fn)copyCode,
+    (evmc_selfdestruct_fn)selfdestruct,
+    (evmc_call_fn)call,
+    (evmc_get_tx_context_fn)getTxContext,
+    (evmc_get_block_hash_fn)getBlockHash,
+    (evmc_emit_log_fn)emitLog,
+};
+
+void evmc_go_free_result_output(const struct evmc_result* result)
```
axic: Actually instead of this one could use `evmc_release_result` from `helpers.h`.

chfast: Actually not. This is the implementation. `evmc_release_result` is just a wrapper that calls the implementation.

axic: My bad, too early.

PR #42, opened 2018-07-13
Extensions

chfast:
- Fixes documentation.
- Add some helpers for EVMC and loader modules.

### DIFF
--- a/include/evmc/helpers.h
+++ b/include/evmc/helpers.h
@@ -1,0 +1,54 @@
+/* EVMC: Ethereum Client-VM Connector API.
+ * Copyright 2018 Pawel Bylica.
+ * Licensed under the MIT License. See the LICENSE file.
+ */
+
+/**
+ * EVMC Helpers
+ *
+ * A collection of helper functions for invoking a VM instance methods.
+ * These are convenient for languages where invoking function pointers
+ * is "ugly" or impossible (such as Go).
+ *
+ * @defgroup helpers EVMC Helpers
+ * @{
+ */
+#pragma once
+
+#include <evmc/evmc.h>
+
+/**
+ * Destroys the VM instance.
+ *
+ * @see evmc_destroy_fn
+ */
+static inline void evmc_destroy(struct evmc_instance* instance)
+{
+    instance->destroy(instance);
+}
+
+/**
+ * Sets the option for the VM instance, if the feature is supported by the VM.
+ *
+ * @see evmc_set_option_fn
+ */
+static inline int evmc_set_option(struct evmc_instance* instance,
+                                  char const* name,
+                                  char const* value)
+{
+    if (instance->set_option)
+        return instance->set_option(instance, name, value);
+    return 0;
+}
+
+/**
+ * Releases the resources allocated to the execution result.
+ *
+ * @see evmc_release_result_fn
+ */
+static inline void evmc_release_result(struct evmc_result* result)
+{
+    result->release(result);
+}
+
+/** @} */
--- a/include/evmc/instructions.h
+++ b/include/evmc/instructions.h
@@ -3,12 +3,18 @@
  * Licensed under the MIT License. See the LICENSE file.
  */
 
+/**
+ * EVM Instruction Tables
+ *
+ * A collection of metrics for EVM1 instruction set.
+ *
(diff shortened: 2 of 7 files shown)


lib/loader/loader.c
```diff
@@ -121,3 +122,26 @@ evmc_create_fn evmc_load(const char* filename, enum evmc_loader_error_code* erro
         *error_code = ec;
     return create_fn;
 }
+
+struct evmc_instance* evmc_load_and_create(const char* filename, enum evmc_loader_error_code* error_code)
+{
+    evmc_create_fn create_fn = evmc_load(filename, error_code);
```
axic: I think it should be safe (at least on POSIX) to call this multiple times and not pollute the memory (since `dlopen` should just return the reference if it was loaded already IIRC). However, is that the case for Windows?

Do we need to maintain our local list of loaded EMVC libraries in order to make this helper callable multiple times?

axic:
I don't think the go bindings file belong to this PR.

lib/loader/loader.c
chfast: It's the same of Windows.

>  Calling LoadLibrary increments the reference count.

chfast:
> I don't think the go bindings file belong to this PR.

Removed.

include/evmc/helpers.h
```diff
@@ -0,0 +1,52 @@
+/* EVMC: Ethereum Client-VM Connector API.
+ * Copyright 2018 Pawel Bylica.
+ * Licensed under the MIT License. See the LICENSE file.
+ */
+
+/**
+ * EVMC Helpers
+ *
+ * A collection of helper functions for invoking a VM instance methods.
+ * These are convenient e.g. for Go bindings, because this is the only way to invoke a function
```
axic: Would reword that to something like: `These are convenient for languages where invoking foreign runtime loaded functions is "ugly" or problematic (such as Go).`

include/evmc/helpers.h
```diff
@@ -0,0 +1,52 @@
...
+#pragma once
+
+#include <evmc/evmc.h>
+
+/**
+ * Destroys the VM instance.
+ *
+ * @see evmc_destroy_fn
+ */
+inline void evmc_destroy(struct evmc_instance* instance)
```
axic: These should be `static inline` even though we have `#pragma once`. Not even sure we should have pragma once since that is more of a C++ism.

include/evmc/instructions.h
```diff
@@ -3,13 +3,19 @@
...
+ *
+ * @defgroup instructions EVM Instructions
+ * @{
+ */
 #pragma once
 
 #include <evmc/evmc.h>
 #include <evmc/utils.h>
 
-#include <stdint.h>
```
axic: Why remove this? `int16_t` still requires it.

include/evmc/loader.h
```diff
@@ -48,18 +70,31 @@ enum evmc_loader_error_code
...
- *                    ::EVMC_ERRC_SUCCESS on success or any other error code as described above.
+ *                    ::EVMC_LOADER_SUCCESS on success or any other error code as described above.
  * @return            The pointer to the EVM create function or NULL.
  */
 evmc_create_fn evmc_load(const char* filename, enum evmc_loader_error_code* error_code);
 
+/**
+ * Dynamically loads the VM DLL and creates the VM instance.
+ *
+ * This is a macro of creating the VM instance with the function returned from evmc_load().
```
axic: `macro for`

include/evmc/helpers.h
chfast: `#pragma once` is not C++. Is... pragma. All compilers support it.

lib/loader/loader.c
axic: It should be documented then that calling `evmc_load` and `evmc_load_and_create` multiple times should be safe and not blow up memory usage.

include/evmc/instructions.h
chfast: It's already included in `<evmc/evmc.h>`.

include/evmc/helpers.h
axic: Still, if we target FFI in languages which may not support pragma, can we just have `static` here since this in a header?

PR #45, opened 2018-07-24
Support CREATE2 opcode in Constantinople

gumb0:


gumb0:
Still need to add it to instructions lib.

include/evmc/evmc.h
```diff
@@ -70,7 +70,8 @@ enum evmc_call_kind
     EVMC_CALL = 0,         /**< Request CALL. */
     EVMC_DELEGATECALL = 1, /**< Request DELEGATECALL. The value param ignored. */
     EVMC_CALLCODE = 2,     /**< Request CALLCODE. */
-    EVMC_CREATE = 3        /**< Request CREATE. Semantic of some params changes. */
+    EVMC_CREATE = 3,       /**< Request CREATE. Semantic of some params changes. */
```
chfast: You can remove the "Semantic of some params changes" part in both cases.

axic: I think these should state which `evmc_revision` they are valid in (frontier, homestead, byzantium, constantinople).

include/evmc/evmc.h
```diff
@@ -116,6 +118,13 @@ struct evmc_message
      */
     struct evmc_uint256be code_hash;
 
+    /**
+     * The optional value used in new contract address construction.
+     *
+     *  Ignored unless kind == EVMC_CREATE2.
+     */
+    struct evmc_uint256be salt;
```
chfast: How about we name it `create_salt`?

include/evmc/evmc.h
```diff
@@ -117,6 +118,13 @@ struct evmc_message
      */
     struct evmc_uint256be code_hash;
 
+    /**
+     * The optional value used in new contract address construction.
+     *
+     *  Ignored unless kind == EVMC_CREATE2.
```
axic: `unless kind is EVMC_CREATE2`.

test/vmtester/tests.cpp
```diff
@@ -12,7 +12,6 @@ static_assert(sizeof(evmc_uint256be) == 32, "evmc_uint256be is too big");
 static_assert(sizeof(evmc_address) == 20, "evmc_address is too big");
 static_assert(sizeof(evmc_result) <= 64, "evmc_result does not fit cache line");
 static_assert(sizeof(evmc_instance) <= 64, "evmc_instance does not fit cache line");
-static_assert(sizeof(evmc_message) <= 18 * 8, "evmc_message not optimally packed");
```
axic: Should this be kept with an updated size?

chfast: It does not anything useful. I was just checking obvious struct fields alignment.

include/evmc/evmc.h
axic: `create2_salt` for explicitlness?

axic: Otherwise `creation_salt`?

chfast: `create2_salt` :+1: 
Similarly to `create_address`.

gumb0:
Renamed to `create_salt`

axic:
This also needs an ABI version bump.

gumb0:
Bumped ABI version to 2

PR #46, opened 2018-07-25
Add EVMC_NOEXCEPT helper

axic:
Moved from https://github.com/ethereum/cpp-ethereum/blob/develop/libaleth-interpreter/interpreter.h#L11.

### DIFF
--- a/include/evmc/utils.h
+++ b/include/evmc/utils.h
@@ -10,3 +10,9 @@
 #else
 #define EVMC_EXPORT __attribute__((visibility("default")))
 #endif
+
+#if __cplusplus
+#define EVMC_NOEXCEPT noexcept
+#else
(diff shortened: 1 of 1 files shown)


include/evmc/utils.h
```diff
@@ -10,3 +10,9 @@
 #else
 #define EVMC_EXPORT __attribute__((visibility("default")))
 #endif
+
+#if __cplusplus
+#define EVMC_NOEXCEPT noexcept
```
chfast: I was thinking about naming it just `NOEXCEPT` but checking if the macro was not previously defined.

axic: The same could be said for `EXPORT` I think? A lot of systems use it for the same purpose.

chfast: I would go for it.

PR #49, opened 2018-07-31
Support for EXTCODEHASH

gumb0:


gumb0:
@axic I've fixed the tests and updated the instrucrtions lib, you might want to take another look

examples/capi.c
```diff
@@ -16,13 +16,6 @@ struct evmc_uint256be balance(struct evmc_context* context, const struct evmc_ad
     return ret;
 }
 
-struct evmc_address address(struct evmc_context* context)
```
axic: Ah this wasn't used anymore?

gumb0: Right.

chfast:
Looks good. But we can drop `account_exists()` now.

axic:
I think it is fine to keep, because the hashing would be an extra unneeded step if `account_exists` is satisfactory.

axic:
We could make it optional saying if it is not implemented, `get_code_hash` can be used instead.

PR #52, opened 2018-08-08
Storage status

chfast:
Extend set_storage() by reporting the storage status.

Surprisingly, this was quite easy and aleth-interpreter is simpler and works without any test failure.

### DIFF
--- a/examples/capi.c
+++ b/examples/capi.c
@@ -45,10 +45,10 @@ static void get_storage(struct evmc_uint256be* result,
     printf("\n");
 }
 
-static void set_storage(struct evmc_context* context,
-                        const struct evmc_address* address,
-                        const struct evmc_uint256be* key,
-                        const struct evmc_uint256be* value)
+static enum evmc_storage_status set_storage(struct evmc_context* context,
+                                            const struct evmc_address* address,
+                                            const struct evmc_uint256be* key,
+                                            const struct evmc_uint256be* value)
 {
     (void)context;
     (void)key;
@@ -56,6 +56,7 @@ static void set_storage(struct evmc_context* context,
     printf("EVM-C: SSTORE @");
     print_address(address);
     printf("\n");
+    return EVMC_STORAGE_UNCHANGED;
 }
 
 static void get_balance(struct evmc_uint256be* result,
--- a/examples/capi.c
+++ b/examples/capi.c
@@ -142,14 +142,12 @@ static void get_block_hash(struct evmc_uint256be* result,
 }
 
 /// EVM log callback.
-///
-/// @note The `evm_log` name is used to avoid conflict with `log()` C function.
-static void evm_log(struct evmc_context* context,
-                    const struct evmc_address* address,
-                    const uint8_t* data,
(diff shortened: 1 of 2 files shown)


examples/capi.c
```diff
@@ -142,14 +142,12 @@ static void get_block_hash(struct evmc_uint256be* result,
...
 /// EVM log callback.
-///
-/// @note The `evm_log` name is used to avoid conflict with `log()` C function.
-static void evm_log(struct evmc_context* context,
-                    const struct evmc_address* address,
-                    const uint8_t* data,
-                    size_t data_size,
-                    const struct evmc_uint256be topics[],
-                    size_t topics_count)
+static void emit_log(struct evmc_context* context,
```
axic: Ack this change.

include/evmc/evmc.h
```diff
@@ -451,21 +451,35 @@ typedef void (*evmc_get_storage_fn)(struct evmc_uint256be* result,
...
- *  @param key      The index of the storage entry.
- *  @param value    The value to be stored.
+ * This callback function is used by an EVM to update the given contract
+ * storage entry.
+ * @param context  The pointer to the Host execution context.
+ *                 @see ::evmc_context.
+ * @param address  The address of the contract.
+ * @param key      The index of the storage entry.
+ * @param value    The value to be stored.
+ * @return         The effect on the storage item.
```
axic: Should have `@see ::evmc_storage_ststu`

include/evmc/evmc.h
```diff
@@ -451,21 +451,35 @@ typedef void (*evmc_get_storage_fn)(struct evmc_uint256be* result,
...
+
+/**
+ * The effect of an attempt to modify a contract storage item.
+ */
+enum evmc_storage_status
+{
+    EVMC_STORAGE_UNCHANGED = 0, /**< The storage item value unchanged. */
+    EVMC_STORAGE_MODIFIED = 1,  /**< The storage item value modified. */
+    EVMC_STORAGE_ADDED = 2,     /**< The storage item added. */
+    EVMC_STORAGE_DELETED = 3,   /**< The storage item deleted. */
```
axic: Does this needs to expose special cases about 0?

Adding explanation above may be good. We have the following cases, please map them to codes:
- zero -> zero (UNCHANGED? can this be MODIFIED?)
- zero -> non-zero (ADDED)
- non-zero -> zero (DELETED)
- non-zero -> non-zero (UNCHANGED/MODIFIED)

examples/capi.c
chfast: The ABI is bumped in this PR already, but that's not needed in this case, because this is a static function in a example.

chfast:
Please review again.

examples/capi.c
axic: Right I thought it was renaming the member, but that was renamed before.

include/evmc/evmc.h
axic: I guess the caller can deduce that bat value == 0 && unchanged.

Issue #20, opened 2018-04-17
Release version 0.1.0 matching ABI v1


chfast:

axic:
Do you suggest that `x.y.z` corresponds to
- x = fixed at 0
- y = matches the ABI version
- z = patch level?

Or we'll come up with two disparate set of versions vs. abi versions?

I was wondering if we should just tag the releases as `abi-1`, `abi-2`, but then question is: does it correspond to the first time the new abi was added? What about patches to it? `abi-1.0`, `abi-1.1`, ... ?

As a result I think we should go with semver x.y.z where x corresponds to the ABI version, that means we are 3.0.0 now.

Which I guess could be reasonable (3 major breaking releases) in the last 3 years evmc existed.

chfast:
I'm not sure. I'd rather map it like this
```
v0.1 - ABI v1
v0.2 - ABI v2
v0.3 - ABI v3
```
Then at some point release v1.0.

Also the ABI version could be `100 * major_version + minor_version`.

axic:
What is a 1.0 release?

I like semver because it fixes this question by removing the vanity aspect of version numbers.

gumb0:
Semver looks like less thinking to decide when we go to 1.0, then 2.0 etc.

gumb0:
BTW if "Net gas metering for SSTORE" EIP makes it into Constaninople, it will require another change I think - a function for VM to request whether storage slot is "dirty"

chfast:
The semver is something that is mentioned a lot but nobody is using it in the end.

axic:
Everybody is using it outside of C++ space 😉

chfast:
For 1.0 I'd like to have all missing changes that are currently planned, i.e. Constantinople support, two-phase initialization (makes Go bindings easier) and Go bindings.

So far I'd like to make 0.x releases following semver option for "development" releases.

I think it would be simpler to use two numbers instead of three in the full version number.

axic:
I'm not sure what Constantinople, etc. has to do with the "magical 1.0".

I think EVMC is a rolling API and I'd strongly suggest to go with the stable version of evmc (it is not development mode anymore), that is we are at `5.0` (ABIv5). I think it makes a lot of sense that the major is matching the ABI version, it removes a lot of ambiguities. (A: I'm using EVMC 1.22. B: Hm, is that the one with ABIv10 or ABIv11?)

chfast:
[...]

PR #54, opened 2018-08-10
Mark version 5.0.0-dev and add bumpversion config

chfast:
Part of #20.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,0 +1,25 @@
+[bumpversion]
+current_version = 5.0.0-dev
+tag = False
+commit = True
+message = EVMC {new_version}
+	
+	Bump version: {current_version} → {new_version}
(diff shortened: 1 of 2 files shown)


.bumpversion.cfg
```diff
@@ -0,0 +1,25 @@
...
+	{major}.{minor}.{patch}
+
+[bumpversion:part:prerel]
+optional_value = rel
+values = 
+	-dev
+	rel
+
+[bumpversion:file:CMakeLists.txt]
+
```
axic: Will this replace the one in cmakelists?

chfast: Yes. It will replace "current_version" with "new_version" using matching described in the [bumpversion] section. For evmc.h I have to add different parsing rules to match only major version number and not to replace all `5` digits in the file.

So this should sync EVMC_ABI_VERSION with the semantic version in CMakeLists.txt. We don't use the semantic version anywhere else.

PR #56, opened 2018-08-10
Add EVMC_TRAP_INSTRUCTION error code

axic:


include/evmc/evmc.h
```diff
@@ -275,6 +275,12 @@ enum evmc_status_code
      */
     EVMC_CONTRACT_VALIDATION_FAILURE = 13,
 
+    /**
+     * A trap instruction has been hit during exection.
+     * This is useful for WebAssembly implementations, where the `unreachable`
+     * instruction and various execution errors are turned into a trap.
+     */
+    EVMC_TRAP_INSTRUCTION = 14,
```
axic: Should this be merged with `EVMC_INVALID_INSTRUCTION`?

chfast: Rather not, because trap can be caused not only by a contract design.

chfast:
I'm not sure this is so great idea, because it only reference to the way the exception was implemented in Ewasm. If you cannot distinguish division by zero (wasm builtin trap condition) from ones caused by contract code explicitly, you might want to use generic `EVMC_FAILURE`.

include/evmc/evmc.h
axic: Decision: use `EVMC_UNREACHABLE_INSTRUCTION` here.

axic: In addition we could have `EVMC_WEBASSEMBLY_TRAP` for the generic trap case as specified by the wasm spec.

axic:
@chfast updated.

PR #57, opened 2018-08-10
Add axic to the emvc.h header

axic:


chfast:
Can it look like this?

axic:
Sure, though being second in `evmc.h` would be more fair.

chfast:
Should be in alphabetic order.

PR #67, opened 2018-08-20
Add evmc_execute helper

axic:


include/evmc/helpers.h
```diff
@@ -41,6 +41,21 @@ static inline int evmc_set_option(struct evmc_instance* instance,
     return 0;
 }
 
+/**
+ * Executes the VM.
```
chfast: Extend documentation to something like `Executes code in the VM`.

axic: Done.

PR #68, opened 2018-08-20
Add evmc_has_compatible_abi helper

axic:
Depends on #67.


axic:
> Why this is useful?

Removing the pointer dereferencing from callers. The instance struct can be opaque from a callers perspective if using these helpers.

include/evmc/helpers.h
```diff
@@ -17,6 +17,14 @@
 
 #include <evmc/evmc.h>
 
+/**
+ * Returns true if the instance has a compatible ABI version.
+ */
+static inline int evmc_has_compatible_abi(struct evmc_instance *instance)
```
axic: Not sure about how to nicely name this function.

chfast: `evmc_is_abi_compatible()`?

axic: Changed.

PR #72, opened 2018-08-20
Go bindings: add missing error codes

axic:


bindings/go/evmc/evmc.go
```diff
@@ -120,7 +126,7 @@ func (err Error) Error() string {
 		return fmt.Sprintf("evmc: unknown internal error (%d)", int32(code))
```
chfast: Maybe remove "unknown" from here as we do not predefine any internal errors at the moment.

axic: Perhaps "specific internal error" ? "user-define internal error" ?

chfast: I'd go with "internal error" only.

bindings/go/evmc/evmc.go
```diff
@@ -110,6 +110,12 @@ func (err Error) Error() string {
 		return "evmc: precompile failure"
 	case C.EVMC_CONTRACT_VALIDATION_FAILURE:
 		return "evmc: contract validation failure"
+	case C.EVMC_ARGUMENT_OUT_RANGE:
```
chfast: Should be `EVMC_ARGUMENT_OUT_OF_RANGE`. See Go CI builds.

axic: Fixed already, just didn't push before the other one is decided.

bindings/go/evmc/evmc.go
axic: Then the other `internal error` should also have a value IIRC.

PR #73, opened 2018-08-20
Add return code to get_block_hash_fn

### LINKED ISSUE #63: Host: Return error code from blockhash()
In case the argument is out of range return `EVMC_ARGUMENT_OUT_OF_RANGE` as proposed in https://github.com/ewasm/hera/pull/336#discussion_r211114443.

axic:
Fixes #63.

### DIFF
--- a/bindings/go/evmc/host.go
+++ b/bindings/go/evmc/host.go
@@ -79,7 +79,7 @@ type HostContext interface {
 	Selfdestruct(addr common.Address, beneficiary common.Address)
 	GetTxContext() (gasPrice common.Hash, origin common.Address, coinbase common.Address, number int64, timestamp int64,
 		gasLimit int64, difficulty common.Hash)
-	GetBlockHash(number int64) common.Hash
+	GetBlockHash(number int64) (common.Hash, error)
 	EmitLog(addr common.Address, topics []common.Hash, data []byte)
 	Call(kind CallKind,
 		destination common.Address, sender common.Address, value *big.Int, input []byte, gas int64, depth int,
@@ -186,11 +186,8 @@ func getBlockHash(pResult *C.struct_evmc_uint256be, pCtx unsafe.Pointer, number 
 	idx := int((*C.struct_extended_context)(pCtx).index)
 	ctx := getHostContext(idx)
 
-	blockhash := ctx.GetBlockHash(number)
-
-	// FIXME: should we instead adjust `ctx.GetBlockHash`?
-	if blockhash == (common.Hash{}) {
-		// All zeroes hash is considered a failure in lookup.
+	blockhash, err := ctx.GetBlockHash(number)
+	if err != nil {
 		return C.int(0)
 	}
 
--- a/bindings/go/evmc/host.go
+++ b/bindings/go/evmc/host.go
@@ -182,11 +182,20 @@ func getTxContext(pCtx unsafe.Pointer) C.struct_evmc_tx_context {
 }
 
 //export getBlockHash
-func getBlockHash(pResult *C.struct_evmc_uint256be, pCtx unsafe.Pointer, number int64) {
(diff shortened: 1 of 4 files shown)


include/evmc/evmc.h
```diff
@@ -163,10 +163,11 @@ typedef void (*evmc_get_tx_context_fn)(struct evmc_tx_context* result,
  *  @param      context  The pointer to the Host execution context.
  *  @param      number   The block number. Must be a value between
  *                       (and including) 0 and 255.
+ *  @return              1 if exists, 0 otherwise.
  */
-typedef void (*evmc_get_block_hash_fn)(struct evmc_uint256be* result,
-                                       struct evmc_context* context,
-                                       int64_t number);
+typedef int (*evmc_get_block_hash_fn)(struct evmc_uint256be* result,
```
chfast: Should it return `evmc_status_code` (https://github.com/ethereum/evmc/blob/master/include/evmc/evmc.h#L191)?

axic: I don't think so, but we could consider having another enum for more regular cases.

Success/failure/invalid args.

axic: @chfast do we want an enum instead?

chfast: I don't know yet. I have the review all similar cases and create the superset of possible error codes. If we end up with up to 4 items we can use it in all cases.

chfast: We can merge it with `int`, but you have to improve documentation of `@result` and `@param number`.

axic: Should we wait for #116 first?

chfast:
Quick analysis in #117, the `int` should be fine here.

Should I take it over?

include/evmc/evmc.h
chfast: The #116 is ready, just needs review.

bindings/go/evmc/host.go
```diff
@@ -182,11 +182,21 @@ func getTxContext(pCtx unsafe.Pointer) C.struct_evmc_tx_context {
...
 
-	*pResult = evmcUint256be(ctx.GetBlockHash(number))
+	blockhash := ctx.GetBlockHash(number)
+
+	*pResult = evmcUint256be(blockhash)
+
+	r := C.int(0)
+	if blockhash != (common.Hash{}) {
+		// All zeroes hash is considered a failure in lookup.
+		// FIXME: should we instead adjust `ctx.GetBlockHash`?
```
axic: I think this is the main question. Leaning towards this.

axic: @chfast if you think that's a better idea, can you please push a new commit which changes it? (just leave the current one as it is)

include/evmc/evmc.h
```diff
@@ -160,10 +160,11 @@ typedef struct evmc_tx_context (*evmc_get_tx_context_fn)(struct evmc_context* co
  *  @param      context  The pointer to the Host execution context.
  *  @param      number   The block number. Must be a value between
  *                       (and including) 0 and 255.
+ *  @return              1 if exists, 0 otherwise.
```
chfast: This comment is wrong. It should be something like `1 if the information is available, 0 otherwise` or something better.
Also:
- The comment for `number` is wrong. In this case leave "The block number" only and explain when the information is available in the main description.
- In the `result` explain that result is written only if the function returns 1 i.e. when information is available.

bindings/go/evmc/host.go
chfast: This should be changed to `blockhash, err := ctx.GetBlockHash(number)`. I can push the change later on.

axic:
> This should be changed to blockhash, err := ctx.GetBlockHash(number). I can push the change later on.

I do the first fix, but you should do this one.

Will let you know when I pushed my fix.

axic:
@chfast updated, you can take it over now.

PR #76, opened 2018-08-21
 Introduce AUTHORS file 

chfast:
Add a file with list of copyright holders known as "Authors".


chfast:
@axic @gumb0 @ehildenb @pirapira,
Tell we if you want to be on the list and in what form.

PR #84, opened 2018-08-26
Add evmc_instance_name/version into helpers

axic:


bindings/go/evmc/evmc.go
```diff
@@ -176,11 +176,11 @@ func (instance *Instance) Destroy() {
 }
 
 func (instance *Instance) Name() string {
-	return C.GoString(instance.handle.name)
+	return C.GoString(C.evmc_instance_name(instance.handle))
```
chfast: Do not change Go bindings. It will have to create a function wrapper for `evmc_instance_name()`.

include/evmc/helpers.h
```diff
@@ -18,13 +18,29 @@
...
  */
 static inline int evmc_is_abi_compatible(struct evmc_instance *instance)
 {
     return instance->abi_version == EVMC_ABI_VERSION;
 }
 
+/**
+ * Returns the name of the VM instance.
+ */
+static inline const char* evmc_instance_name(struct evmc_instance *instance)
```
chfast: Rename it to `evmc_get_vm_name()` or `evmc_vm_get_name()`. The later is better because I want to at least split VM and Host functions with prefixes: `evmc_vm_` and `evmc_host_`.

bindings/go/evmc/evmc.go
axic: Too bad, well at least it is maintained by us so it should be fine, but my goal is that external users of EVMC would go through the helpers to detach themselves from the instance internals.

axic: Btw, it compiles, so you just mean it is slower because Go generates an extra layer of abstraction?

chfast: For every C function Go have to generate a wrapper that will switch context from Go runtime to C runtime.

axic: Do we consider it too big of an overhead for these two rearely used functions?

chfast: At the moment yes. I'd have to inspect the generated code. You can leave a TODO comment.

axic: If you insist, these two methods should be called once per loading, so it shouldn't be a big deal.

axic:
@chfast updated.

PR #85, opened 2018-08-26
Go bindings: move free_result C wrapper closer to where it is used

axic:


axic:
@chfast by the same logic could `evmc_go_fn_table` be moved into `evmc.go`?

axic:
I guess linking Go with inline C is more complex than I thought.

chfast:
The gotcha here is C code in .go files works like a header and is compiled twice (if you export Go functions with `//export` like in host.go). That's why you cannot define a function there unless it is a `static` one.

PR #89, opened 2018-08-27
Add vm test for calling destroy

axic:


axic:
Never mind, the tear down does this already.

Issue #90, opened 2018-08-27
Tests do not compile on MacOS


axic:
```
[ 70%] Building CXX object test/unittests/CMakeFiles/evmc-test.dir/test_instructions.cpp.o
In file included from /Users/alex/Projects/evmc/test/unittests/test_instructions.cpp:5:
In file included from /Users/alex/Projects/evmc/include/evmc/instructions.h:16:
/Users/alex/Projects/evmc/include/evmc/evmc.h:450:29: error: commas at the end of enumerator lists are a C++11 extension [-Werror,-Wc++11-extensions]
    EVMC_STORAGE_DELETED = 3,
                            ^
```

chfast:
what compiler version?

axic:
Right, this is my older Mac:
> Apple LLVM version 8.0.0 (clang-800.0.42.1)

axic:
It actually happens on my up-to-date macos too:
> Apple LLVM version 9.1.0 (clang-902.0.39.2)

chfast:
I've seen it somewhere else, but I cannot reproduce it with clang 6... The fix is easy, but I'd like to add a CI check for this case.

chfast:
This is actually funny, you are building something without C++11. Fix: https://github.com/ethereum/evmc/pull/92.

axic:
I think in general somehow C++(11) is not enabled:
```
[ 70%] Building CXX object test/unittests/CMakeFiles/evmc-test.dir/test_instructions.cpp.o
In file included from /Users/alex/Projects-New/evmc/test/unittests/test_instructions.cpp:5:
In file included from /Users/alex/Projects-New/evmc/include/evmc/instructions.h:16:
/Users/alex/Projects-New/evmc/include/evmc/evmc.h:450:29: error: commas at the end of enumerator lists are a C++11 extension [-Werror,-Wc++11-extensions]
    EVMC_STORAGE_DELETED = 3,
                            ^
In file included from /Users/alex/Projects-New/evmc/test/unittests/test_instructions.cpp:5:
/Users/alex/Projects-New/evmc/include/evmc/instructions.h:176:27: error: commas at the end of enumerator lists are a C++11 extension [-Werror,-Wc++11-extensions]
    OP_SELFDESTRUCT = 0xff,
                          ^
/Users/alex/Projects-New/evmc/test/unittests/test_instructions.cpp:11:11: error: 'auto' type specifier is a C++11 extension [-Werror,-Wc++11-extensions]
[...]

PR #92, opened 2018-08-27
Fix compatibility with C++98

chfast:


axic:
While this doesn't fixes #90, I think it still is useful to compile the headers with C++98 and C++11.

PR #94, opened 2018-08-27
Add vm test for execute

axic:
Unfortunately need to adjust all the test VMs.

### DIFF
--- a/test/vmtester/mock_context.cpp
+++ b/test/vmtester/mock_context.cpp
@@ -1,0 +1,130 @@
+#include <evmc/evmc.h>
+#include <evmc/helpers.h>
+
+#include <inttypes.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+evmc_uint256be balance(evmc_context* context, const evmc_address* address)
+{
+    (void)context;
+    (void)address;
+    evmc_uint256be ret = {{1, 2, 3, 4}};
+    return ret;
+}
+
+static int account_exists(evmc_context* context, const evmc_address* address)
+{
+    (void)context;
+    (void)address;
+    return 0;
+}
+
+static void get_storage(evmc_uint256be* result,
+                        evmc_context* context,
+                        const evmc_address* address,
+                        const evmc_uint256be* key)
+{
+    (void)result;
+    (void)context;
+    (void)address;
+    (void)key;
+}
+
+static enum evmc_storage_status set_storage(evmc_context* context,
+                                            const evmc_address* address,
+                                            const evmc_uint256be* key,
+                                            const evmc_uint256be* value)
+{
+    (void)context;
+    (void)address;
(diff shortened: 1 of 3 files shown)


axic:
Oh, well great, this uses `~/install/bin/evmc-vmtester ~/install/lib/libevmc-examplevm.so` which is passing the tests.

axic:
It is broken now:
```
Build-agent version 0.1.437-d180d4d7 (2018-08-22T16:07:19+0000)
Starting container ethereum/cpp-build-env
  image is cached as ethereum/cpp-build-env, but refreshing...

Error response from daemon: manifest for ethereum/cpp-build-env:latest not found
```

test/vmtester/mock_context.cpp
```diff
@@ -0,0 +1,136 @@
...
+{
+    (void)context;
+    (void)address;
+    (void)data;
+    (void)data_size;
+    (void)topics;
+    (void)topics_count;
+}
+
+extern const struct evmc_context_fn_table mock_context_fn_table = {
```
axic: In the future we could extend this to have a factory which instantiates a context with certain values.

axic:
@chfast this is ready now

test/vmtester/mock_context.cpp
```diff
@@ -0,0 +1,136 @@
+#include <evmc/evmc.h>
+#include <evmc/helpers.h>
+
+#include <inttypes.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+struct evmc_uint256be balance(struct evmc_context* context, const struct evmc_address* address)
```
chfast: You don't need `struct` prefix in C++.

test/vmtester/tests.cpp
```diff
@@ -26,11 +26,35 @@ static constexpr size_t optionalDataSize =
...
     ASSERT_EQ(vm->abi_version, EVMC_ABI_VERSION);
 }
 
+TEST_F(evmc_vm_test, execute)
+{
+    struct evmc_context context = {&mock_context_fn_table};
+    struct evmc_message msg
+    {
+    };
+    std::vector<uint8_t> code({0xfe, 0x00});
```
chfast: `std::array` would go here. Or even a raw array + `std::begin` and `std::size`.

test/vmtester/mock_context.cpp
axic: This is a copy from `examples/capi.c`. Should we keep it as C (with a `.c` extension) or go full in on C++?

axic:
> I'm just thinking about renaming "context" to "host"...

Sounds good. Obviously that is a separate breaking change.

axic:
@chfast updated

chfast:
I think mock_context.c should actually replace capi.c as a host implementation example. But this can be done later.

PR #111, opened 2018-08-29
go: Add Host exported functions type checks

chfast:


bindings/go/evmc/host.c
```diff
@@ -22,7 +35,70 @@ const struct evmc_context_fn_table evmc_go_fn_table = {
...
+    evmc_get_block_hash_fn get_block_hash_fn = NULL;
+    get_block_hash_fn(uint256be, context, number);
+    getBlockHash(uint256be, context, number);
+
+    evmc_emit_log_fn emit_log_fn = NULL;
+    emit_log_fn(context, address, data, size, uint256be, size);
+    emitLog(context, address, data, size, uint256be, size);
+
+    (void)status;
+    (void)storage_status;
```
axic: Are these needed? If so, why not place them after their appropriate last uses?

chfast: Needed to suppress "unused variable" warning.

axic: Yeah but they're assigned too, doesn't that count?

chfast: It does not if you only assign them but never read them.

chfast:
Waits for #113

PR #114, opened 2018-08-30
Return evmc_tx_context in get_tx_context (as opposed to pointer parameter)

### LINKED ISSUE #95: Return evmc_tx_context directly in evmc_get_tx_context_fn
As opposed to taking a pointer to the result as an argument.

axic:
Fixes #95.


chfast:
Similar changes to capi.c and mock_context suggest they should be merged into a single example host. I think I will handle this next.

axic:
Rebased.

PR #116, opened 2018-08-30
Host example

chfast:


test/vmtester/tests.cpp
```diff
@@ -9,6 +9,12 @@
 #include <array>
 #include <cstring>
 
+// Declarations of functions from example host:
+
+evmc_context* example_host_create_context();
+void example_host_destroy_context(evmc_context* context);
```
axic: Why not create a header for it now?

chfast: I don't want it to be a part of en example.

examples/capi.c
```diff
@@ -1,219 +0,0 @@
...
-    printf("EVM-C: LOG%d\n", (int)topics_count);
-}
-
-static const struct evmc_context_fn_table ctx_fn_table = {
-    account_exists, get_storage,  set_storage, get_balance,    get_code_size,  get_code_hash,
-    copy_code,      selfdestruct, call,        get_tx_context, get_block_hash, emit_log,
-};
-
-/// Example how the API is supposed to be used.
-int main()
```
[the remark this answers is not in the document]
chfast: Will do: https://github.com/ethereum/evmc/issues/118
I have idea how to more improve examples, but need to move some files. I'd prefer in following PRs.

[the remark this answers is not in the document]
axic: Would it be hard to keep it in this PR? It should just use the context from `example-host`

chfast:
Done except it does not work now.

chfast:
Needs #119

axic:
This may be nitpicking, but I'd like to have `git mv capi.c example-host.c` in `Replace capi.c exampled with example_host.cpp`, so to avoid the need for the second commit and keep the tiny history of it.

axic:
I meant `git mv capi.c examples.c` and then removing everything except `main` (in the same commit), that should keep the history.

I can do it on a different branch if you want to have a look at it?

chfast:
No, I will do it.

chfast:
Actually I did it in https://github.com/ethereum/evmc/pull/116/commits/ab480f2e0ed0a869d0c8380ae611239f3552b34c. It will not work because changes are too big.

axic:
Tried myself, it seems even if it is a git move, it thinks the diff is too large and just keeps it as delete/create.

chfast:
I can add intermediate commit to have it.

examples/example_host.cpp
```diff
@@ -1,12 +1,15 @@
...
-#include <inttypes.h>
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
+#include "example_host.h"
+
+#include <evmc/helpers.h>
 
-evmc_uint256be balance(evmc_context* context, const evmc_address* address)
+static evmc_uint256be balance(evmc_context* context, const evmc_address* address)
```
axic: Not relevant to this PR, but I guess if this is `.cpp` we should use an anonymous namespace instead?

axic:
> I can add intermediate commit to have it.

Would be nice in order to see the actual changes in `main`, but I'm not pushing full shoulder for it.

examples/example_host.cpp
chfast: I'm trying to stay in C dialect as much as possible.

chfast:
I have it, at least locally: https://github.com/ethereum/evmc/pull/116/commits/98213f93ec4d87e74ad18f8c48934f9937639adc#diff-ca5036d6b23291fb8dcac8244c1bc9a6

For some reason GitHub does not show the history of example.c, but my git does.

PR #119, opened 2018-08-30
Generalize loader searching algorithm

chfast:
Allow loader to shorten the name word by word to the point where only single word is left.

### DIFF
--- a/include/evmc/loader.h
+++ b/include/evmc/loader.h
@@ -60,15 +60,16 @@ enum evmc_loader_error_code
  *   "libexample-interpreter.so",
  * - the "lib" prefix and file extension are stripped from the name:
  *   "example-interpreter"
- * - all "-" are replaced with "_" to construct _full name_:
+ * - all "-" are replaced with "_" to construct _base name_:
  *   "example_interpreter",
- * - the _full name_ is split by "_" char and the last item is taken to form the _short name_:
- *   "interpreter",
- * - the name "evmc_create_" + _full name_ is checked in the library:
+ * - the function name "evmc_create_" + _base name_ is searched in the library:
  *   "evmc_create_example_interpreter",
- * - then, the name "evmc_create_" + _short name_ is checked in the library:
- *   "evmc_create_interpreter".
- * - lastly, the name "evmc_create" is checked in the library
+ * - if function not found, the _base name_ is shorten by skipping the first word separated by "_":
+ *   "interpreter",
+ * - then, the function of the shorter name "evmc_create_" + _base name_ is searched in the library:
+ *   "evmc_create_interpreter",
+ * - the name shortening continues until a function is found or the name cannot be shorten more,
+ * - lastly, when no function found, the function name "evmc_create" is searched in the library.
  *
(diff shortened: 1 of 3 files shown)


include/evmc/loader.h
```diff
@@ -60,15 +60,16 @@ enum evmc_loader_error_code
...
- * - the name "evmc_create_" + _full name_ is checked in the library:
+ * - the function name "evmc_create_" + _base name_ is searched in the library:
  *   "evmc_create_example_interpreter",
- * - then, the name "evmc_create_" + _short name_ is checked in the library:
- *   "evmc_create_interpreter".
- * - lastly, the name "evmc_create" is checked in the library
+ * - if function not found, the _base name_ is shorten by skipping the first word separated by "_":
+ *   "interpreter",
+ * - then, the function of the shorter name "evmc_create_" + _base name_ is searched in the library:
+ *   "evmc_create_interpreter",
```
axic: So what exactly is the logic here? It removes the prefixes? `example` off `example_interpreter`

chfast: yes

chfast: But also `a_b_c_d` -> `b_c_d` -> `c_d` -> `d`.

axic: What is the reasoning for that?

chfast: `evmc-example-vm` :) I want it to come up with function name `evmc_create_example_vm()` while previously it tried only `evmc_create_evmc_example_vm()` and `evmc_create_vm()`.

axic: I think that may not be the most sound reasoning :)

Why not just name the file `example-vm` (and the creator appropriately) and be done with it?

chfast: For 3 reasons:
- I don't want to name the library example-vm in case it is going to be installed.
- This code looks better than before.
- Even the author of loader (me) was no able to remember how exactly it was working before so #116 does not work.

lib/loader/loader.c
```diff
@@ -87,30 +87,28 @@ evmc_create_fn evmc_load(const char* filename, enum evmc_loader_error_code* erro
...
-        }
+        // Shorten the base name by skipping the `word_` segment.
+        const char* shorter_name_pos = strchr(base_name, '_');
+        if (!shorter_name_pos)
+            break;
+
+        memmove(base_name, shorter_name_pos + 1, strlen(shorter_name_pos) + 1);
     }
 
     if (!create_fn)
```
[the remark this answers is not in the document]
chfast: True. I will check it out.

include/evmc/loader.h
chfast: It was, but not fully. It was skipping all leading words except the last one.

lib/loader/loader.c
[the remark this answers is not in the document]
chfast: No, it will get messy. I would need to add another loop terminating condition (like strlen == 0) and we don't want to check for "evmc_create_" but "evmc_create".

Also, the naming convention will change soon and we might want to try "evmc_create" first.

PR #127, opened 2018-09-03
SSTORE net gas metering support

gumb0:
The single "modified dirty" status is enough to implement EIP-1283, see https://github.com/ethereum/aleth/pull/5238

Closes https://github.com/ethereum/evmc/issues/91

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -450,7 +450,12 @@ enum evmc_storage_status
     /**
      * A storage item has been deleted: X -> 0.
      */
-    EVMC_STORAGE_DELETED = 3
+    EVMC_STORAGE_DELETED = 3,
+
+    /**
+     * A storage item has been modified after being modified before: X -> Y -> Z.
+     */
+    EVMC_STORAGE_MODIFIED_DIRTY = 4
 };
 
 


chfast:
Please rebase.

gumb0:
Rebased.

include/evmc/evmc.h
```diff
@@ -450,7 +450,12 @@ enum evmc_storage_status
     /**
      * A storage item has been deleted: X -> 0.
      */
-    EVMC_STORAGE_DELETED = 3
+    EVMC_STORAGE_DELETED = 3,
+
+    /**
+     * A storage item has been modified after being modified before: X -> Y -> Z.
+     */
+    EVMC_STORAGE_MODIFIED_DIRTY = 4
```
axic: Why not place this next to `MODIFIED`? And why not just name it `MODIFIED_AGAIN` ?

axic: cc @gumb0 @chfast

chfast: 1. The `EVMC_STORAGE_MODIFIED_AGAIN` is better name indeed, we could change it. I just wanted if the EIP mentions this case as dirty. We could go even as far as proposing a change to the EIP.

2. In some sense it is closer to `EVMC_STORAGE_MODIFIED` but also describes a case that was historically added later. I was thinking about this but both options are fine for me.

axic: > The EVMC_STORAGE_MODIFIED_AGAIN is better name indeed, we could change it.

How about changing it before ABI6 is released?

> In some sense it is closer to EVMC_STORAGE_MODIFIED but also describes a case that was historically added later. 

I think since this is added a breaking change I'd add it next to it, with the values pushed down.

If this enum would have been around for months I may be on a different opinion, but not in this case.

axic: Created an issue to track this: https://github.com/ethereum/evmc/issues/129

axic:
Also this missed adding the definition to the Go binding.

Issue #123, opened 2018-08-31
message: Remove code hash


chfast:

axic:
Why was this needed/used? Only for caching in evmjit?

chfast:
I think yes. It should be pressed to the execute function directly.

PR #128, opened 2018-09-03
Reorder evmc_message fields

chfast:
Resolves https://github.com/ethereum/evmc/issues/123.
Resolves https://github.com/ethereum/evmc/issues/122.

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -70,6 +70,21 @@ enum evmc_flags
  */
 struct evmc_message
 {
+    /** The kind of the call. For zero-depth calls ::EVMC_CALL SHOULD be used. */
+    enum evmc_call_kind kind;
+
+    /**
+     * Additional flags modifying the call execution behavior.
+     * In the current version the only valid values are ::EVMC_STATIC or 0.
+     */
+    uint32_t flags;
+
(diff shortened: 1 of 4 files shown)


include/evmc/evmc.h
```diff
@@ -70,58 +70,58 @@ enum evmc_flags
  */
 struct evmc_message
 {
+    /** The kind of the call. For zero-depth calls ::EVMC_CALL SHOULD be used. */
```
axic: While this isn't changed in this PR, but I think this comment is invalid, it can be create even for depth 0.

axic:
So didn't #123 say that move `code_hash` from message to `execute`?

chfast:
The `extcodehash` is not useful here because you don't know the address of the code.

axic:
> The extcodehash is not useful here because you don't know the address of the code.

Isn't #123 saying that `execute()` should have a `code_hash` parameter? The code is certainly known there.

chfast:
Yes it says that, but now I prefer to skip it.

axic:
I don't need it myself so OK to remove it.

chfast:
I prefer add features when there is use for them. And execute() together with message is a ton of params anyway.

PR #136, opened 2018-09-05
 Introduce evmc_set_option_result enum

### LINKED ISSUE #102: set_option to have more response codes (success, fail, invalid option, invalid value)
Suggestion:
- 0: success
- 1: failure (general)
- 2: invalid option
- 3: invalid value for option

Perhaps it should return an enum?

axic:
Fixes #102.

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -654,6 +654,15 @@ struct evmc_instance;
  */
 typedef void (*evmc_destroy_fn)(struct evmc_instance* evm);
 
+/**
+ * Possible outcomes of evmc_set_option.
+ */
+enum evmc_set_option_result
+{
+    EVMC_SET_OPTION_SUCCESS = 0,
+    EVMC_SET_OPTION_INVALID_NAME = 1,
+    EVMC_SET_OPTION_INVALID_VALUE = 2
+};
 
 /**
  * Configures the EVM instance.
@@ -666,9 +675,11 @@ typedef void (*evmc_destroy_fn)(struct evmc_instance* evm);
  *  @param evm    The EVM instance to be configured.
  *  @param name   The option name. NULL-terminated string. Cannot be NULL.
  *  @param value  The new option value. NULL-terminated string. Cannot be NULL.
- *  @return       1 if the option set successfully, 0 otherwise.
+ *  @return       The outcome of the operation.
  */
-typedef int (*evmc_set_option_fn)(struct evmc_instance* evm, char const* name, char const* value);
+typedef enum evmc_set_option_result (*evmc_set_option_fn)(struct evmc_instance* evm,
+                                                          char const* name,
+                                                          char const* value);
 
 
 /**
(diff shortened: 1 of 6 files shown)


include/evmc/evmc.h
```diff
@@ -666,9 +675,11 @@ typedef void (*evmc_destroy_fn)(struct evmc_instance* evm);
  *  @param evm    The EVM instance to be configured.
  *  @param name   The option name. NULL-terminated string. Cannot be NULL.
  *  @param value  The new option value. NULL-terminated string. Cannot be NULL.
- *  @return       1 if the option set successfully, 0 otherwise.
+ *  @return       The outcome of the operation, @see ::evmc_set_option_result.
```
chfast: Remove `@see`.

test/vmtester/tests.cpp
```diff
@@ -64,19 +64,19 @@ TEST_F(evmc_vm_test, set_option_unknown)
 {
     if (vm->set_option)
     {
-        int r = vm->set_option(vm, "unknown_option_csk9twq", "v");
-        EXPECT_EQ(r, 0);
+        enum evmc_set_option_result r = vm->set_option(vm, "unknown_option_csk9twq", "v");
+        EXPECT_NE(r, EVMC_SET_OPTION_SUCCESS);
```
chfast: The result should be "invalid name".

test/vmtester/tests.cpp
```diff
@@ -64,19 +64,19 @@ TEST_F(evmc_vm_test, set_option_unknown)
...
         r = vm->set_option(vm, "unknown_option_csk9twq", "x");
-        EXPECT_EQ(r, 0);
+        EXPECT_NE(r, EVMC_SET_OPTION_SUCCESS);
```
chfast: The result should be "invalid name"

test/vmtester/tests.cpp
```diff
@@ -64,19 +64,19 @@ TEST_F(evmc_vm_test, set_option_unknown)
...
 }
 
 TEST_F(evmc_vm_test, set_option_empty_value)
 {
     if (vm->set_option)
     {
-        int r = vm->set_option(vm, "unknown_option_csk9twq", nullptr);
-        EXPECT_EQ(r, 0);
+        enum evmc_set_option_result r = vm->set_option(vm, "unknown_option_csk9twq", nullptr);
+        EXPECT_NE(r, EVMC_SET_OPTION_SUCCESS);
```
chfast: "invalid name"

test/vmtester/tests.cpp
axic: I had those here before, but figured it may better negating the expectation.

axic:
@chfast updated.

axic:
> You can add another unit test like this:

That will not work with Hera. I'd suggest, outside of this PR, that we define a standard option, which must be supported by all VMs (`debug` or `verbose`, with `on` and `off` and that can be tested)

PR #138, opened 2018-09-05
Use bool type instead of int with 1 and 0

### LINKED ISSUE #131: Introduce enum evmc_bool
In https://github.com/ethereum/evmc/blob/master/include/evmc/evmc.h#L164 `evmc_block_hash_fn` returns 1 on success, 0 on failure.

Compared to that, `evmc_status_code` uses the POSIX style 0 for success, anything else for failure.

This kind of indicates an inconsistency. `evmc_block_hash_fn` tries to map a boolean into a uint, that's why.

A solution could be either using `stdbool.h` or creating an enum to replicate it: `evmc_bool`.

### LINKED ISSUE #135: evmc_account_exists_fn should return evmc_bool

chfast:
Resolves #131 
Resolves #135

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -11,8 +11,9 @@
 #ifndef EVMC_H
 #define EVMC_H
 
-#include <stddef.h> /* Definition of size_t. */
-#include <stdint.h> /* Definition of int64_t, uint64_t. */
+#include <stdbool.h> /* Definition of bool, true and false. */
+#include <stddef.h>  /* Definition of size_t. */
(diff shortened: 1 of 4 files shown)


include/evmc/evmc.h
```diff
@@ -11,6 +11,7 @@
 #ifndef EVMC_H
 #define EVMC_H
 
+#include <stdbool.h>
```
axic: Not sure, but could add `Definition of bool, true, false` here :)

chfast: It seemed self explained.

chfast: I added the comment.

PR #140, opened 2018-09-06
Return bool from some Host methods

### LINKED ISSUE #117: Review Host callbacks return types
- [ ] `get_balance()`, `get_code_size()`, `get_code_hash()`.
   This should not return `int` to inform if an account exists. It would make it unnecessary complicated and a VM usually does not care (a contract might, but usually don't).

chfast:
Resolves #117


axic:
It looks good to me.

axic:
> Return bool from evmc_gen_balance_fn()

->

> Return bool from evmc_get_balance_fn()

PR #148, opened 2018-09-06
Move C++ helpers into evmc/helpers.hpp

axic:
I actually implemented the same operators in Hera to simplify the contract preloading. It may just make sense exposing this here.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -100,7 +100,7 @@ WARN_LOGFILE           =
 # Configuration options related to the input files
 #---------------------------------------------------------------------------
 INPUT                  = \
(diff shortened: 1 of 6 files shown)


axic:
@chfast you can take over now.

include/evmc/evmcpp.h
```diff
@@ -0,0 +1,80 @@
...
+
+#include <evmc/evmc.h>
+
+#include <cstring>
+#include <functional>
+
+/// The comparator for std::map<evmc_address, ...>.
+bool operator<(const evmc_address& a, const evmc_address& b)
+{
+    return std::memcmp(a.bytes, b.bytes, sizeof(a)) < 0;
```
axic: Actually this should be `sizeof(a.bytes)`

chfast:
@axic Please review my changes.

Issue #142, opened 2018-09-06
Enforce "debug" or "verbose" option to be supported by all VMs


axic:
From https://github.com/ethereum/evmc/pull/136#issuecomment-419036254:

The `debug` or `verbose` option (pick one) should be required by all VMs. Accepted values should be `on` or `off`.

axic:
@chfast any comments?

chfast:
I don't want to enforce them.

We can specify some "standard" options so if a VM implement any of them it should follow the specification. But I think it's too early for that.

PR #149, opened 2018-09-06
Add test for EVMC_SET_OPTION_INVALID_VALUE

chfast:
Part of #142.

### DIFF
--- a/test/vmtester/tests.cpp
+++ b/test/vmtester/tests.cpp
@@ -60,11 +60,11 @@ TEST_F(evmc_vm_test, execute)
     example_host_destroy_context(context);
 }
 
-TEST_F(evmc_vm_test, set_option_unknown)
+TEST_F(evmc_vm_test, set_option_unknown_name)
 {
     if (vm->set_option)
     {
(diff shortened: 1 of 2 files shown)


test/vmtester/tests.cpp
```diff
@@ -75,11 +75,26 @@ TEST_F(evmc_vm_test, set_option_empty_value)
...
-        enum evmc_set_option_result r = vm->set_option(vm, "unknown_option_csk9twq", nullptr);
+        evmc_set_option_result r = vm->set_option(vm, "unknown_option_csk9twq", nullptr);
         EXPECT_EQ(r, EVMC_SET_OPTION_INVALID_NAME);
     }
 }
 
+TEST_F(evmc_vm_test, set_option_unknown_value)
+{
+    auto r = evmc_set_option(vm, "verbose", "1");
+    if (r == EVMC_SET_OPTION_SUCCESS || r == EVMC_SET_OPTION_INVALID_VALUE)
```
axic: Perhaps add a comment here so that is clear: execute more tests if the `verbose` as an option name is valid.

That being said you may just want to to invert the condition to: `r != EVMC_SET_OPTION_INVALID_NAME`

PR #151, opened 2018-09-06
Update storage methods

### LINKED ISSUE #146: Return bool from get_storage and set_storage
Continuation of #117 

### get_storage

For `get_storage` we can do similarly to `get_code_hash` and return `false` when an account does not exist, but this is very artificial case, because the `get_storage` is only used for the currently executed account. Any VM should not expect `get_storage` to fail.

### set_storage

Similar story as for `get_storage` but worse. The `set_storage` returns `evmc_storage_status`. This one would need to be extended with `EVMC_STORAGE_NO_SUCH_ACCOUNT`.

chfast:
Closes #146

### DIFF
--- a/bindings/go/evmc/host.go
+++ b/bindings/go/evmc/host.go
@@ -72,7 +72,7 @@ func goByteSlice(data *C.uint8_t, size C.size_t) []byte {
 type HostContext interface {
 	AccountExists(addr common.Address) bool
 	GetStorage(addr common.Address, key common.Hash) common.Hash
-	SetStorage(addr common.Address, key common.Hash, value common.Hash) StorageStatus
+	SetStorage(addr common.Address, key common.Hash, value common.Hash) (StorageStatus, error)
 	GetBalance(addr common.Address) (common.Hash, error)
 	GetCodeSize(addr common.Address) (int, error)
 	GetCodeHash(addr common.Address) (common.Hash, error)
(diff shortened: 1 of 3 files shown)


bindings/go/evmc/host.go
```diff
@@ -106,7 +106,11 @@ func getStorage(pResult *C.struct_evmc_uint256be, pCtx unsafe.Pointer, pAddr *C.
 func setStorage(pCtx unsafe.Pointer, pAddr *C.struct_evmc_address, pKey *C.struct_evmc_uint256be, pVal *C.struct_evmc_uint256be) C.enum_evmc_storage_status {
 	idx := int((*C.struct_extended_context)(pCtx).index)
 	ctx := getHostContext(idx)
-	return C.enum_evmc_storage_status(ctx.SetStorage(goAddress(*pAddr), goHash(*pKey), goHash(*pVal)))
+	status, err := ctx.SetStorage(goAddress(*pAddr), goHash(*pKey), goHash(*pVal))
```
axic: Don't care too much, but why not just require `SetStorage` return this code too?

examples/example_host.cpp
```diff
@@ -52,15 +52,19 @@ static bool account_exists(evmc_context* context, const evmc_address* address)
...
 {
-    (void)result;
-    (void)context;
-    (void)address;
-    (void)key;
+    example_host_context* host = static_cast<example_host_context*>(context);
+    auto it = host->accounts.find(*address);
+    if (it != host->accounts.end())
+    {
+        *result = it->second.storage[*key];
```
axic: What happens when the key is not found?

bindings/go/evmc/host.go
chfast: I guess this is more aligned with Go convention to return errors as `error` type and additional output.

examples/example_host.cpp
chfast: It magically inserts the null hash to the map at this position and returns it.

axic: Ah right because it is the index and not `.at`. Damn C++, damn.

chfast: I think this is mostly to allow insertion in form of `m["hello"] = "world"`.

PR #152, opened 2018-09-06
Rename uint256be to bytes32

### LINKED ISSUE #121: Rename evmc_uint256be to evmc_bytes32 to reflect its content better
This of course means for price values `evmc_uint256be` needs to be added back, but `storage_set`/`storage_get`/`hashes` `evmc_bytes32` would be used.

chfast:
Closes #121 

Some things to consider
- Add alias `evmc_hash256` for explicit crypto-hashes.
- Add alias `evmc_bytes32` and `evmc_address` to avoid `struct` prefix.

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -34,19 +34,22 @@ enum
     EVMC_ABI_VERSION = 6
 };
 
+
 /**
- * Big-endian 256-bit integer.
+ * The fixed size array of 32 bytes.
  *
- * 32 bytes of data representing big-endian 256-bit integer. I.e. bytes[0] is
- * the most significant byte, bytes[31] is the least significant byte.
- * This type is used to transfer to/from the VM values interpreted by the user
- * as both 256-bit integers and 256-bit hashes.
+ * 32 bytes of data capable of storing e.g. 256-bit hashes.
  */
-struct evmc_uint256be
+struct evmc_bytes32
 {
-    /** The 32 bytes of the big-endian integer or hash. */
+    /** The 32 bytes. */
     uint8_t bytes[32];
 };
+
+/**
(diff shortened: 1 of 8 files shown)


include/evmc/evmc.h
```diff
@@ -34,20 +34,23 @@ enum
...
 {
-    /** The 32 bytes of the big-endian integer or hash. */
+    /** The 32 bytes. */
     uint8_t bytes[32];
 };
 
+/**
+ * The alias for evmc_bytes32 to represent a big-endian 256-bit integer.
+ */
+typedef struct evmc_bytes32 evmc_uint256be;
```
axic: So I am thinking perhaps we should have `struct __evmc_bytes32 {..}` and then two typedefs (and the same applies to `evmc_address`) so that it is consistent requiring (or not) `struct` in further use cases.

Right now it is weird that one has to use `struct evmc_bytes32 a;` and `evmc_uint256be a;`

axic: Also wonder if we do have the shared underlying type, operators applied to them propagate to the two typedefs?

chfast: These are the same types effectively. Yes, it feels weird to me too.

include/evmc/evmc.h
```diff
@@ -116,27 +119,27 @@ struct evmc_message
...
      *
      * Ignored unless kind is EVMC_CREATE2.
      */
-    struct evmc_uint256be create2_salt;
+    evmc_bytes32 create2_salt;
 };
 
 
 /** The transaction and block data for execution. */
 struct evmc_tx_context
```
gumb0: Maybe typedef this one, too, to write `evmc_tx_context` instead od `struct evmc_tx_context`?

PR #154, opened 2018-09-09
Simplify signatures of Host methods

chfast:
This reverts most of the Host method changes introduced recently.

Should I also change evmc_get_block_hash_fn?

I tried the previous API in Aleth and it is very messy. Here is an example of trying to expose the information if an account exists when checking balance: https://github.com/ethereum/aleth/pull/5261

This information is not useful in any of the use cases (no place where you would not to fold it to 0 balance).

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -413,19 +413,15 @@ typedef bool (*evmc_account_exists_fn)(struct evmc_context* context, const evmc_
  *
  * This callback function is used by a VM to query the given contract storage entry.
  *
- * @param[out] result   The pointer to the place where to put the result value.
- * @param      context  The pointer to the Host execution context.
- * @param      address  The address of the account.
- * @param      key      The index of the account's storage entry.
- * @return              If the account exists the value is put at the location
- *                      pointed by @p result and true is returned.
- *                      If the account does not exist false is returned without
- *                      modifying the memory pointed by @p result.
+ * @param context  The Host execution context.
+ * @param address  The address of the account.
+ * @param key      The index of the account's storage entry.
+ * @return         The storage value at the given storage key or null bytes
+ *                 if the account does not exist.
  */
-typedef bool (*evmc_get_storage_fn)(evmc_bytes32* result,
-                                    struct evmc_context* context,
-                                    const evmc_address* address,
-                                    const evmc_bytes32* key);
+typedef evmc_bytes32 (*evmc_get_storage_fn)(struct evmc_context* context,
+                                            const evmc_address* address,
+                                            const evmc_bytes32* key);
 
 
 /**
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -511,19 +511,12 @@ typedef size_t (*evmc_get_code_size_fn)(struct evmc_context* context, const evmc
  * in the account at the given address. For existing accounts not having a code, this
  * function returns keccak256 hash of empty data.
  *
- * @param[out] result   The pointer to the place where to put the result code hash.
- *                      The pointed memory is only modified when the function returns true.
- *                      The pointer MUST NOT be null.
- * @param      context  The pointer to the Host execution context.
- * @param      address  The address of the account.
- * @return              If the account exists the hash of its code is put at the location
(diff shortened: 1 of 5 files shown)


gumb0:
ABI stays v6?

chfast:
> ABI stays v6?

I think there are still some changes. I will compare later on wit v5.0.0.

include/evmc/evmc.h
```diff
@@ -494,39 +485,24 @@ typedef enum evmc_storage_status (*evmc_set_storage_fn)(struct evmc_context* con
...
- *                      The pointer MUST NOT be null.
- * @param      context  The pointer to the Host execution context.
- * @param      address  The address of the account.
- * @return              If the account exists its balance is put at the location
- *                      pointed by @p result and true is returned.
- *                      If the account does not exist false is returned without
- *                      modifying the memory pointed by @p result.
+ * @param context  The pointer to the Host execution context.
+ * @param address  The address of the account.
+ * @return         The balance of the given storage key or 0 if the account does not exist.
```
gumb0: The balance of the given account

include/evmc/evmc.h
```diff
@@ -535,19 +511,12 @@ typedef bool (*evmc_get_code_size_fn)(size_t* result,
...
- *                      pointed by @p result and true is returned.
- *                      If the account does not exist false is returned without
- *                      modifying the memory pointed by @p result.
- */
-typedef bool (*evmc_get_code_hash_fn)(evmc_bytes32* result,
-                                      struct evmc_context* context,
-                                      const evmc_address* address);
+ * @param context  The pointer to the Host execution context.
+ * @param address  The address of the account.
+ * @return         The hash of the code in the account or null hash if the account does not exist.
```
gumb0: `null bytes` maybe?

gumb0: `evmc_get_storage_fn` comment here above says `null bytes`, so maybe it would be less confusing to say the same here

chfast: After reconsideration I changed it to "null bytes".

chfast:
> ABI stays v6?
> 
> I think there are still some changes. I will compare later on wit v5.0.0.

Oh, you mean it should be v7? No, we bump it after the previous release if breaking changes are planned and it stays 6 until 6.0.0 is released.

include/evmc/evmc.h
gumb0: Still a typo here

chfast: Yes. Funny. It looks like this typo was introduced in some other commit than I thought.

include/evmc/evmc.h
```diff
@@ -413,19 +413,15 @@ typedef bool (*evmc_account_exists_fn)(struct evmc_context* context, const evmc_
...
-                                    struct evmc_context* context,
-                                    const evmc_address* address,
-                                    const evmc_bytes32* key);
+ * @param context  The Host execution context.
+ * @param address  The address of the account.
+ * @param key      The index of the account's storage entry.
+ * @return         The storage value at the given storage key or null bytes
+ *                 if the account does not exist.
+ */
+typedef evmc_bytes32 (*evmc_get_storage_fn)(struct evmc_context* context,
```
axic: I'm not sold this change makes sense.

I can agree with the account lookups, but storage is different.

include/evmc/evmc.h
```diff
@@ -464,12 +460,7 @@ enum evmc_storage_status
     /**
      * A storage item has been deleted: X -> 0.
      */
-    EVMC_STORAGE_DELETED = 4,
-
-    /**
-     * An attempt to modify storage of an non-existing account.
-     */
-    EVMC_STORAGE_NON_EXISTING_ACCOUNT = 5
```
axic: Same here.

include/evmc/evmc.h
chfast: I partly agree, but also this case is exotic at the moment.
I can do the following for now:
- extract this commit to a separate PR for a reference (not to be merge soon),
- add a comment to `evmc_get_storage_fn` and `evmc_set_storage_fn` that the account **MUST** exist.

PR #155, opened 2018-09-09
Simplify get_block_hash() method by returning null hash

chfast:
Continuation of changes in #154.

### DIFF
--- a/bindings/go/evmc/host.c
+++ b/bindings/go/evmc/host.c
@@ -99,8 +99,8 @@ static inline void go_exported_functions_type_checks()
     tx_context = getTxContext(context);
 
     evmc_get_block_hash_fn get_block_hash_fn = NULL;
-    bool_flag = get_block_hash_fn(&bytes32, context, number);
-    bool_flag = getBlockHash(&bytes32, context, number);
+    bytes32 = get_block_hash_fn(context, number);
+    bytes32 = getBlockHash(context, number);
 
     evmc_emit_log_fn emit_log_fn = NULL;
     emit_log_fn(context, address, data, size, &bytes32, size);
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -158,19 +158,16 @@ typedef struct evmc_tx_context (*evmc_get_tx_context_fn)(struct evmc_context* co
 /**
  * Get block hash callback function.
  *
- * This callback function is used by an VM to query the block hash of
- * a given block. If the requested block is not found, then an appropriate
- * result code is returned.
+ * This callback function is used by a VM to query the hash of the header of the given block.
+ * If the information about the requested block is not available, then this is signalled by
+ * returning null bytes.
  *
- * @param[out] result   The returned block hash value. Only written to
- *                      if the return value is 1 (information is available).
- * @param      context  The pointer to the Host execution context.
- * @param      number   The block number.
- * @return              true if the information is available, false otherwise.
+ * @param context  The pointer to the Host execution context.
+ * @param number   The block number.
+ * @return         The block hash or null bytes
+ *                 if the information about the block is not available.
  */
-typedef bool (*evmc_get_block_hash_fn)(evmc_bytes32* result,
-                                       struct evmc_context* context,
-                                       int64_t number);
+typedef evmc_bytes32 (*evmc_get_block_hash_fn)(struct evmc_context* context, int64_t number);
 
 /**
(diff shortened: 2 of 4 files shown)


chfast:
@gumb0 @axic, this one is up to discussion. For me both variants are ok. We can treat this as an experiment.

gumb0:
I'm fine with this, I think it's nice to be consistent with functions in https://github.com/ethereum/evmc/pull/154 and it might simplify calling code in VM.

bindings/go/evmc/host.c
```diff
@@ -41,7 +41,7 @@ static inline void go_exported_functions_type_checks()
 {
     struct evmc_context* context = NULL;
     evmc_address* address = NULL;
-    evmc_bytes32* bytes32 = NULL;
+    evmc_bytes32 bytes32;
```
gumb0: These names `bytes32` & `uint256be` look too much as type names, it confuses me a bit every time.

docs/EVMC.md
```diff
@@ -40,6 +40,7 @@ can be referenced as EVMC ABIv3 or just EVMC 3.
 2. **Host** – An entity controlling the VM. 
    The Host requests code execution and responses to VM queries by callback 
    functions. This usually represents an Ethereum Client.
+3. **null hash** – a hash which all bytes are zeros.
```
gumb0: "for which", "in which" maybe? Or maybe just call it "null bytes" for `get_block_hash`, too

include/evmc/evmc.h
```diff
@@ -158,19 +158,16 @@ typedef struct evmc_tx_context (*evmc_get_tx_context_fn)(struct evmc_context* co
 /**
  * Get block hash callback function.
  *
- * This callback function is used by an VM to query the block hash of
- * a given block. If the requested block is not found, then an appropriate
- * result code is returned.
+ * This callback function is used by a VM to query the hash of the header of the given block.
+ * If the information about the requested block is not available then this is signaled by
```
gumb0: "signalled" and comma before "then"

bindings/go/evmc/host.c
chfast: That's the intention. I have to do hard-core function pointer casts in Go, because cgo (Go's C tool) does not support `const` types. Then whenever I change the EVMC part there is no error in Go bindings.

In this "fake" function I'm using the function pointer before case and the one after with exactly the same arguments and return types so this detect almost all mismatches (because non-const types bind to const ones).

chfast:
This has been rebased. Any comments about the design?

chfast:
# Use cases

1. Contract asks about "recent" block hash (default use case). The VM can pass the request to the Host safely and put the result on the stack.
2. Contract asks about a future block hash. The VM can pass the request to the Host safely and will get the null hash and/or an error.
3. Contract asks about an old block hash. The VM cannot pass the request to the Host because the block might be available but `BLOCKHASH` instructions should put null hash on the stack anyway.

# Possible VM implementations

1. The VM checks if the requested block is in the safe range (256 recent blocks). Only then asks the Host. VM does not expect to get the error / null hash in return. 
- https://github.com/ethereum/aleth/blob/master/libaleth-interpreter/VM.cpp#L1092-L1099
- https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go#L547-L552

2. The VM asks the Host first. When error is received it puts null hash on the stack. If the requests is successful it check it that's not an old block and nulls it if it is.
This variant at least does requests when they are not needed.

# Conclusions

- The first implementation is better.
- In the first implementation is slightly better when the Host returns null hash instead of an error.

```cpp
// Variant with error:
bytes32 hash;
bool success = host->get_block_hash(&hash, number);
assert(success);
if (!success)
    hash = {};  // Make is deterministic.
stack[0] = hash;
``` 

```cpp
// Variant with null hash:
bytes32 hash = host->get_block_hash(number);
assert(hash != {});
stack[0] = hash;
```

axic:
I'm still strictly against this change. Would have preferred not merging it.

axic:
> Contract asks about an old block hash. The VM cannot pass the request to the Host because the block might be available but BLOCKHASH instructions should put null hash on the stack anyway.

This applies to EVM, but nothing else. The main point of the #73 change was that hera wouldn't need to do black magic (checking if `get_block_hash` returns the null hash) to exhibit special behaviour.

chfast:
> in case of an error

What kind of error?

PR #158, opened 2018-09-17
Clarify requirements for get/set storage methods

chfast:
This specifies that the account which storage is to accessed or modified MUST exist.

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -411,7 +411,7 @@ typedef bool (*evmc_account_exists_fn)(struct evmc_context* context, const evmc_
 /**
  * Get storage callback function.
  *
- * This callback function is used by a VM to query the given contract storage entry.
+ * This callback function is used by a VM to query the given account storage entry.
  *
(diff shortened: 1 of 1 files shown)


include/evmc/evmc.h
```diff
@@ -411,7 +411,10 @@ typedef bool (*evmc_account_exists_fn)(struct evmc_context* context, const evmc_
 /**
  * Get storage callback function.
  *
- * This callback function is used by a VM to query the given contract storage entry.
+ * This callback function is used by a VM to query the given account storage entry.
+ * The VM MUST make sure that the account exists. This requirement is only a formality because
```
gumb0: Is it really needed for get_storage? `@return` spec below states that null bytes are returned in case account does not exist (so it's not undefined behavior)

chfast: Yes, you're right.

chfast: Changed.

chfast:
Can you check it now?

PR #164, opened 2018-10-24
Update README

chfast:


README.md
```diff
@@ -30,6 +30,17 @@ Please visit the [documentation].
 - [pyevm] (in progress)
 - [pyethereum] (abandoned)
 
+## Internal
```
axic: We usually put this into a different file. In any case it seems to be belong at the end of this file.

chfast: Yeah, but https://github.com/RichardLitt/standard-readme/blob/master/spec.md#extra-sections.

PR #173, opened 2018-11-24
Load HunterGate if the enclosing project didn't for EVMC_TESTING

### LINKED ISSUE #172: EVMC_TESTING requires Hunter
Can the CMake in evmc install Hunter if it is needed due to testing and it is not provided already?

axic:
Fixes #172.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,15 @@ include(GNUInstallDirs)
 
 if(EVMC_TESTING)
 	include(HunterGate)
+
+	# Load Hunter if the project including EVMC didn't.
+	if(not ${HUNTER_GATE_DONE})
+	    HunterGate(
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -21,6 +21,15 @@ include(GNUInstallDirs)
...
+
+	# Load Hunter if the project including EVMC didn't.
+	if(not ${HUNTER_GATE_DONE})
+	    HunterGate(
+		URL "https://github.com/ruslo/hunter/archive/v0.20.24.tar.gz"
+		SHA1 "3e2037a462bcf2ec3440f60298d933e74ffd4df3"
+	    )
+	endif()
+
 	include(HunterConfig)
```
axic: Weird, because this line supposed to load `cmake/HunterConfig.cmake` which does the exact thing.  Perhaps the problem isn't here, but in Hera's cmake.

PR #178, opened 2019-01-15
C++17

### LINKED ISSUE #176: Update to cable 2.0.16 and also compile for C++14/C++17 on CI
See https://github.com/ethereum/cable/

chfast:
Closes #176

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -3,7 +3,7 @@ jobs:
 
   lint:
     docker:
-    - image: ethereum/cpp-build-env
+    - image: ethereum/cpp-build-env:7
     steps:
     - checkout
     - run:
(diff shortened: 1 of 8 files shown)


circle.yml
```diff
@@ -134,7 +142,8 @@ workflows:
     jobs:
       - lint
       - build
```
axic: Perhaps rename this to `build-cx11` for clarity?

axic: Actually does this mean we're not building in pure C mode on the CI anymore?

axic: Ah ok the toolchain configuration is for `CMAKE_CXX_STANDARD`. Is it worth separately to have `CMAKE_C_STANDARD` and build for C99 and C11? I don't see any worth going below C99 in any case.

axic: Note: this was discussed in #19 and is implemented in the tests.

I think it just confusing that CI tests C++ while some of the tests check C standard. Would be nice to document.

axic: Created an issue to track this.

PR #179, opened 2019-01-15
go: Fix support for CREATE2 salt

chfast:


axic:
Is this a breaking change?

axic:
> Is this a breaking change?

> It's a fix.

I think one could argue for both. Need to define what is the public API, only `evmc.h` or the bindings too?

axic:
To be clear: I think we need to define these rules in the README.

In this instance, because changing the ABI is such a pain, I'd be ok just considering this as a fix :wink:

chfast:
Definitely not an ABI change.
Language bindings probably should be "experimental" first, then "stable".
It's arguable if this is a fix or a feature, but it fixes this PR: https://github.com/ethereum/evmc/pull/133.

PR #190, opened 2019-02-04
Restrict the availability of EVMC_DEPRECATED

chfast:
Define EVMC_DEPRECATED attribute only for GCC 6+ because in older versions attributes cannot be applied to enum elements.


chfast:
Added a changelog entry. @axic you have to reapprove.

Issue #168, opened 2018-11-02
Create Rust binding


axic:
See also https://github.com/ETCDEVTeam/sputnikvm-ffi/issues/12

axic:
See also https://github.com/ETCDEVTeam/sputnikvm-ffi/issues/12

axic:
Ideally the API should be something like:
```
struct MyVM {}

impl VM for MyVM {
  fn execute(instance: Instance, host_context: HostContext, revision: Revision, message: Message, code: &[u8]) -> ExecutionResult {
  }
}

evmc_create_vm!("myvm", "1.0", MyVM)
```

PR #201, opened 2019-03-13
Add basic Rust bindings

axic:
Part of #168.


axic:
Next step is to create a `Result` type which wraps a `u8` array.

This type should:
- a `from` trait from a `u8[]` slice
- a function which returns a C function for destroying the instance (the C signature is `extern fn evmc_release_result(result: *const evmc_result)` where `result` is a pointer to object)

bindings/rust/evmc-sys/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "evmc-sys"
+version = "6.2.0-dev"
+authors = ["Alex Beregszaszi <alex@rtfs.hu>"]
+license = "Apache-2.0"
+repository = "https://github.com/ethereum/evmc"
+description = "Bindings to EVMC (low-level)"
+categories = ["external-ffi-bindings"]
+edition = "2018"
```
chfast: 2019

.gitignore
```diff
@@ -1,3 +1,5 @@
 /build
 /cmake-build-*
 /.idea
+Cargo.lock
+/target
```
chfast: Do you need this top level?

bindings/rust/evmc-vm/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "evmc-vm"
+version = "6.2.0-dev"
+authors = ["Alex Beregszaszi <alex@rtfs.hu>"]
+license = "Apache-2.0"
+repository = "https://github.com/ethereum/evmc"
+description = "Bindings to EVMC (VM specific)"
+edition = "2018"
```
chfast: 2019

.gitignore
axic: If we have the top level `Cargo.toml` then unfortunately it is needed. The reason for having the top level one is the ability to run `cargo build` on the top level.

bindings/rust/evmc-sys/Cargo.toml
axic: This is a Rust specific field. 2018 is valid, 2019 isn't.

.bumpversion.cfg
```diff
@@ -24,3 +24,6 @@ serialize = {major}
 search = EVMC_ABI_VERSION = {current_version}
 replace = EVMC_ABI_VERSION = {new_version}
 
+[bumpversion:file:bindings/rust/evmc-sys/Cargo.toml]
```
chfast: It should contain 
```
search = version = \"{current_version}\"
replace = version = \"{new_version}\"
```
or something...

axic: `bumpversion` seems to work without these

PR #202, opened 2019-03-13
Implement higher-level result handling in EVMC Rust bindings [WIP]

jakelang:
Part of #168.


axic:
I think this should have two use cases:
1) Building a result object for data to be sent back from a VM execution
2) Acquiring a result from a call to be used by the VM execution

For case 1) the VM will have a status code, data, gas left (and if it is a create - an address). I suggest we have a constructor for this case.

For case 2) we have a raw `evmc_result` pointer which we need to consume by wrapped it into an `EvmcResult`. When dropping the `EvmcResult` in this case, the underlying `release` must be called.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -5,3 +5,53 @@ pub use evmc_sys as ffi;
 // TODO: Add convenient helpers for evmc_result
```
axic: Remove this.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -5,3 +5,53 @@ pub use evmc_sys as ffi;
...
 // TODO: Add convenient helpers for evmc_execute
 // TODO: Add a derive macro here for creating evmc_create
+
+//pub fn destroy_result(&mut evmc_result)
```
axic: This too.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -5,3 +5,53 @@ pub use evmc_sys as ffi;
...
+
+/// EVMC result structure.
+pub struct EvmcResult(Box<ffi::evmc_result>);
+
+impl EvmcResult {
+    pub fn new(status: ffi::evmc_status_code, gasleft: i64, output: Option<&[u8]>) -> Self {
```
axic: Please add a rustdoc comment.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -5,3 +5,69 @@ pub use evmc_sys as ffi;
...
+    use std::mem::discriminant;
+    
+    #[test]
+    fn constructor() {
+        let result = EvmcResult::new(ffi::evmc_status_code::EVMC_SUCCESS, 420, Some(&[0xde, 0xad, 0xbe, 0xef]));
+        assert!(result.status() == ffi::evmc_status_code::EVMC_SUCCESS);
+        assert!(result.gasleft() == 420);
+        assert!(result.output().is_some());
+        assert!(result.output().unwrap().len() == 4);
+    }
```
axic: Please check for output content to match the one in the constructor.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -2,6 +2,77 @@ pub extern crate evmc_sys;
...
+            None
+        }
+    }
+
+    pub fn status(&self) -> ffi::evmc_status_code {
+        self.0.status_code
+    }
+}
+
+#[no_mangle]
```
axic: I don't think the `no_mangle` is needed since we pass this around as a function pointer.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -2,6 +2,135 @@ pub extern crate evmc_sys;
...
+            420,
+            Some(&[0xde, 0xad, 0xbe, 0xef]),
+        );
+        let inner = result.into_inner();
+        let back = EvmcResult::from_boxed(inner);
+        assert!(back.status() == ffi::evmc_status_code::EVMC_SUCCESS);
+        assert!(back.gasleft() == 420);
+        assert!(back.output().is_some());
+        assert!(back.output().unwrap().len() == 4);
+        assert!(back.output().unwrap() == [0xde, 0xad, 0xbe, 0xef]);
```
axic: `assert_eq!()`?

jakelang: I think `assert!` is more readable.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -2,6 +2,134 @@ pub extern crate evmc_sys;
...
+
+impl EvmcResult {
+    /// Initialize a new result struct given a status code, remaining gas, and optional output
+    /// data.
+    pub fn new(status: ffi::evmc_status_code, gasleft: i64, output: Option<&[u8]>) -> Self {
+        EvmcResult(Box::new(ffi::evmc_result {
+            status_code: status,
+            gas_left: gasleft,
+            output_data: if let Some(buf) = output {
+                buf.clone().as_ptr()
```
axic: I think `clone().as_ptr()` means the clone can go out of scope and be deallocated:
https://doc.rust-lang.org/std/primitive.slice.html#method.as_ptr

> Returns a raw pointer to the slice's buffer.
>
> The caller must ensure that the slice outlives the pointer this function returns, or else it will end up pointing to garbage.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -2,6 +2,134 @@ pub extern crate evmc_sys;
...
+        }))
+    }
+
+    /// Get gas left.
+    pub fn gasleft(&self) -> i64 {
+        self.0.gas_left
+    }
+
+    /// Get the address of the created contract. Does not return valid data if the transaction was
+    /// not a contract creation.
```
axic: "Does not" -> "may not"

chfast:
If you may, prefix you commits with `rust:`

axic:
Didn't we agree that the `ExecutionResult` instance has the data and address objects as Rust data structures (and not keep it boxed)?

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -2,6 +2,179 @@ pub extern crate evmc_sys;
...
+            create_addr: if result.status_code == ffi::evmc_status_code::EVMC_SUCCESS {
+                // NOTE: This field is only Some if the evmc return code is SUCCESS.
+                Some(result.create_address)
+            } else {
+                None
+            },
+        };
+
+        // Release allocated ffi struct.
+        unsafe {
```
axic: Need to check if `release` is `nullptr`.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -2,6 +2,179 @@ pub extern crate evmc_sys;
...
-// TODO: Add convenient helpers for evmc_result
 // TODO: Add convenient helpers for evmc_execute
 // TODO: Add a derive macro here for creating evmc_create
+
+/// EVMC result structure.
+pub struct ExecutionResult {
+    status: ffi::evmc_status_code,
+    gas: i64,
+    output: Option<Vec<u8>>,
+    release_fn: Option<unsafe extern "C" fn(*const ffi::evmc_result)>,
```
axic: Isn't there type for this generated by the bindgen?

Found it, `evmc_release_result_fn`.

axic: But actually this member is not needed.

bindings/rust/evmc-vm/src/lib.rs
axic: Disregard this, it is for the entire struct.

axic: Actually I think `.release` is optional so this should have a nullptr check.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -2,6 +2,190 @@ pub extern crate evmc_sys;
...
+
+/// Returns a pointer to a heap-allocated evmc_result.
+impl Into<*const ffi::evmc_result> for ExecutionResult {
+    fn into(self) -> *const ffi::evmc_result {
+        let (buffer, len) = if let Some(buf) = self.output {
+            let buf_len = buf.len();
+
+            // Manually allocate heap memory for the new home of the output buffer.
+            let memlayout = std::alloc::Layout::from_size_align(buf_len, 1).expect("Bad layout");
+            let new_buf = unsafe { std::alloc::alloc(memlayout) };
```
axic: Does `release_result` need to manually deallocate this?

jakelang: yes. I am working on it right now.

axic: Isn't it easier to Box it?

jakelang: no, because boxing the pointer to the data will do nothing except kill the pointer. and a box cannot be made over an arbitrary-length buffer.

jakelang:
@axic Is this good for merge?

bindings/rust/evmc-vm/src/lib.rs
axic: From https://doc.rust-lang.org/std/vec/struct.Vec.html#guarantees:
> If a Vec has allocated memory, then the memory it points to is on the heap (as defined by the allocator Rust is configured to use by default), and its pointer points to len initialized, contiguous elements in order (what you would see if you coerced it to a slice), followed by capacity-len logically uninitialized, contiguous elements.

jakelang: `Vec` does not own the data. It owns a pointer to the data on the heap, which is safely allocated and deallocated on instantiation and `drop()` respectively.

You can get a pointer to the buffer this way: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.as_ptr because you can dereference a Vec to a slice.
However, the slice does not own the data. When the Vec goes out of scope, the slice will as well because `drop` will be called on the Vec. There is no easy way to make the Vec surrender ownership of the data.
From the docs:
> The caller must ensure that the slice outlives the pointer this function returns, or else it will end up pointing to garbage.

Perhaps you could use `mem::transmute` wizardry but this is terribly difficult to reason about.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -2,6 +2,195 @@ pub extern crate evmc_sys;
...
-// TODO: Add convenient helpers for evmc_result
 // TODO: Add convenient helpers for evmc_execute
 // TODO: Add a derive macro here for creating evmc_create
+
+/// EVMC result structure.
+pub struct ExecutionResult {
+    status: ffi::evmc_status_code,
+    gas: i64,
+    output: Option<Vec<u8>>,
+    create_addr: ffi::evmc_address,
```
axic: Why not name it `create_address`, both here and the helper?

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -2,6 +2,195 @@ pub extern crate evmc_sys;
...
 
 // TODO: Add helpers for host interface
-// TODO: Add convenient helpers for evmc_result
 // TODO: Add convenient helpers for evmc_execute
 // TODO: Add a derive macro here for creating evmc_create
+
+/// EVMC result structure.
+pub struct ExecutionResult {
+    status: ffi::evmc_status_code,
+    gas: i64,
```
axic: I think `status_code` and `gas_left` is better to reduce deviation from the underlying structure.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -2,6 +2,195 @@ pub extern crate evmc_sys;
...
+            } else {
+                None
+            },
+            create_address: result.create_address,
+        };
+
+        // Release allocated ffi struct.
+        unsafe {
+            if result.release.is_some() {
+                result.release.unwrap()(&result as *const ffi::evmc_result);
```
axic: This is nitpicking, but could format this as
```
if ... {
  unsafe { release ..
```

I'm fine either way.

jakelang: I think rustfmt will complain.

axic: Why? The proposal is that only the release call is unsafe and not the condition. Either way it doesn't matter.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -2,6 +2,195 @@ pub extern crate evmc_sys;
...
+            create_address: self.create_address,
+            padding: [0u8; 4],
+        }))
+    }
+}
+
+/// Callback to pass across FFI, de-allocating the struct.
+extern "C" fn release_result(result: *const ffi::evmc_result) {
+    unsafe {
+        if !(*result).output_data.is_null() {
```
axic: Actually another question: would the code be cleaner if you do `Box::from_raw` first and call dealloc on the boxed object?

bindings/rust/evmc-vm/src/lib.rs
jakelang: oh, i thought this was mentioning the indentation. will do that

chfast:
Add it to CHANGELOG if there is anything to mention. You can push new CHANGELOG entry to master if convenient.

axic:
This code on its own is not that useful yet. Will add it to the changelog once we get #223 finished and merged.

PR #208, opened 2019-03-13
Circle CI: Test only Go 1.9 (min) and 1.12 (latest)

chfast:


chfast:
Added Go 1.9 back (instead of 1.10).

PR #212, opened 2019-03-14
Add Travis CI for Windows

chfast:
This experiments with Travis CI / Windows.
It also fixes the hidden issue with running unit tests on Windows where it was missing the `example_vm.dll` but execution failure was silently ignored by the script.
Also Hunter is upgraded because it contains a fix for Travis CI/Windows environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+branches:
+  only:
+    - master
+    - /^release\/.*$/
+    - travis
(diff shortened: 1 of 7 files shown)


chfast:
The `travis-ci/pr` failed because was evicted by `travis-ci/push`, but it will not happen unless the `travis` branch is used.

PR #217, opened 2019-03-14
cpp: Add evmc::host - wrapper around Host context / interface

chfast:
Usage: https://github.com/chfast/evmone/commit/45ff329d29f0805b079d9a5c8beb1fee65f437db

### DIFF
--- a/include/evmc/evmc.hpp
+++ b/include/evmc/evmc.hpp
@@ -76,4 +76,92 @@ public:
 private:
     evmc_instance* const m_instance = nullptr;
 };
+
+
+/// Wrapper around EVMC host context / host interface.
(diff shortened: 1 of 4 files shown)


include/evmc/evmc.hpp
```diff
@@ -76,4 +76,86 @@ class vm
...
+    }
+
+    result call(const evmc_message& message) noexcept
+    {
+        return result{context->host->call(context, &message)};
+    }
+
+    const evmc_tx_context& get_tx_context() noexcept
+    {
+        if (tx_context.block_timestamp == 0)
```
axic: Should document this hack.

PR #221, opened 2019-03-16
vmtester: Drop CLI11 dependency

### LINKED ISSUE #220: Drop CLI11 dependency
The CLI11 is only used in vmtester to handle single argument. I believe we can handle that without external library.

chfast:
Closes #220.

### DIFF
--- a/test/vmtester/vmtester.cpp
+++ b/test/vmtester/vmtester.cpp
@@ -6,8 +6,6 @@
 
 #include <evmc/loader.h>
 
-#include <CLI/CLI.hpp>
-
 #include <iostream>
 #include <memory>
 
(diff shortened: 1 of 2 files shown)


test/vmtester/vmtester.cpp
```diff
@@ -28,29 +26,131 @@ evmc_instance* get_vm_instance()
...
+
+        if (auto error_code = cli.parse(argc, argv))
+        {
+            std::cerr << cli.errors;
+            return error_code;
+        }
+
+        if (cli.help)
+        {
+            std::cout << "\n" << cli.output;
```
gumb0: Not sure why you need the newline, but you could add it in the parsing code like `if (help) { output = "\n" + output + ...; }`

chfast: New line is nice because GTest outputs the message for --help first.

But I reworked this again, by taking `cout` and `cerr` as arguments.

PR #222, opened 2019-03-19
Include evmc.hpp in compilation test

axic:


axic:
I am merging this since it is relatively harmless.

PR #223, opened 2019-03-19
Rust: add support for create_vm and execute

axic:


axic:
Some notes:
> We've got evmc_instance which must be returned by create; what the ffi really needs is that the pointer returned has the members where evmc_instance expects it, but what most VMs do is they store their private instance details after the evmc_instance in the same continuous memory area.
>
> evmc_instance has basically one major function pointer: execute, which also receives evmc_instance as a parameter as a way to access its context
>
> There is code do to the above, but what's missing is the user extending it with private data. I suppose that must be done with macros. The macro takes a struct type and creates the wrapper struct around it doing all the above

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -128,6 +363,144 @@ extern "C" fn release_result(result: *const ffi::evmc_result) {
...
+
+#[macro_export]
+macro_rules! evmc_create_vm {
+    ($__vm:ident, $__version:expr) => {
+        paste::item! {
+            static [<$__vm _NAME>]: &'static str = stringify!($__vm);
+            static [<$__vm _VERSION>]: &'static str = $__version;
+        }
+
+        paste::item! {
```
axic: Should have here a large comment explaining this will lay out memory in a format acceptable to C (`repr(C)`) and that's why we can include both data structures here.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -128,6 +363,144 @@ extern "C" fn release_result(result: *const ffi::evmc_result) {
...
+//  pub fn set_option();
+//  pub fn set_tracer();
+}
+*/
+
+#[macro_export]
+macro_rules! evmc_create_vm {
+    ($__vm:ident, $__version:expr) => {
+        paste::item! {
+            static [<$__vm _NAME>]: &'static str = stringify!($__vm);
```
axic: Please separate the VM name and the name of the VM struct.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -128,6 +363,144 @@ extern "C" fn release_result(result: *const ffi::evmc_result) {
...
+            static [<$__vm _VERSION>]: &'static str = $__version;
+        }
+
+        paste::item! {
+            #[derive(Clone)]
+            #[repr(C)]
+            pub struct [<$__vm Instance>] {
```
axic: Does this needs to be `pub`?

bindings/rust/evmc-vm/src/lib.rs
axic: I mean the VM name may contain space and this will fail in that case.

bindings/rust/evmc-vm/src/lib.rs
jakelang: ah yeah that's a good point

bindings/rust/evmc-vm/src/lib.rs
jakelang: will do

bindings/rust/evmc-vm/src/lib.rs
jakelang: discussed offline. figuring this out later

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -128,10 +366,171 @@ extern "C" fn release_result(result: *const ffi::evmc_result) {
...
+                assert!(!instance.is_null());
+                let host = unsafe {
+                    InterfaceManager::new(&rev,
+                        &*msg,
+                        &mut *context,
+                        &mut *instance)
+                };
+
+                let instance = unsafe { [<$__vm Instance>]::coerce_from_raw(instance) };
+                let result: ExecutionResult = instance.get_vm().execute(code_ref, &host);
```
axic: Should wrap this in `std::catch_panic`.

axic:
This is passing vmtester now with the macro monstrosity. At least locally.

axic:
@jakelang @chfast how about we split out `InterfaceManager` from this PR and rename it to `ExecutionContext`? It should also have tests on its own.

Then we can finish the core wrapper with more confidence.

axic:
Superseded by #262.

Issue #189, opened 2019-01-25
Create C++ bindings for writing VMs


axic:

chfast:
We can add `evmc_result make_result(...)` helper that handles memory copy, but it is difficult to combine this with existing `evmc::result` class.

PR #226, opened 2019-03-26
cpp: Add Host interface for the Host side

chfast:
This adds the Host part of the C++ API.

The plan was different, to keep it physically separated, but I realized both Host and VM sides of the API implement the same interface... `HostInterface`.

This introduces some naming issues. We have

- `HostInterface` as pure C++ interface
- `HostContext` as wrapper for the host context pointer to be used by VMs, implements `HostInterface`.
- `Host` - the abstract class for Host implementations, implements `HostInterface`.

Suggestions for better names welcome.

Effectively closes https://github.com/ethereum/evmc/issues/189.

The only tunes left for C++ are aliases like `evmc::address = evmc_address`, merging helpers.hpp, maybe add better types for `evmc_address`, `evmc_uint256be`, etc.

### DIFF
--- a/examples/example_host.cpp
+++ b/examples/example_host.cpp
@@ -8,7 +8,7 @@
 
 #include "example_host.h"
 
-#include <evmc/helpers.h>
+#include <evmc/evmc.hpp>
 #include <evmc/helpers.hpp>
 
 #include <map>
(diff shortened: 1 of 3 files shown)


examples/example_host.cpp
```diff
@@ -21,153 +21,122 @@ struct account
...
-    (void)data_size;
-    (void)topics;
-    (void)topics_count;
-}
 
-static const evmc_host_interface interface = {
-    account_exists, get_storage,  set_storage, get_balance,    get_code_size,  get_code_hash,
-    copy_code,      selfdestruct, call,        get_tx_context, get_block_hash, emit_log,
+public:
+    bool account_exists(const evmc_address& addr) noexcept override
```
gumb0: could be `final`

include/evmc/evmc.hpp
```diff
@@ -79,61 +79,105 @@ class vm
...
+                          size_t data_size,
+                          const evmc_bytes32 topics[],
+                          size_t num_topics) noexcept = 0;
+};
+
+
 /// Wrapper around EVMC host context / host interface.
-class host
+///
+/// To be used by VM implementations as better alternative than using ::evmc_context directly.
```
gumb0: alternative to

include/evmc/evmc.hpp
```diff
@@ -160,9 +204,92 @@ class host
...
                   size_t data_size,
                   const evmc_bytes32 topics[],
-                  size_t topics_count) noexcept
+                  size_t topics_count) noexcept final
     {
         context->host->emit_log(context, &address, data, data_size, topics, topics_count);
     }
 };
+
+/// Abstract class to be used by Host implementations. Hides ::evmc_context vtable.
```
gumb0: Documentation here could be better. `evmc_context` doesn't have vtable.

Better explicitly point out, that this class should be derived when implementing Host

examples/example_host.cpp
chfast: Yes, I noticed. Decided to go `override` because more common. I tend to not to use less common C++ features in examples. Although, in this case it might be good suggestion to use `final` here.

PR #230, opened 2019-04-03
loader: Expose DLL load error

### LINKED ISSUE #225: loader: Get error from dlerror()

chfast:
Only works on Linux and macos, otherwise returns null.

Closes #225.

### DIFF
--- a/bindings/go/evmc/evmc.go
+++ b/bindings/go/evmc/evmc.go
@@ -158,7 +158,13 @@ func Load(filename string) (instance *Instance, err error) {
 	case C.EVMC_LOADER_SUCCESS:
 		instance = &Instance{handle}
 	case C.EVMC_LOADER_CANNOT_OPEN:
-		err = fmt.Errorf("evmc loader: cannot open %s", filename)
+		optionalErrMsg := C.evmc_last_error_msg()
+		if optionalErrMsg != nil {
+			msg := C.GoString(optionalErrMsg)
+			err = fmt.Errorf("evmc loader: %s", msg)
+		} else {
+			err = fmt.Errorf("evmc loader: cannot open %s", filename)
+		}
 	case C.EVMC_LOADER_SYMBOL_NOT_FOUND:
 		err = fmt.Errorf("evmc loader: the EVMC create function not found in %s", filename)
 	case C.EVMC_LOADER_INVALID_ARGUMENT:
--- a/include/evmc/loader.h
+++ b/include/evmc/loader.h
@@ -1,6 +1,6 @@
 /* EVMC: Ethereum Client-VM Connector API.
- * Copyright 2018 The EVMC Authors.
- * Licensed under the Apache License, Version 2.0. See the LICENSE file.
+ * Copyright 2019 The EVMC Authors.
+ * Licensed under the Apache License, Version 2.0.
  */
 
 /**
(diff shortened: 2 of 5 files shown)


lib/loader/loader.c
```diff
@@ -130,6 +130,16 @@ evmc_create_fn evmc_load(const char* filename, enum evmc_loader_error_code* erro
     return create_fn;
 }
 
+const char* evmc_cannot_open_error()
```
axic: I'd say this should be named `evmc_last_error_msg()` and document that the returned pointer should never be freed.

This also gives the flexibility to use a static buffer if Windows would return a heap pointer.

include/evmc/loader.h
```diff
@@ -87,6 +87,15 @@ enum evmc_loader_error_code
...
 evmc_create_fn evmc_load(const char* filename, enum evmc_loader_error_code* error_code);
 
+/**
+ * Exposes additional information about ::EVMC_LOADER_CANNOT_OPEN error.
+ *
+ * This function is not thread-safe and not supported on all operating systems.
+ *
+ * @return Error message or NULL if no additional information is available.
+ */
+const char* evmc_cannot_open_error();
```
axic: I'd say this should be named `evmc_last_error_msg()` and document that the returned pointer should never be freed. Also mention that if `null` returned then no granular error message exists and caller should rely on the error code only.

This also gives the flexibility to use a static buffer if Windows would return a heap pointer.

chfast: Ok, make sense.

include/evmc/loader.h
```diff
@@ -103,6 +103,19 @@ evmc_create_fn evmc_load(const char* filename, enum evmc_loader_error_code* erro
...
+/**
+ * Returns the human-readable message describing the most recent error
+ * that occurred in EVMC loading.
+ *
+ * In case any loading function returned ::EVMC_LOADER_SUCCESS this function always returns NULL.
+ * In case of error code other than success returned, this function MAY return the error message.
+ * This function is not thread-safe and not supported on all operating systems.
+ *
+ * @return Error message or NULL if no additional information is available.
+ *         The returned pointed MUST not be freed by the caller.
```
gumb0:
```suggestion
 *         The returned pointer MUST NOT be freed by the caller.
```

include/evmc/loader.h
```diff
@@ -103,6 +103,19 @@ evmc_create_fn evmc_load(const char* filename, enum evmc_loader_error_code* erro
 struct evmc_instance* evmc_load_and_create(const char* filename,
                                            enum evmc_loader_error_code* error_code);
 
+/**
+ * Returns the human-readable message describing the most recent error
+ * that occurred in EVMC loading.
+ *
+ * In case any loading function returned ::EVMC_LOADER_SUCCESS this function always returns NULL.
+ * In case of error code other than success returned, this function MAY return the error message.
+ * This function is not thread-safe and not supported on all operating systems.
```
gumb0: Reading this as a user leaves me confused about whether I can call it on my operating system. Maybe specify that it returns NULL if not supported?

bindings/go/evmc/evmc.go
```diff
@@ -158,7 +158,13 @@ func Load(filename string) (instance *Instance, err error) {
 	case C.EVMC_LOADER_SUCCESS:
 		instance = &Instance{handle}
 	case C.EVMC_LOADER_CANNOT_OPEN:
-		err = fmt.Errorf("evmc loader: cannot open %s", filename)
+		optionalErrMsg := C.evmc_last_error_msg()
+		if optionalErrMsg != nil {
+			msg := C.GoString(optionalErrMsg)
+			err = fmt.Errorf("evmc loader: %s", msg)
```
gumb0: why not include `filename` in this case, maybe it's useful, too?

include/evmc/loader.h
chfast: I will remove the sentence about the operating system.

PR #232, opened 2019-04-17
loader: Add error messages

chfast:


lib/loader/loader.c
```diff
@@ -44,8 +57,30 @@ static void strcpy_s(char* dest, size_t destsz, const char* src)
...
+                                             const char* format,
+                                             ...)
+{
+    va_list args;
+    va_start(args, format);
+    if (vsnprintf(last_error_msg_buffer, LAST_ERROR_MSG_BUFFER_SIZE, format, args) <
+        LAST_ERROR_MSG_BUFFER_SIZE)
+        last_error_msg = last_error_msg_buffer;
+    va_end(args);
+    return error_code;
```
gumb0: what's the point of passing `error_code` here and doing nothing with it, just returning back?

chfast: It's only to make usage of it nicer
```
ec = set_error(EVMC_LOADER_INVALID_ARGUMENT, "invalid argument: file name cannot be empty");
```
Note it's an internal function.

gumb0: Additional step for the code reader to figure out what's happening...

PR #233, opened 2019-04-23
examples: Test Rust VM

axic:
Part of #168.
Depends on #234.

PR #234, opened 2019-04-23
vmtester: do not crash if create_vm returns nullptr

axic:
Split off #233.


test/vmtester/vmtester.cpp
```diff
@@ -16,6 +16,8 @@ evmc_create_fn create_fn;
 std::unique_ptr<evmc_instance, evmc_destroy_fn> create_vm()
 {
     auto vm = create_fn();
+    if (vm == nullptr)
+        return {nullptr, nullptr};
     return {vm, vm->destroy};
```
axic: `vm->destroy` derefs `vm` even if it is `nullptr`.

PR #241, opened 2019-04-24
vmtester: dereference if output_data is present

axic:


test/vmtester/tests.cpp
```diff
@@ -46,6 +46,14 @@ TEST_F(evmc_vm_test, execute)
     {
         EXPECT_EQ(result.output_size, 0);
     }
+    else
+    {
+        EXPECT_NE(result.output_size, 0);
+
+        // Dereference pointer.
+        uint8_t tmp = result.output_data[0];
```
chfast: This is quite fragile check, because compiler is free to optimize it out.

Please check every byte of the output and use the `read()` helper function from https://godbolt.org/z/qlXzbV.

axic: Added.

PR #244, opened 2019-04-24
rust: make create_address optional in ExecutionResult::new()

axic:


bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -18,13 +18,19 @@ impl ExecutionResult {
...
-        _create_address: ffi::evmc_address,
+        _create_address: Option<ffi::evmc_address>,
     ) -> Self {
         ExecutionResult {
             status_code: _status_code,
             gas_left: _gas_left,
             output: _output,
-            create_address: _create_address,
+            create_address: {
+                if let Some(_create_address) = _create_address {
```
axic: Could also do
```
if _create_address.is_some() {
    _create_address.unwrap()
} else {
```

Not sure which is nicer.

jakelang: i think it is fine as it is.

PR #246, opened 2019-04-24
Update copyright year, include the years of files creation

chfast:


axic:
I think it is fine. (As long as changing the date doesn't consume too much time.)

PR #248, opened 2019-04-24
docs: Include documentation for C++ API

chfast:


chfast:
> Also why are some commit messages uppercase?

By file names. Will change to lowercase, which looks nicer.

> Would want to see the generated output.

Docs are always published as an artifact of test-docs CI job. https://6882-66214407-gh.circle-artifacts.com/0/docs/index.html

I learned that you can copy docs from other elements with `@copydoc`. It gives consistency by avoiding documenting some things 3 times. Not ideal yet, but can be further improved. Not sure about experience for uses seeing comment in code though.

PR #252, opened 2019-04-24
cpp: Rename evmc::vm to evmc::VM

### LINKED ISSUE #249: C++: Rename evmc::vm to evmc::VM
In `evmc.hpp`, in namespace `evmc`, rename:
- `vm` to `VM`,
- `result` to `Result`.

chfast:
Closes #249

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning].
   
 ### Changed
 
+- The `evmc::vm` renamed to `evmc::VM` in C++ API.
+  [[#252](https://github.com/ethereum/evmc/pull/252)]
 - Previously deprecated `helpers.hpp` header file has been removed.
   [[#410](https://github.com/ethereum/evmc/pull/410)]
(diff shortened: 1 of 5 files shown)


axic:
Was this introduced in this release or is this a breaking change?

The rest of the C++ interfaces are camel cased so this is consistent with it, with one exception: `result`. Would changing that to `Result` be breaking?

chfast:
Nothing strict here, but have a feeling that types representing OOP classes should have camel case names. The `result` was left because it's more like `unique_ptr` but can be renamed as well.

The `vm` class was added in 6.2. I have to check the `result`.

chfast:
Both were added in 6.1: https://github.com/ethereum/evmc/blob/v6.1.1/include/evmc/evmc.hpp.

axic:
@chfast reviving?

axic:
@chfast will `Result` handled in a separate PR?

chfast:
> @chfast will `Result` handled in a separate PR?

I'm not sure about `result`. It is rather a value type (like `string`).

PR #254, opened 2019-04-24
Rust: Implement ExecutionContext

jakelang:
Part of #168.

### DIFF
--- a/bindings/rust/evmc-vm/src/lib.rs
+++ b/bindings/rust/evmc-vm/src/lib.rs
@@ -6,7 +6,6 @@
 pub extern crate evmc_sys;
 pub use evmc_sys as ffi;
 
-// TODO: Add helpers for host interface
 // TODO: Add convenient helpers for evmc_execute
 // TODO: Add a derive macro here for creating evmc_create
 
@@ -16,6 +15,14 @@ pub struct ExecutionResult {
     gas_left: i64,
     output: Option<Vec<u8>>,
     create_address: ffi::evmc_address,
+}
+
+/// EVMC context structure. Exposes the EVMC host functions, message data, and transaction context
+/// to the executing VM.
+pub struct ExecutionContext<'a> {
+    message: &'a ffi::evmc_message,
+    context: &'a mut ffi::evmc_context,
+    tx_context: ffi::evmc_tx_context,
 }
 
 impl ExecutionResult {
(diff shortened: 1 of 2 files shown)


bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -18,6 +18,13 @@ pub struct ExecutionResult {
     create_address: ffi::evmc_address,
 }
 
+pub struct ExecutionContext<'a> {
+    rev: &'a ffi::evmc_revision,
+    msg: &'a ffi::evmc_message,
+    ctx: &'a mut ffi::evmc_context,
+    inst: &'a mut ffi::evmc_instance,
```
axic: I'd prefer to use long names (or at least not 3-4 character abbreviations).

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -61,6 +68,223 @@ impl ExecutionResult {
...
+        &self.msg.value
+    }
+
+    pub fn create2_salt(&self) -> &ffi::evmc_bytes32 {
+        &self.msg.create2_salt
+    }
+
+    pub fn account_exists(&mut self, addr: &ffi::evmc_address) -> bool {
+        unsafe {
+            assert!(!(*self.ctx.host).account_exists.is_some());
```
axic: I think the assertions should take place in `new`.

However, in the long-run, no such assertions should be made, just catch panics in the macro and return an EVMC failure.

axic:
Please add a commit making use of this in `example-rust-vm` for call kind.

Also please add some unit tests.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -61,6 +68,223 @@ impl ExecutionResult {
...
+                addr as *const ffi::evmc_address,
+                data as *const [u8] as *const u8,
+                data.len(),
+                topics as *const [ffi::evmc_bytes32] as *const ffi::evmc_bytes32,
+                topics.len(),
+            )
+        }
+    }
+
+    pub fn get_instance(&self) -> &ffi::evmc_instance {
```
axic: Is this only needed by the macro stuff? If so I'd remove it from this PR.

bindings/rust/evmc-vm/src/lib.rs
jakelang: i plan to scrap these members anyway in favor of a higher level representation of all the context data. will keep it in mind.

bindings/rust/evmc-vm/src/lib.rs
jakelang: this makes sense.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -61,6 +68,223 @@ impl ExecutionResult {
...
+
+    pub fn destination(&self) -> &ffi::evmc_address {
+        &self.msg.destination
+    }
+
+    pub fn sender(&self) -> &ffi::evmc_address {
+        &self.msg.sender
+    }
+
+    pub fn input_data(&self) -> Option<Vec<u8>> {
```
axic: Can this be made into a slice to the input?

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -61,6 +67,216 @@ impl ExecutionResult {
...
+        &self.message.destination
+    }
+
+    pub fn sender(&self) -> &ffi::evmc_address {
+        &self.message.sender
+    }
+
+    pub fn input_data(&self) -> Option<Vec<u8>> {
+        if !self.message.input_data.is_null() {
+            let mut ret: Vec<u8> = Vec::with_capacity(self.message.input_size);
```
axic: Need to also use `set_len`.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -61,6 +67,219 @@ impl ExecutionResult {
...
+            tx_context: _tx_context,
+        }
+    }
+
+    pub fn call_kind(&self) -> &ffi::evmc_call_kind {
+        &self.message.kind
+    }
+
+    pub fn is_static(&self) -> bool {
+        self.message.flags == 1
```
axic: Not sure we want this helper, but in any case this should use `ffi::EVMC_STATIC` or `ffii::evmc_flags::EVMC_STATIC`.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -18,6 +18,12 @@ pub struct ExecutionResult {
     create_address: ffi::evmc_address,
 }
 
+pub struct ExecutionContext<'a> {
```
axic: Rustdoc?

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -308,4 +480,75 @@ mod tests {
             }
         }
     }
+
+    unsafe extern "C" fn get_dummy_txctx(context: *mut ffi::evmc_context) -> ffi::evmc_tx_context {
```
axic: Any reason for the shorthand `txctx`? :)

jakelang: its a test helper leave me alone :(
will fix it in a bit

axic: Can we name this nicely?

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -308,4 +480,75 @@ mod tests {
...
+            sender: ffi::evmc_address { bytes: [0u8; 20] },
+            input_data: std::ptr::null() as *const u8,
+            input_size: 0,
+            value: ffi::evmc_uint256be { bytes: [0u8; 32] },
+            create2_salt: ffi::evmc_uint256be { bytes: [0u8; 32] },
+        }
+    }
+
+    #[test]
+    fn execution_context() {
```
axic: Please test at last one function, for example `get_code_size`.

axic:
Can we squash some commits or at least name them properly?

bindings/rust/evmc-vm/src/lib.rs
jakelang: Done in a separate test case.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -308,4 +480,99 @@ mod tests {
...
+            block_gas_limit: 105023,
+            block_difficulty: ffi::evmc_uint256be { bytes: [0xaa; 32] },
+        }
+    }
+
+    unsafe extern "C" fn get_dummy_code_size(
+        _context: *mut ffi::evmc_context,
+        _addr: *const ffi::evmc_address,
+    ) -> usize {
+        105023usize
```
axic: Nitpicking but can you use `105023 as usize`?

Never liked these compound forms in C++ either.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -61,6 +69,170 @@ impl ExecutionResult {
...
+        addr: &ffi::evmc_address,
+        data: &[u8],
+        topics: &[ffi::evmc_bytes32],
+    ) {
+        unsafe {
+            assert!((*self.context.host).emit_log.is_some());
+            (*self.context.host).emit_log.unwrap()(
+                self.context as *mut ffi::evmc_context,
+                addr as *const ffi::evmc_address,
+                data as *const [u8] as *const u8,
```
axic: Can you avoid this cast?

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -61,6 +69,170 @@ impl ExecutionResult {
...
+                data.len(),
+                topics as *const [ffi::evmc_bytes32] as *const ffi::evmc_bytes32,
```
axic: Not sure about alignment here and this looks nasty. At the bare minimum with need a FIXME note and explanation.

PR #262, opened 2019-05-03
Rust: Attribute-style procedural macro to generate EVMC FFI code with user-defined data

### LINKED ISSUE #168: Create Rust binding

jakelang:
Closes #168.
Closes #223.

### DIFF
--- a/bindings/rust/evmc-declare-tests/src/lib.rs
+++ b/bindings/rust/evmc-declare-tests/src/lib.rs
@@ -1,0 +1,25 @@
+/* EVMC: Ethereum Client-VM Connector API.
+ * Copyright 2019 The EVMC Authors.
+ * Licensed under the Apache License, Version 2.0.
+ */
+
+use evmc_vm::EvmcVm;
+use evmc_vm::ExecutionContext;
+use evmc_vm::ExecutionResult;
+#[macro_use]
+use evmc_declare::evmc_declare_vm;
+
+#[evmc_declare_vm("Foo VM", "ewasm, evm", "1.42-alpha.gamma.starship")]
+pub struct FooVM {
+    a: i32,
+}
+
+impl EvmcVm for FooVM {
+    fn init() -> Self {
+        FooVM { a: 105023 }
+    }
+
+    fn execute(&self, code: &[u8], context: &ExecutionContext) -> ExecutionResult {
+        ExecutionResult::success(1337, None)
+    }
+}
--- a/bindings/rust/evmc-declare/Cargo.toml
+++ b/bindings/rust/evmc-declare/Cargo.toml
@@ -1,0 +1,27 @@
+# EVMC: Ethereum Client-VM Connector API.
+# Copyright 2019 The EVMC Authors.
+# Licensed under the Apache License, Version 2.0.
+
+[package]
+name = "evmc-declare"
+version = "6.3.0-dev"
+authors = ["Jake Lang <jak3lang@gmail.com>"]
+license = "Apache-2.0"
+repository = "https://github.com/ethereum/evmc"
+description = "Bindings to EVMC (VM declare macro)"
+edition = "2018"
+
+[dependencies]
+quote = "0.6.12"
+heck = "0.3.1"
+proc-macro2 = "0.4.29"
+
+# For documentation examples
+evmc-vm = { path = "../evmc-vm" }
+
+[dependencies.syn]
+version = "0.15.33"
+features = ["full"]
+
+[lib]
+proc-macro = true
--- a/bindings/rust/evmc-declare/src/lib.rs
+++ b/bindings/rust/evmc-declare/src/lib.rs
@@ -1,0 +1,374 @@
+/* EVMC: Ethereum Client-VM Connector API.
+ * Copyright 2019 The EVMC Authors.
+ * Licensed under the Apache License, Version 2.0.
+ */
+
+//! evmc-declare is an attribute-style procedural macro to be used for automatic generation of FFI
+//! code for the EVMC API with minimal boilerplate.
+//!
+//! evmc-declare can be used by applying its attribute to any struct which implements the `EvmcVm`
+//! trait, from the evmc-vm crate.
+//!
+//! The macro takes two or three arguments: a valid UTF-8 stylized VM name, a comma-separated list of
+//! capabilities, and an optional custom version string. If only the name and capabilities are
+//! passed, the version string will default to the crate version.
+//!
+//! # Example
+//! ```
+//! #[evmc_declare::evmc_declare_vm("This is an example VM name", "ewasm, evm", "1.2.3-custom")]
+//! pub struct ExampleVM;
+//!
+//! impl evmc_vm::EvmcVm for ExampleVM {
+//!     fn init() -> Self {
+//!             ExampleVM {}
+//!     }
+//!
+//!     fn execute(&self, code: &[u8], context: &evmc_vm::ExecutionContext) -> evmc_vm::ExecutionResult {
+//!             evmc_vm::ExecutionResult::success(1337, None)
+//!     }
+//! }
+//! ```
+
+// Set a higher recursion limit because parsing certain token trees might fail with the default of 64.
+#![recursion_limit = "128"]
+
+extern crate proc_macro;
+
+use heck::ShoutySnakeCase;
+use heck::SnakeCase;
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::parse_macro_input;
+use syn::spanned::Spanned;
+use syn::AttributeArgs;
+use syn::Ident;
+use syn::IntSuffix;
+use syn::ItemStruct;
+use syn::Lit;
+use syn::LitInt;
+use syn::LitStr;
+use syn::NestedMeta;
+
+struct VMNameSet {
+    type_name: String,
+    name_allcaps: String,
+    name_lowercase: String,
+}
+
+struct VMMetaData {
+    capabilities: u32,
+    // Not included in VMNameSet because it is parsed from the meta-item arguments.
+    name_stylized: String,
+    custom_version: Option<String>,
+}
+
+#[allow(dead_code)]
+impl VMNameSet {
+    fn new(ident: String) -> Self {
+        let caps = ident.to_shouty_snake_case();
+        let lowercase = ident
+            .to_snake_case()
+            .chars()
+            .filter(|c| *c != '_')
+            .collect();
+        VMNameSet {
+            type_name: ident,
+            name_allcaps: caps,
+            name_lowercase: lowercase,
+        }
+    }
+
+    /// Return a reference to the struct name, as a string.
+    fn get_type_name(&self) -> &String {
+        &self.type_name
+    }
+
+    /// Return a reference to the name in shouty snake case.
+    fn get_name_caps(&self) -> &String {
+        &self.name_allcaps
+    }
+
+    /// Return a reference to the name in lowercase, with all underscores removed. (Used for
+    /// symbols like evmc_create_vmname)
+    fn get_name_lowercase(&self) -> &String {
+        &self.name_lowercase
+    }
+
+    /// Get the struct's name as an explicit identifier to be interpolated with quote.
+    fn get_type_as_ident(&self) -> Ident {
+        Ident::new(&self.type_name, self.type_name.span())
+    }
+
+    /// Get the lowercase name appended with arbitrary text as an explicit ident.
+    fn get_lowercase_as_ident_append(&self, suffix: &str) -> Ident {
+        let concat = format!("{}{}", &self.name_lowercase, suffix);
+        Ident::new(&concat, self.name_lowercase.span())
+    }
+
+    /// Get the lowercase name prepended with arbitrary text as an explicit ident.
+    fn get_lowercase_as_ident_prepend(&self, prefix: &str) -> Ident {
+        let concat = format!("{}{}", prefix, &self.name_lowercase);
+        Ident::new(&concat, self.name_lowercase.span())
+    }
+
+    /// Get the lowercase name appended with arbitrary text as an explicit ident.
+    fn get_caps_as_ident_append(&self, suffix: &str) -> Ident {
+        let concat = format!("{}{}", &self.name_allcaps, suffix);
+        Ident::new(&concat, self.name_allcaps.span())
+    }
+}
+
+impl VMMetaData {
+    fn new(args: AttributeArgs) -> Self {
+        assert!(
+            args.len() == 2 || args.len() == 3,
+            "Incorrect number of arguments supplied"
+        );
+
+        let vm_name_meta = &args[0];
+        let vm_capabilities_meta = &args[1];
+
+        let vm_version_meta = if args.len() == 3 {
+            Some(&args[2])
+        } else {
+            None
+        };
+
+        let vm_name_string = match vm_name_meta {
+            NestedMeta::Literal(lit) => {
+                if let Lit::Str(s) = lit {
+                    s.value()
+                } else {
+                    panic!("Literal argument type mismatch")
+                }
+            }
+            _ => panic!("Argument 1 must be a string literal"),
+        };
+
+        let vm_capabilities_string = match vm_capabilities_meta {
+            NestedMeta::Literal(lit) => {
+                if let Lit::Str(s) = lit {
+                    s.value().to_string()
+                } else {
+                    panic!("Literal argument type mismatch")
+                }
+            }
+            _ => panic!("Argument 2 must be a string literal"),
+        };
+
+        // Parse the individual capabilities out of the list and prepare a capabilities flagset.
+        // Prune spaces and underscores here to make a clean comma-separated list.
+        let capabilities_list_pruned: String = vm_capabilities_string
+            .chars()
+            .filter(|c| *c != '_' && *c != ' ')
+            .collect();
+        let capabilities_flags = {
+            let mut ret: u32 = 0;
+            for capability in capabilities_list_pruned.split(",") {
+                match capability {
+                    "ewasm" => ret |= 0x1 << 1,
+                    "evm" => ret |= 0x1,
+                    _ => panic!("Invalid capability specified."),
+                }
+            }
+            ret
+        };
+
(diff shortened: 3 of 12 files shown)


bindings/rust/evmc-declare-tests/src/lib.rs
```diff
@@ -0,0 +1,22 @@
+use evmc_vm::EvmcVm;
+use evmc_vm::ExecutionResult;
+use evmc_vm::ExecutionContext;
+#[macro_use]
+use evmc_declare::evmc_declare_vm;
+
+#[evmc_declare_vm(FooVm = "ewasm")]
```
axic: Is it possible to merge `evmc-declare-tests` into a test section in `evmc-declare`?

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -9,6 +9,11 @@ pub use evmc_sys as ffi;
 // TODO: Add convenient helpers for evmc_execute
 // TODO: Add a derive macro here for creating evmc_create
```
axic: Both of these should be removed.

bindings/rust/evmc-declare-tests/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+}
+
+impl EvmcVm for FooVM {
+    fn init() -> Self {
+        FooVM { 
+            a: 105023,
+        }
+    }
+
+    fn execute(&self, code: &[u8], context: &ExecutionContext) -> ExecutionResult {
```
axic: As a next step perhaps we could use the `Result` pattern, where `Error` is translated into a failure code in `ExecutionResult`. But I wouldn't touch that in this PR.

bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+                (name, capabilities)
+            } else {
+                panic!("Argument passed to evmc_declare_vm is not a name-value item.")
+            }
+        }
+        _ => panic!("Argument passed to evmc_declare_vm is of incorrect type. Expected a name-value item.")
+    };
+    
+    // Get the VM version from the crate version.
+    let vm_version_string = env!("CARGO_PKG_VERSION").to_string();
```
axic: While I like this we need to document it somehow.

bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+    // Extract the name and capabilities from the attribute's key-value item.
+    let (vm_name_stylized, vm_capabilities) = match &meta.nested[0] {
+        NestedMeta::Meta(m) => {
+            // Try to form a string from the identifier if a meta-item was supplied.
+            if let Meta::NameValue(pair) = m {
+                let name = pair.ident.to_string();
+                let capabilities = match pair.lit {
+                    Lit::Str(ref s) => match s.value().as_str() {
+                        "evm1" => 0x1u32,
+                        "ewasm" => 0x1u32 << 1u32,
```
axic: It seems it only allows a single capability currently?

bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+}
+
+/// Generates a definition and impl for a struct which contains the EVMC instance needed by FFI,
+/// and the user-defined VM.
+// TODO: Move this struct and impl into evmc_vm.
+fn build_vm_container() -> proc_macro2::TokenStream {
+    quote! {
+        struct __EvmcContainer<T: ::evmc_vm::EvmcVm + Sized> {
+            instance: ::evmc_sys::evmc_instance,
+            vm: T,
```
axic: So we're still doing the same trick here :wink:

bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+
+    quote! {
+        static #static_name_ident: &'static str = #stylized_name_literal;
+        static #static_version_ident: &'static str = #version_literal;
+    }
+}
+
+/// Generates a definition and impl for a struct which contains the EVMC instance needed by FFI,
+/// and the user-defined VM.
+// TODO: Move this struct and impl into evmc_vm.
```
axic: How would you do that?

jakelang:
It is not macro-parameterized, so we can just stick the entire thing in
evmc-vm's lib.rs.

bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+    }
+}
+
+/// Takes a capabilities flag and builds the evmc_get_capabilities callback.
+fn build_capabilities_fn(name_lowercase: &String, capabilities: u32) -> proc_macro2::TokenStream {
+    // Could avoid using a special name and just use get_capabilities.
+    let concatenated = format!("{}_get_capabilities", name_lowercase);
+    let capabilities_fn_ident = Ident::new(&concatenated, name_lowercase.span());
+    let capabilities_literal =
+        LitInt::new(capabilities as u64, IntSuffix::U32, capabilities.span());
```
axic: Why `u64` when it says `IntSuffix::U32`? Looks weird.

bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+    quote! {
+        extern "C" fn #fn_ident(instance: *mut ::evmc_sys::evmc_instance) {
+            Box::new(__EvmcContainer::from_ffi_pointer::<#type_ident>(instance));
+        }
+    }
+}
+
+/// Takes a capabilities flag and builds the evmc_get_capabilities callback.
+fn build_capabilities_fn(name_lowercase: &String, capabilities: u32) -> proc_macro2::TokenStream {
+    // Could avoid using a special name and just use get_capabilities.
```
axic: Yep, the special name is only required for `create` as that is the public symbol, everything else should be scoped for the file.

The only benefit we get with prefixes is that multiple VMs could be declared in the same file, but not sure we want that.

jakelang:
I switched back in order to get this done more quickly. To do the field
manipulation (i had this done in earlier commits), we also need to be able
to properly initialize arbitrary fields in evmc_create_vm

bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+
+            let code_ref: &[u8] = unsafe {
+                ::std::slice::from_raw_parts(code, code_size);
+            }
+
+            let container = __EvmcContainer::from_ffi_pointer::<#type_name_ident>(instance);
+
+            let result = container.execute(code_ref, &execution_context);
+
+            container.into_ffi_pointer();
```
axic: Not sure why shuffling this one struct around is needed? Also `from_ffi_pointer` seems to create a clone first?

bindings/rust/evmc-declare/src/lib.rs
axic: Also this should use values from `evmc::ffi::EVMC_CAPABILITIES` and not hardcoded ones.

axic:
Also we need to change the examplevm to use this in order to test it.

bindings/rust/evmc-declare/src/lib.rs
jakelang: because the constructor takes a u64, independent of the actual width of the literal we are trying to construct. It will, nonetheless, produce a U32 literal.

bindings/rust/evmc-declare/src/lib.rs
jakelang: Done.

bindings/rust/evmc-declare/src/lib.rs
jakelang: Done.

bindings/rust/evmc-declare/src/lib.rs
jakelang: Allows multiple capabilities.

jakelang: also, using the enum and playing with the discriminants doesn't make this cleaner IMO.

bindings/rust/evmc-declare/src/lib.rs
jakelang: Yeah, not sure where to.

bindings/rust/evmc-declare-tests/src/lib.rs
jakelang: Yeah, later

bindings/rust/evmc-vm/src/lib.rs
jakelang: Done

bindings/rust/evmc-declare-tests/src/lib.rs
jakelang: No, you cannot test a procedural macro in the same crate :(
`proc_macro2` does have some facilities for this but I prefer to test via code expansion, etc.

bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+
+            let code_ref: &[u8] = unsafe {
+                ::std::slice::from_raw_parts(code, code_size)
+            };
+
+            let container = unsafe {
+                ::evmc_vm::EvmcContainer::<#type_name_ident>::from_ffi_pointer(instance)
+            };
+
+            let result = container.execute(code_ref, &execution_context);
```
axic: Should wrap this in panic unwinding and return `EVMC_INTERNAL_ERROR`:
```
panic::catch_unwind(|| {
    container.execute(..)
});
```

axic: Well actually we should do that for the entire `execute` function except the `assert!`s on top.

axic: And potentially for panics caused but this macro should be ensured that result in abort and not unwinding.

bindings/rust/evmc-declare-tests/src/lib.rs
axic: Ok, lets keep the extra crates then.

bindings/rust/evmc-declare-tests/src/lib.rs
axic: Created issue #309 to track this.

bindings/rust/evmc-declare/src/lib.rs
axic: Was there an answer to this or the code is gone (haven't checked)?

axic:
>  NOTE: Removed the evmc_sys as ffi alias. May reintroduce later if I make a workaround.

Why? Can it be done in a separate PR then?

axic:
Have you tested that the bumpversion config works?

bindings/rust/evmc-declare/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "evmc-declare"
```
axic: Should this be called declare or derive?

bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -0,0 +1,310 @@
+#![recursion_limit = "128"]
```
axic: Why is this needed?

bindings/rust/evmc-declare/src/lib.rs
axic: Created issue #310 to track this.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -4,30 +4,35 @@
  */
 
 pub extern crate evmc_sys;
-pub use evmc_sys as ffi;
```
axic: This seems to be a trend among `-sys` crates, that's the reason I did it.

examples/example-rust-vm/src/lib.rs
```diff
@@ -2,62 +2,25 @@ extern crate evmc_vm;
...
-    code_size: usize,
-) -> ffi::evmc_result {
-    let execution_ctx = unsafe {
-        ExecutionContext::new(
-            msg.as_ref().expect("tester passed nullptr as message"),
-            context.as_mut().expect("tester passed nullptr as context"),
-        )
-    };
-    let is_create = execution_ctx.get_message().kind == ffi::evmc_call_kind::EVMC_CREATE;
+#[evmc_declare_vm("ExampleRustVM", "evm")]
```
axic: Is `evm` the version here? If so, we should use a usual version string because otherwise it is confusing, e.g. "0.1.0".

jakelang: that's the capabilities set. It is a `", "` -separated list of capabilities, which at the moment includes `"ewasm"` and `"ewasm"`. I need to document somewhere visible.

bindings/rust/evmc-vm/src/lib.rs
jakelang: I will re-introduce it in a bit. I just needed to get rid of it for the moment to organize the visibility layout for the procedural macro.

bindings/rust/evmc-declare/src/lib.rs
jakelang: The default recursion limit is `64`, which at the moment has not proven viable for parsing some structures in this macro.

bindings/rust/evmc-declare/Cargo.toml
jakelang: This isn't a derive macro. It is an attribute macro. `derive`-mode macros auto-impl a trait, while this macro does not. It generates FFI code for calling the EVMC VM to which it is applied.

jakelang:
> 
> 
> Have you tested that the bumpversion config works?

How do I test this?

bindings/rust/evmc-declare/src/lib.rs
jakelang: I'd be cautious about trying to introduce unwinding infrastructure this early. Maybe in a separate PR.

examples/example-rust-vm/src/lib.rs
jakelang: The version is the crate version. I may add an option to specify a crate-independent version, as I have figured out how to make this possible.

bindings/rust/evmc-vm/src/lib.rs
axic: I don't fully understand why it needs to be removed now and be readded later introducing a lot of changes across the way?

examples/example-rust-vm/src/lib.rs
axic: Well these should be rustdoc documented in the `evmc-declare` crate as well as with a comment in this example file.

bindings/rust/evmc-declare/src/lib.rs
axic: Can you leave this comment above the setting?

bindings/rust/evmc-declare/Cargo.toml
axic: Wasn't this a derive macro initially?

bindings/rust/evmc-declare/src/lib.rs
jakelang: It is not cloned anymore.

bindings/rust/evmc-declare-tests/src/lib.rs
```diff
@@ -0,0 +1,20 @@
+use evmc_vm::EvmcVm;
+use evmc_vm::ExecutionContext;
+use evmc_vm::ExecutionResult;
+#[macro_use]
+use evmc_declare::evmc_declare_vm;
+
+#[evmc_declare_vm("Foo VM", "ewasm")]
```
axic: Since this is a comprehensive test, I'd say we need to VMs declared here:
1) supports single capability and takes version from crate
2) supports multiple capabilities and takes version from declare macro

bindings/rust/evmc-declare-tests/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+    a: i32,
+}
+
+impl EvmcVm for FooVM {
+    fn init() -> Self {
+        FooVM { a: 105023 }
+    }
+
+    fn execute(&self, code: &[u8], context: &ExecutionContext) -> ExecutionResult {
+        ExecutionResult::new(evmc_sys::evmc_status_code::EVMC_SUCCESS, 235117, None)
```
axic: Could just use `ExecutionResult::success()` here.

bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -0,0 +1,383 @@
+//! evmc-declare is an attribute-style procedural macro to be used for automatic generation of FFI
+//! code for the EVM-C API with minimal boilerplate.
```
axic: EVMC, not EVM-C

bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+    quote! {
+        static #static_name_ident: &'static str = #stylized_name_literal;
+        static #static_version_ident: &'static str = #version_tokens;
+    }
+}
+
+/// Takes a capabilities flag and builds the evmc_get_capabilities callback.
+fn build_capabilities_fn(capabilities: u32) -> proc_macro2::TokenStream {
+    let capabilities_literal =
+        LitInt::new(capabilities as u64, IntSuffix::U32, capabilities.span());
```
axic: Clippy gives this:
```
warning: casting u32 to u64 may become silently lossy if types change
   --> bindings/rust/evmc-declare/src/lib.rs:281:21
    |
281 |         LitInt::new(capabilities as u64, IntSuffix::U32, capabilities.span());
    |                     ^^^^^^^^^^^^^^^^^^^ help: try: `u64::from(capabilities)`
```

jakelang: This is expected. This is a problem with `syn` because it takes a u64 for any integer literal type. In our case, there should be no issue.

axic: Is `u64::from` nicer?

Issue #259, opened 2019-04-29
EVMC precompiles


chfast:
### Precompiles capability, Host driven

1. Extend EVMC capabilities set with "precompiles".
2. Assume [EIP-1352](https://eips.ethereum.org/EIPS/eip-1352) is ready.
3. When a client is to execute a call within precomiles address range, it sends it to the "precompiles" VM.

chfast:
Currently there is no "full" way to bypass Host when executing precompiles because they might be executed with `CALL` having vaue and EVMC does not provide a way to modify account balances.

chfast:
Informational EIP proposed: https://github.com/ethereum/EIPs/pull/2003.

PR #267, opened 2019-05-08
 Add experimental PRECOMPILES capability

chfast:
Part of #259.

### DIFF
--- a/examples/example_precompiles_vm/example_precompiles_vm.cpp
+++ b/examples/example_precompiles_vm/example_precompiles_vm.cpp
@@ -1,0 +1,127 @@
+/* EVMC: Ethereum Client-VM Connector API.
+ * Copyright 2019 The EVMC Authors.
+ * Licensed under the Apache License, Version 2.0.
+ */
+
+#include <evmc/evmc.h>
+#include <evmc/utils.h>
+#include <algorithm>
+
+static evmc_result execute_identity(const evmc_message* msg)
+{
+    auto result = evmc_result{};
+
+    // Check the gas cost.
+    auto gas_cost = 15 + 3 * ((int64_t(msg->input_size) + 31) / 32);
+    auto gas_left = msg->gas - gas_cost;
+    if (gas_left < 0)
+    {
+        result.status_code = EVMC_OUT_OF_GAS;
+        return result;
+    }
+
+    // Execute.
+    auto data = new uint8_t[msg->input_size];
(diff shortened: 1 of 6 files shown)


chfast:
> writing this example in C++

Any opinions? How does it look now in C?

test/vmtester/tests.cpp
```diff
@@ -104,12 +104,6 @@ TEST_F(evmc_vm_test, execute_create)
         read_buffer(result.output_data, result.output_size);
     }
 
-    if (result.status_code == EVMC_SUCCESS)
-    {
-        // This assumes that CREATE returns a non-zero address on success.
```
axic: Why remove this?

chfast: This is wrong test. VM will not set the create address in the result. It didn't cause problems before because the `result.status_code == EVMC_SUCCESS` is never true for EVM not Ewasm.

axic: > VM will not set the create address in the result. 

~~What do you mean by this?~~

Ah right, we need to make this more explicit in the documentation.

chfast: I will pull the commit to another PR with documentation updated.

chfast:
Squashed. Do you think we should merge this?

include/evmc/evmc.h
```diff
@@ -823,7 +823,15 @@ enum evmc_capabilities
     /**
      * The VM is capable of executing ewasm bytecode.
      */
-    EVMC_CAPABILITY_EWASM = (1u << 1)
+    EVMC_CAPABILITY_EWASM = (1u << 1),
+
+    /**
+     * The VM is capable of executing the precompiled contracts
```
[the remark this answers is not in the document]
axic: Can you add the reference to EIP-1352 in evmc.h?

examples/example_precompiles_vm/example_precompiles_vm.cpp
```diff
@@ -0,0 +1,113 @@
+/* EVMC: Ethereum Client-VM Connector API.
+ * Copyright 2019 The EVMC Authors.
+ * Licensed under the Apache License, Version 2.0.
+ */
+
+#include <evmc/evmc.h>
+#include <evmc/utils.h>
+#include <algorithm>
+
+static evmc_result execute_identity(const evmc_message* msg)
```
chfast: Oh shit, I forgot to compute the gas cost.

PR #273, opened 2019-05-08
test: Include helpers.h in compilation tests

axic:


chfast:
I'd like `evmc.h` to compile in every reasonable standard. Can you use some predefined macros tricks and not included `helpers.h` in C90 compilation?

axic:
I've actually added the helper you've requested, just forgot to push.

PR #274, opened 2019-05-09
Add C90 support for helpers.h

axic:
Fixes #273.

### DIFF
--- a/include/evmc/helpers.h
+++ b/include/evmc/helpers.h
@@ -19,18 +19,11 @@
 
 #include <evmc/evmc.h>
 
+#if defined(__cplusplus) || __STD_VERSION__ >= 199901L
+#define EVMC_INLINE inline
+#else
 #define EVMC_INLINE
-#ifdef __cplusplus
-#  undef EVMC_INLINE
-#  define EVMC_INLINE inline
 #endif
-#if __STDC__
-#  if __STDC_VERSION__ >= 199901L
-#    undef EVMC_INLINE
-#    define EVMC_INLINE inline
-#  endif
-#endif
-
 
 /**
  * Returns true if the VM instance has a compatible ABI version.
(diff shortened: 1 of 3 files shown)


include/evmc/helpers.h
```diff
@@ -19,26 +19,39 @@
 
 #include <evmc/evmc.h>
 
+#define EVMC_INLINE
+#ifdef __cplusplus
+#  undef EVMC_INLINE
+#  define EVMC_INLINE inline
+#endif
+#if __STDC__
+#  if __STDC_VERSION__ >= 199901L
```
axic: Could also merge these into `if defined(__cplusplus) || __STD_VERSION__ >= 199901L` if we assume the preprocessor for C90-only compiler is clever enough...

chfast: This would work, because if `__STDC_VERSION__` is undefined it will be substituted with `0`.

axic: By the clever enough I meant that it supports the `defined()` expression as well as the `or` operator`.

axic:
@chfast simplified detection, will squash the last commit into the first if we are going to merge this.

chfast:
This is not useful at all and your are wasting your time.

axic:
> This is not useful at all and your are wasting your time.

Thanks! I don't get the point why you are saying EVMC in C90 compatible when it isn't :D

I've suggested in the previous PR that we state it is not C90 compatible.

chfast:
> I've suggested in the previous PR that we state it is not C90 compatible.

Sorry, I missed that.
The `evmc.h` is. The rest is not as not being the core part of it. But we can remove the `C90` claim. The 20+ year old C standard does not matter.

axic:
> The evmc.h is. The rest is not as not being the core part of it. But we can remove the C90 claim. The 20+ year old C standard does not matter.

I'd propose the README/docs just says "limited support for C90", the core header (evmc.h) is kept compatible, but that is not guaranteed for anything else.

PR #275, opened 2019-05-09
Circle CI: Test for configuration validity of bumpversion

axic:
@chfast not sure this is useful, please feel free to close.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -16,6 +16,14 @@ jobs:
         name: "Run codespell"
         command: |
           codespell --quiet-level=4 --ignore-words=./.codespell-whitelist
+    - run:
+        name: "Check bumpversion"
+        command: |
+          export PATH="/home/builder/.local/bin:$PATH"
+          pip3 install bumpversion
+          bumpversion --dry-run --verbose major
+          bumpversion --dry-run --verbose minor
+          bumpversion --dry-run --verbose patch
 
   build: &build
     docker:


axic:
Also apparently `bumpversion` is not maintained, but a fork is: https://github.com/c4urself/bump2version

chfast:
I noticed. The bumpversion from Ubuntu is newer from the one from the original repo. And it has additional features like signed annotated tags.

circle.yml
```diff
@@ -17,6 +17,21 @@ jobs:
...
+    docker:
+    docker:
+    - image: ethereum/cpp-build-env:10
+    steps:
+    - checkout
+    - run:
+        name: "Check bumpversion"
+        command: |
+          pip3 install bumpversion
+          export PATH="/home/builder/.local/bin:$PATH"
```
chfast: Do it before `pip3 install`, otherwise `pip3` will complain.

circle.yml
```diff
@@ -17,6 +17,21 @@ jobs:
...
+  bumpversion:
+    docker:
+    docker:
+    - image: ethereum/cpp-build-env:10
+    steps:
+    - checkout
+    - run:
+        name: "Check bumpversion"
+        command: |
+          pip3 install bumpversion
```
chfast: Don't you want `pip3 install --user`?

circle.yml
```diff
@@ -17,6 +17,21 @@ jobs:
         command: |
           codespell --quiet-level=4 --ignore-words=./.codespell-whitelist
 
+  bumpversion:
```
chfast: Add it as a step in `lint` job because starting the `ethereum/cpp-build-env` container is expensive.

circle.yml
axic: Changed.

circle.yml
axic: Dunno, can try.

circle.yml
axic: Changed.

circle.yml
axic: That uses the same path, still needs the `export PATH`.

axic:
@chfast ready to be merged

PR #281, opened 2019-05-14
cpp: Fix evmc::result::raw()

chfast:


include/evmc/evmc.hpp
```diff
@@ -56,8 +56,8 @@ class result : private evmc_result
     /// releases the ownership of the resources and invalidates this object.
     evmc_result raw() noexcept
     {
-        auto out = evmc_result{};
-        std::swap<evmc_result>(out, *this);
+        const auto out = evmc_result{*this};  // Copy data.
```
axic: Why is the `const` needed here?

test/unittests/test_cpp.cpp
```diff
@@ -120,3 +124,35 @@ TEST(cpp, host_call)
...
+        raw_result.status_code = EVMC_INTERNAL_ERROR;
+        raw_result.release = release_fn;
+
+        auto raii_result = evmc::result{raw_result};
+        raii_result.gas_left = -1;
+
+        auto raw_result2 = raii_result.raw();
+        EXPECT_EQ(raw_result2.status_code, EVMC_INTERNAL_ERROR);
+        EXPECT_EQ(raw_result2.gas_left, -1);
+        EXPECT_EQ(raw_result.release, release_fn);
```
axic: Should `result2` have `release_fn` and `result` have `nullptr`? Can you check for both here since they are ought to differ?

test/unittests/test_cpp.cpp
```diff
@@ -120,3 +124,35 @@ TEST(cpp, host_call)
...
+    EXPECT_FALSE(release_called);
+
+    {
+        auto raw_result = evmc_result{};
+        raw_result.status_code = EVMC_INTERNAL_ERROR;
+        raw_result.release = release_fn;
+
+        auto raii_result = evmc::result{raw_result};
+        EXPECT_EQ(raii_result.status_code, EVMC_INTERNAL_ERROR);
+    }
```
axic: Can you check for the value of `release` here?

include/evmc/evmc.hpp
chfast: It is not needed, just good practice (I believe).

test/unittests/test_cpp.cpp
chfast: I cannot, because it's `private`.

test/unittests/test_cpp.cpp
chfast: Both will have the same `release` function. The check added.

test/unittests/test_cpp.cpp
```diff
@@ -120,3 +124,36 @@ TEST(cpp, host_call)
...
+    static auto release_called = false;
+    release_called = false;
+    auto release_fn = [](const evmc_result*) noexcept { release_called = true; };
+
+    {
+        auto raw_result = evmc_result{};
+        raw_result.status_code = EVMC_INTERNAL_ERROR;
+        raw_result.release = release_fn;
+
+        auto raii_result = evmc::result{raw_result};
```
axic: Actually another question. When we wrap it here, shouldn't it set `raw_result.release` to `nullptr`?

chfast: That might be good idea, but requires changing constructor from `const evmc_result&` to `evmc_result&&`. That requires at least another PR.

test/unittests/test_cpp.cpp
```diff
@@ -120,3 +124,36 @@ TEST(cpp, host_call)
 
     example_host_destroy_context(host_context);
 }
+
+TEST(cpp, result_raii)
+{
+    static auto release_called = false;
```
gumb0: Do you need it to be static? Why not local and capture it in lambda?

test/unittests/test_cpp.cpp
axic: Tracked in #285.

test/unittests/test_cpp.cpp
axic: Right, this is the C++ object.

test/unittests/test_cpp.cpp
chfast: Because the `.release` field is coming from a C struct that only uses pointers to functions. For convenience, we use the fact you can convert non-capturing lambdas to function pointers.

test/unittests/test_cpp.cpp
```diff
@@ -120,3 +124,36 @@ TEST(cpp, host_call)
...
+        auto raw_result = evmc_result{};
+        raw_result.status_code = EVMC_INTERNAL_ERROR;
+        raw_result.release = release_fn;
+
+        auto raii_result = evmc::result{raw_result};
+        raii_result.gas_left = -1;
+
+        auto raw_result2 = raii_result.raw();
+        EXPECT_EQ(raw_result2.status_code, EVMC_INTERNAL_ERROR);
+        EXPECT_EQ(raw_result2.gas_left, -1);
```
axic: For completeness can you also check for `gas_left` above for `raw_result` that it equals to `0`?

test/unittests/test_cpp.cpp
```diff
@@ -120,3 +124,36 @@ TEST(cpp, host_call)
...
+        EXPECT_EQ(raw_result2.release, release_fn);
+        EXPECT_EQ(raw_result.release, release_fn);
+    }
+    EXPECT_FALSE(release_called);
```
gumb0: You could also make it a counter and check that it's called exactly once

test/unittests/test_cpp.cpp
chfast: According to my spec, it is undefined what the value of `raw_result.gas_left` would be. Here it will be still `-1` because we only zero the `.release`.

chfast: I was considering zeroing it, but there are other pitfalls related to using "move-like" designs that we cannot easily eliminate.

axic: It is implicitly zero by memory initialization, isn't it? I'm talking about this test case here.

chfast:
> Why is this merging into `release/6.2` and not `master`?

Because it's a fix for 6.2.1.

axic:
> > Why is this merging into release/6.2 and not master?
>
> Because it's a fix for 6.2.1.

Will it be merged into `master` afterwards?

chfast:
> Will it be merged into `master` afterwards?

Yes.

PR #282, opened 2019-05-14
cpp: Fix evmc::result's move assignment operator

chfast:


include/evmc/evmc.hpp
```diff
@@ -39,25 +39,29 @@ class result : private evmc_result
     /// Move constructor.
     result(result&& other) noexcept : evmc_result{other}
     {
-        // Disable releaser of the rvalue object.
-        other.release = nullptr;
+        other.release = nullptr;  // Disable releasing of the rvalue object.
```
gumb0: Should you do also `other.output_data = nullptr` to avoid dangling other?

chfast: I was considering zeroing the `other`, but using an object that has been moved is an error anyway. I decided not to because there are other cases which will be hard to cover. Fortunately, some compilers warn you about using an object after `std::move()`.

gumb0: For `raw()` there are no language rules though, and after calling `raw()` this contains dangling pointer

chfast: There is a number of things we can do about it.
1. This is already documented.
2. The function is rather internal, needed to the Host interface. We might hide it by making it private and friend with the Host.
3. We can zero the `other` in `raw()`. It might be better than the current but still not protecting against using the object after `raw()` (it might actually hide the problem deeper) and that will set the `.status_code` to `EVMC_SUCCESS`.

gumb0: You could zero out everything but `status_code`.

Anyway I don't think it's super-critical, having it documented might be enough.
(maybe I would name it `release()` to be similar to smart pointers)

chfast: > `release()`

I wanted that, but unfortunately `evmc_result` already has the `release` method. I wander how to rename it to make the usage more obvious. `release_raw()`? `move_raw()`? `take_raw()`?

gumb0: `release_raw` sounds better to me than `raw`

PR #283, opened 2019-05-14
tests: add test case for precompile implementation 

axic:
Part of #259.

### DIFF
--- a/examples/example_precompiles_vm/example_precompiles_vm.cpp
+++ b/examples/example_precompiles_vm/example_precompiles_vm.cpp
@@ -44,7 +44,7 @@ static evmc_result execute_empty(const evmc_message* msg)
 static evmc_result not_implemented()
 {
     auto result = evmc_result{};
-    result.status_code = EVMC_INTERNAL_ERROR;
+    result.status_code = EVMC_REJECTED;
     return result;
 }
 
--- a/test/vmtester/tests.cpp
+++ b/test/vmtester/tests.cpp
@@ -168,3 +168,53 @@ TEST_F(evmc_vm_test, set_tracer)
     if (vm->set_tracer)
         vm->set_tracer(vm, tracer_callback, nullptr);
 }
+
+TEST_F(evmc_vm_test, precompile_test)
+{
+    // This logic is based on and should match the description in EIP-2003.
+
+    if (!evmc_vm_has_capability(vm, EVMC_CAPABILITY_PRECOMPILES))
+        return;
+
+    // Iterate every address (as per EIP-1352)
+    for (size_t i = 0; i < 0xffff; i++)
+    {
+        auto destination = evmc_address{};
+        destination.bytes[18] = static_cast<uint8_t>(i >> 8);
+        destination.bytes[19] = static_cast<uint8_t>(i & 0xff);
+
+        evmc_context* context = example_host_create_context();
+        evmc_message msg{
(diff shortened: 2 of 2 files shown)


examples/example_precompiles_vm/example_precompiles_vm.cpp
```diff
@@ -44,7 +44,7 @@ static evmc_result execute_empty(const evmc_message* msg)
 static evmc_result not_implemented()
 {
     auto result = evmc_result{};
-    result.status_code = EVMC_INTERNAL_ERROR;
+    result.status_code = EVMC_REJECTED;
```
axic: We have this special case for non-supported code/message. I think it fits better.

Also we have the specialised `EVMC_PRECOMPILE_FAILURE` which we don't use in this example.

test/vmtester/tests.cpp
```diff
@@ -168,3 +168,67 @@ TEST_F(evmc_vm_test, set_tracer)
...
+                                  0,
+                                  0,
+                                  0,
+                                  0,
+                                  0,
+                                  static_cast<uint8_t>(i >> 8),
+                                  static_cast<uint8_t>(i & 0xff)}};
+        evmc_context* context = example_host_create_context();
+        evmc_message msg{
+            EVMC_CREATE,   0, 0, 65536, destination, evmc_address{}, NULL, 0, evmc_uint256be{},
```
axic: Should change this to `CALL`.

test/vmtester/tests.cpp
```diff
@@ -168,3 +168,67 @@ TEST_F(evmc_vm_test, set_tracer)
...
+            evmc_bytes32{}};
+
+        evmc_result result = vm->execute(vm, context, EVMC_MAX_REVISION, &msg, nullptr, 0);
```
axic: Should even set `context` to `nullptr` according to the EIP.

axic:
@chfast can I get a review?

test/vmtester/tests.cpp
chfast: Yes.

test/vmtester/tests.cpp
chfast: Yes.

test/vmtester/tests.cpp
```diff
@@ -168,3 +168,67 @@ TEST_F(evmc_vm_test, set_tracer)
...
+
+TEST_F(evmc_vm_test, precompile_test)
+{
+    if (!evmc_vm_has_capability(vm, EVMC_CAPABILITY_PRECOMPILES))
+        return;
+
+    // Iterate every address (as per EIP-1352)
+    for (size_t i = 0; i < 0xffff; i++)
+    {
+        evmc_address destination{{0,
```
chfast: Looks like giraffe. Maybe write it like:
```
auto destination = evmc_address{};
destination.bytes[18] = static_cast<uint8_t>(i >> 8);
...
```

axic: Its not a giraffe, but this guy:
image

test/vmtester/tests.cpp
axic: Actually it may be needed to check `callvalue`.

chfast: Both are available in the message.

axic: Oh right, that's great.

axic:
I think now this should be in line with https://github.com/ethereum/EIPs/pull/2003

PR #287, opened 2019-05-14
rust: remove support for create_address in ExecutionResult::new

axic:


bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -303,7 +292,11 @@ impl Into<*const ffi::evmc_result> for ExecutionResult {
             output_data: buffer,
             output_size: len,
             release: Some(release_heap_result),
-            create_address: self.create_address,
+            create_address: if self.create_address.is_some() {
```
jakelang: Just a cosmetic comment: I think this can be done more cleanly with an `if let` pattern.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -327,7 +320,11 @@ impl Into<ffi::evmc_result> for ExecutionResult {
             output_data: buffer,
             output_size: len,
             release: Some(release_stack_result),
-            create_address: self.create_address,
+            create_address: if self.create_address.is_some() {
```
jakelang: Same here.

bindings/rust/evmc-vm/src/lib.rs
axic: How is the `if let` pattern any clearer?

axic:
> I initially avoided this because it could possibly introduce some weird behavior that isn't necessarily reflected in the low-level bindings. 

We'd need it if it would be generic, but realised we name the create as `evmc-vm` so it is the VM side and not the client side. The VM will never fill out that field, only the clint -- hence why change makes sense.

bindings/rust/evmc-vm/src/lib.rs
jakelang: personal opinion.

PR #288, opened 2019-05-14
 cpp: Change result constructor to take rvalue reference 

### LINKED ISSUE #285: C++: Allow only moving evmc_result to evmc::result
Change constructor from `result(const evmc_result&)` to `result(evmc_result&&)`. That would indicate that the new `result` object takes the responsibility of the result.

See https://github.com/ethereum/evmc/pull/281#pullrequestreview-237242775.

chfast:
Closes #285
Changelog entry is missing.

### DIFF
--- a/include/evmc/evmc.hpp
+++ b/include/evmc/evmc.hpp
@@ -27,7 +27,17 @@ public:
     using evmc_result::status_code;
 
     /// Converting constructor from raw evmc_result.
-    explicit result(evmc_result const& res) noexcept : evmc_result{res} {}
+    ///
+    /// The constructed object takes ownership and is responsible for releasing
+    /// the provided result. Therefore, the rvalue reference is used to
+    /// indicate the ownership transfer and that
+    /// the raw evmc_result object MUST NOT be used after.
+    ///
+    /// @param raw_result  The rvalue reference to the evmc_result object.
+    result(evmc_result&& raw_result) noexcept : evmc_result{raw_result}  // NOLINT
+    {
+        raw_result.release = nullptr;
+    }
 
     /// Destructor responsible for automatically releasing attached resources.
     ~result() noexcept
--- a/examples/example_host.cpp
+++ b/examples/example_host.cpp
@@ -102,7 +102,7 @@ public:
         res.output_size = msg.input_size;
         res.output_data = output;
         res.release = [](const evmc_result* r) noexcept { delete[] r->output_data; };
-        return evmc::result{res};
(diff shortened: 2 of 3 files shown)


include/evmc/evmc.hpp
```diff
@@ -27,7 +27,17 @@ class result : private evmc_result
...
     /// Converting constructor from raw evmc_result.
-    explicit result(evmc_result const& res) noexcept : evmc_result{res} {}
+    ///
+    /// The constructed object takes ownership and is responsible for releasing
+    /// the provided result. Therefore, the rvalue reference is used to
+    /// indicate the ownership transfer and that
+    /// the raw evmc_result object MUST NOT be used after.
+    ///
+    /// @param raw_result  The rvalue reference to the evmc_result object.
+    result(evmc_result&& raw_result) noexcept : evmc_result{raw_result}  // NOLINT
```
axic: Isn't this in conflict with the move constructor below now?

chfast: No because this is not the move constructor per definition.

axic: Ah it was late I suppose and confused `evmc_result` with `result`.

include/evmc/evmc.hpp
```diff
@@ -27,7 +27,17 @@ class result : private evmc_result
     using evmc_result::status_code;
 
     /// Converting constructor from raw evmc_result.
-    explicit result(evmc_result const& res) noexcept : evmc_result{res} {}
```
axic: Do we consider this as a breaking change?

chfast: It is. Can be moved to 6.3 but as we are fixing C++ API in 6.2 anyway I think we can also include this change.

axic:
clang fails:
```
/home/builder/project/examples/example_host.cpp:105:16: error: no viable conversion from returned value of type 'evmc_result' to function return type 'evmc::result'
        return res;
               ^~~
```

include/evmc/evmc.hpp
axic: I don't think this is fixing a bug in 6.2.x, but improving the API. Following that logic it should be added to 6.3.x, if I'm completely pragmatic.

chfast:
Closing this because I don't like it and in `std::unique_ptr` you can use the pattern we want to prevent here:
```
auto raw = new X{};
auto p = std::unique_ptr{raw};
// Do not: delete raw;
```

PR #293, opened 2019-05-15
cpp: Rename result::raw() to result::release_raw()

chfast:


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## [6.2.2] - unreleased
 
+- Changed: [[#293](https://github.com/ethereum/evmc/pull/293)]
+  In C++ API `evmc::result::raw()` renamed to `evmc::result::release_raw()`.
```
axic: Is this breaking?

axic: I think it is, but the comment should at least be merged into 6.2

chfast: It's breaking, but 6.2 has broken C++ API anyway. Besides, this method is used internally.

axic: Just checked, `raw` was introduced in 6.2.1 via #257. I don't think renaming it is fixing a bug, the bug was fixed before the renaming, wasn't it?

chfast:
Based on `master`.

axic:
`acquire_raw` ?

chfast:
Renamed to `acquire_raw()`.

gumb0:
But "acquire" is the opposite of what happens. Maybe it could look fine if it were a free function like `res = acquire_raw(evmc_res)`, but `evmc_res.acquire_raw()` look like `evmc_result` acquires something.

PR #297, opened 2019-05-22
rust: add failure() and success() helpers to ExecutionResult

axic:
This handles the two common cases for VMs.

### DIFF
--- a/bindings/rust/evmc-vm/src/lib.rs
+++ b/bindings/rust/evmc-vm/src/lib.rs
@@ -39,6 +39,14 @@ impl ExecutionResult {
         }
     }
 
+    pub fn failure() -> Self {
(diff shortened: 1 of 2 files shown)


bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -39,6 +39,14 @@ impl ExecutionResult {
         }
     }
 
+    pub fn failure() -> Self {
```
jakelang: Can you document these?

axic: Can you please?

axic: There's a lot to document. I suggest we create a bigger PR which documents the entire file.

axic: Created issue #308.

Issue #265, opened 2019-05-06
Add get_capabilities() method to EVMC C++


chfast:

axic:
In C we have `has_capabilities` helper: https://github.com/ethereum/evmc/blob/master/include/evmc/helpers.h#L51

Maybe this one should follow the same?

PR #301, opened 2019-05-29
 cpp: Add vm::get_capabilities() method 

### LINKED ISSUE #141: Add VM capabilities support
```
enum evmc_capabilities {
  EVMC_CAPABILITY_EVM1 = 1,
  EVMC_CAPABILITY_EWASM = 2,
}

// set_option may invalidate this, have to be called after set_options have been executed
// this must be always defined in evmc_instance, cannot be null
int get_capabilities()
```

Addresses https://github.com/ewasm/testnet/issues/75 and https://github.com/ewasm/tests/issues/87.

cc @gballet @chfast

chfast:
Closes https://github.com/ethereum/evmc/issues/265.

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -974,12 +974,12 @@ struct evmc_instance
     evmc_execute_fn execute;
 
     /**
-     * Pointer to function returning capabilities supported by the VM instance.
+     * A method returning capabilities supported by the VM instance.
      *
-     * The value returned might change when different options are requested via set_option.
+     * The value returned MAY change when different options are set via the set_option() method.
      *
-     * A Client SHOULD only rely on the value returned here if it has queried it after
-     * it has called set_option.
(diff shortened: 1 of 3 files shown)


include/evmc/evmc.h
```diff
@@ -974,12 +974,12 @@ struct evmc_instance
     evmc_execute_fn execute;
 
     /**
-     * Pointer to function returning capabilities supported by the VM instance.
+     * A method returning capabilities supported by the VM instance.
      *
-     * The value returned might change when different options are requested via set_option.
+     * The value returned MAY change when different options are set via the set_option() method.
```
axic: Is there a way we could add a test for this? Does it make sense to add a test for this?

chfast: We could add behavior like that to example_vm but there is nothing to test here about conformance. Anyway, I'm going to remove this "feature" in EVMC 7.

Issue #198, opened 2019-03-12
Standardize config string 


chfast:
In geth we propose to invoke EVMC module as
```
--vm.evm=./path/to/vm.so,option1=value1,option2=value2
```

Processing of `./path/to/vm.so,option1=value1,option2=value2` can be moved to the loader library.

PR #313, opened 2019-06-05
loader: Standarize EVMC module config

chfast:
Closes https://github.com/ethereum/evmc/issues/198.

### DIFF
--- a/lib/loader/loader.c
+++ b/lib/loader/loader.c
@@ -231,3 +231,93 @@ exit:
 
     return instance;
 }
+
+/// Gets the token delimited by @p delim character of the string pointed by the @p str_ptr.
+/// If the delimiter is not found, the whole string is returned.
+/// The @p str_ptr is also slided after the delimiter or to the string end
+/// if the delimiter is not found (in this case the @p str_ptr points to an empty string).
+static char* get_token(char** str_ptr, char delim)
+{
+    char* str = *str_ptr;
+    char* delim_pos = strchr(str, delim);
+    if (delim_pos)
+    {
+        // If the delimiter is found, null it to get null-terminated prefix
+        // and slide the str_ptr after the delimiter.
+        *delim_pos = '\0';
+        *str_ptr = delim_pos + 1;
+    }
+    else
+    {
+        // Otherwise, slide the str_ptr to the end and return the whole string as the prefix.
+        *str_ptr += strlen(str);
+    }
+    return str;
+}
+
+struct evmc_instance* evmc_load_and_configure(const char* config,
+                                              enum evmc_loader_error_code* error_code)
+{
+    enum evmc_loader_error_code ec = EVMC_LOADER_SUCCESS;
+    struct evmc_instance* instance = NULL;
+
+    char config_copy_buffer[PATH_MAX_LENGTH];
+    if (strcpy_sx(config_copy_buffer, sizeof(config_copy_buffer), config) != 0)
+    {
+        ec = set_error(EVMC_LOADER_INVALID_ARGUMENT,
+                       "invalid argument: configuration is too long (maximum allowed length is %d)",
+                       (int)sizeof(config_copy_buffer));
+        goto exit;
+    }
+
+    char* options = config_copy_buffer;
+    const char* path = get_token(&options, ',');
+
+    instance = evmc_load_and_create(path, error_code);
+    if (!instance)
+        return NULL;
+
+    if (instance->set_option == NULL && strlen(options) != 0)
+    {
+        ec = set_error(EVMC_LOADER_INVALID_OPTION_NAME, "%s (%s) does not support any options",
+                       instance->name, path);
+        goto exit;
+    }
+
+
+    while (strlen(options) != 0)
+    {
+        char* option = get_token(&options, ',');
+
+        // Slit option into name and value by taking the name token.
+        // The option variable will have the value, can be empty.
+        const char* name = get_token(&option, '=');
+
+        enum evmc_set_option_result r = instance->set_option(instance, name, option);
+        switch (r)
+        {
+        case EVMC_SET_OPTION_SUCCESS:
+            break;
+        case EVMC_SET_OPTION_INVALID_NAME:
+            ec = set_error(EVMC_LOADER_INVALID_OPTION_NAME, "%s (%s): unknown option '%s'",
+                           instance->name, path, name);
+            goto exit;
+        case EVMC_SET_OPTION_INVALID_VALUE:
+            ec = set_error(EVMC_LOADER_INVALID_OPTION_VALUE,
+                           "%s (%s): unsupported value '%s' for option '%s'", instance->name, path,
+                           option, name);
+            goto exit;
+        }
+    }
+
+exit:
+    if (error_code)
+        *error_code = ec;
+
+    if (ec == EVMC_LOADER_SUCCESS)
+        return instance;
+
+    if (instance)
+        evmc_destroy(instance);
+    return NULL;
+}
--- a/lib/loader/loader.c
+++ b/lib/loader/loader.c
@@ -42,20 +42,14 @@
 #define ATTR_FORMAT(...)
 #endif
 
-#if _WIN32
-#define strcpy_sx strcpy_s
-#else
 /*
  * Limited variant of strcpy_s().
- *
- * Provided for C standard libraries where strcpy_s() is not available.
- * The availability check might need to adjusted for other C standard library implementations.
  */
 #if !defined(EVMC_LOADER_MOCK)
 static
 #endif
     int
-    strcpy_sx(char* restrict dest, size_t destsz, const char* restrict src)
+    strcpy_sx(char* dest, size_t destsz, const char* src)
 {
     size_t len = strlen(src);
     if (len >= destsz)
@@ -70,7 +64,6 @@ static
     dest[len] = 0;
     return 0;
 }
-#endif
 
 #define PATH_MAX_LENGTH 4096
 
--- a/test/unittests/test_loader.cpp
+++ b/test/unittests/test_loader.cpp
@@ -16,12 +16,8 @@ static constexpr bool is_windows = false;
 #endif
 
 extern "C" {
-#if _WIN32
-#define strcpy_sx strcpy_s
-#else
 /// Declaration of internal function defined in loader.c.
 int strcpy_sx(char* dest, size_t destsz, const char* src);
-#endif
 
 /// The library path expected by mocked evmc_test_load_library().
 extern const char* evmc_test_library_path;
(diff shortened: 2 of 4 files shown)


lib/loader/loader.c
```diff
@@ -238,3 +231,90 @@ struct evmc_instance* evmc_load_and_create(const char* filename,
...
+            *opt_sep = '\0';
+            options = opt_sep + 1;
+        }
+        else
+            options = "";
+
+
+        char* kv_sep = strchr(option, '=');
+        const char* name = option;
+        char* value = NULL;
```
axic: `evmc.h` currently mandates that `value` cannot be NULL.

lib/loader/loader.c
```diff
@@ -283,7 +283,7 @@ struct evmc_instance* evmc_load_and_configure(const char* config,
 
         char* kv_sep = strchr(option, '=');
         const char* name = option;
-        char* value = NULL;
+        char* value = "";  // Set value to empty to handle the "...,option,..." case.
```
axic: Do you want to create an issue so this is changed in the next ABI version?

chfast: After considered this, I like it this way. Not passing null pointers to VMs is nice. Otherwise, all VM implementations would need to check for null values. Also, I don't want to encourage the `,option=,` being interpreted differently from `,option,`. Also, you can see that the number of test cases has decreased.

I will squash the two "loader:" commits, but put it separately for now because consequences are quite interesting.

lib/loader/loader.c
```diff
@@ -42,20 +42,14 @@
 #define ATTR_FORMAT(...)
 #endif
 
-#if _WIN32
```
axic: Why is this change mixed in here?

lib/loader/loader.c
```diff
@@ -238,3 +231,90 @@ struct evmc_instance* evmc_load_and_create(const char* filename,
...
+
+    if (instance->set_option == NULL && strlen(options) != 0)
+    {
+        ec = set_error(EVMC_LOADER_INVALID_OPTION_NAME, "%s (%s) does not support any options",
+                       instance->name, path);
+        goto exit;
+    }
+
+
+    while (strlen(options) != 0)  // TODO: Can be optimized as *options != 0 because MSVC sucks.
```
axic: Can or Can't?

lib/loader/loader.c
```diff
@@ -238,3 +231,90 @@ struct evmc_instance* evmc_load_and_create(const char* filename,
...
+    {
+        char* option = options;
```
axic: Why not use `strtok` to do all these?

chfast: Not thread safe.

lib/loader/loader.c
chfast: Can. The `strlen(options) != 0` is not optimized by MSVC, but is more readable.

lib/loader/loader.c
chfast: The `strcpy_s` from MSVC does not work in this new case. Instead of returning non-zero value it triggers some global handler to handle the security issue.

lib/loader/loader.c
axic: Well you can use `strtok_r`. But I wonder if a helper similar to `strtok` in this file could simplify the code?

chfast: In my opinion it is not worth it. I fear it will be the same story as with `strcpy_s`.
- https://linux.die.net/man/3/strtok_r
- https://docs.microsoft.com/pl-pl/cpp/c-runtime-library/reference/strtok-s-strtok-s-l-wcstok-s-wcstok-s-l-mbstok-s-mbstok-s-l?view=vs-2019
- https://stackoverflow.com/questions/9021502/whats-the-difference-between-strtok-r-and-strtok-s-in-c

In practice anything from C standard lib besides `memcpy` and `strlen` is barely usable in portable way.

Do you want to try `strtok`? I might check it out our of curiosity.

axic: I mean you could have your own strtok-inspired helper in the code to be used in the 3 places where you tokenize. Not sure if it would simplify or just complicate the code.

chfast: There is interesting behavior of `strtok`:

> A sequence of two or more contiguous delimiter bytes in the parsed string is considered to be a single delimiter. Delimiter bytes at the start or end of the string are ignored. Put another way: the tokens returned by strtok() are always nonempty strings. 

Should we follow this?

chfast: Added `get_prefix()` helper, which is a bit different from `strtok()`. Code looks nice after changes.
I also added one more commit that optimizes `get_prefix()`, but I'm not sure this is the right change.

axic: I guess it is fine, considering with a comma delimited list of options, adding extra commas shouldn't do any harm. Question is do you want to allow that or raise an error?

lib/loader/loader.c
```diff
@@ -232,6 +232,29 @@ struct evmc_instance* evmc_load_and_create(const char* filename,
     return instance;
 }
 
+/// Gets the prefix delimited by @p delim character of the string pointed by the @p str_ptr.
+/// If the delimiter is not found, the whole string is returned.
+/// The @p str_ptr is also slided after the delimiter or to the string end
+/// if the delimiter is not found (in this case the @p str_ptr points to an empty string).
+static char* get_prefix(char** str_ptr, char delim)
```
axic: I think `get_token` is a better name, but not keen on it.

lib/loader/loader.c
```diff
@@ -270,25 +287,8 @@ struct evmc_instance* evmc_load_and_configure(const char* config,
...
-
-        char* kv_sep = strchr(option, '=');
-        const char* name = option;
-        char* value = "";  // Set value to empty to handle the "...,option,..." case.
-        if (kv_sep)
-        {
-            *kv_sep = '\0';
-            value = kv_sep + 1;
-        }
+        char* value = get_prefix(&options, ',');
```
axic: Should add a comment here why parsing is done "in reverse" (e.g. to support options without a value)

lib/loader/loader.c
chfast: > I guess it is fine, considering with a comma delimited list of options, adding extra commas shouldn't do any harm. Question is do you want to allow that or raise an error?

After considerations, this introduces other edge cases, e.g. `option===value` would mean `option=value`. So I prefer leave it as it is.

lib/loader/loader.c
chfast: I moved it as a first commit. We can land it separately, but I'd rather save out time.

test/unittests/test_loader.cpp
```diff
@@ -328,3 +321,250 @@ TEST_F(loader, load_and_create_abi_mismatch)
...
+    EXPECT_EQ(recorded_options[1].first, "b");
+    EXPECT_EQ(recorded_options[1].second, "_b");
+    EXPECT_EQ(recorded_options[2].first, "c");
+    EXPECT_EQ(recorded_options[2].second, "_c");
+    EXPECT_EQ(recorded_options[3].first, "b");
+    EXPECT_EQ(recorded_options[3].second, "_b");
+    EXPECT_EQ(ec, EVMC_LOADER_SUCCESS);
+
+    recorded_options.clear();
+    vm = evmc_load_and_configure("path,a=_a,b=_b,a=_a,", &ec);
```
axic: The last instance of the same option overrides the earlier ones. Can you test something like `a=_a,b=_b,a=_c` and ensure that `a` == `_c` ?

Also this probably should be documented.

test/unittests/test_loader.cpp
```diff
@@ -328,3 +321,250 @@ TEST_F(loader, load_and_create_abi_mismatch)
...
+TEST_F(loader, load_and_configure_degenerated_values)
+{
+    supported_options["flag"] = "";  // Empty value expected.
+    supported_options["e"] = "";     // Empty value expected.
+    supported_options[""] = "xxx";   // Empty name.
+
+    setup("path", "evmc_create", create_vm_with_set_option);
+
+    evmc_loader_error_code ec;
+    auto vm = evmc_load_and_configure("path,flag,e=,=xxx", &ec);
```
axic: So empty option keys are supported? The documentation states the opposite.

test/unittests/test_loader.cpp
```diff
@@ -328,3 +321,250 @@ TEST_F(loader, load_and_create_abi_mismatch)
...
+}
+
+TEST_F(loader, load_and_configure_degenerated_names)
+{
+    supported_options[""] = "";  // Allows ,, option.
+
+    setup("path", "evmc_create", create_vm_with_set_option);
+
+    evmc_loader_error_code ec;
+    auto vm = evmc_load_and_configure("path,,,=,,", &ec);
```
axic: Are these cases still supported?

lib/loader/loader.c
axic: Its fine, the commit describes the reason now.

test/unittests/test_loader.cpp
chfast: They are. Do you mean this part `<path> ("," <option-name> ["=" <option-value>])*`?

test/unittests/test_loader.cpp
chfast: Yes.

test/unittests/test_loader.cpp
chfast: Done.

test/unittests/test_loader.cpp
axic: Value seems to be optional, but the name.

`=xxx` suggests the name can be optional too.

test/unittests/test_loader.cpp
```diff
@@ -328,3 +324,272 @@ TEST_F(loader, load_and_create_abi_mismatch)
...
+    EXPECT_EQ(recorded_options[0].first, "x");
+    EXPECT_EQ(recorded_options[0].second, "x");
+    EXPECT_EQ(ec, EVMC_LOADER_SUCCESS);
+    evmc_destroy(vm);
+    EXPECT_EQ(destroy_count, create_count);
+}
+
+TEST_F(loader, load_and_configure_vm_without_set_option)
+{
+    supported_options[""] = {""};  // Allows ,, option.
```
axic: This comment seems to be misleading, since below this is an error.

test/unittests/test_loader.cpp
axic: This is still a question.

chfast: It's fine to me. I don't want to encourage this usage. Nobody said `<option-name>` cannot be empty.

axic: Ok.

axic: We can discuss that the ABI7.

test/unittests/test_loader.cpp
chfast: Fixed.

PR #314, opened 2019-06-05
Rust: Fix leaks in unit tests

jakelang:


bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -516,14 +518,26 @@ mod tests {
         }
     }
 
+    // Helper to safely dispose of the dummy context, and not bring up false positives in the
+    // sanitizers.
+    fn dummy_context_dispose(context: ffi::evmc_context) {
```
axic: Can you group this next to `get_dummy_context `above?

jakelang:
Update: There actually is no real leak, just some hastily written unit tests.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -360,7 +360,7 @@ mod tests {
             gas_left: 1337,
             output_data: Box::into_raw(Box::new([0xde, 0xad, 0xbe, 0xef])) as *const u8,
             output_size: 4,
-            release: None,
+            release: Some(release_stack_result),
```
axic: I don't think this is correct, as the output data here is just a boxed vector, while `release_stack_result` expects it was allocated via`std::alloc`.

I'd say you need to create a local release function in the tests which just do `Box::from_raw` to consume it.

jakelang: Both `release_stack_result` and `release_heap_result` call into `deallocate_output_data`, so there is no difference on that end.
It doesn't really expect that the buffer is allocated via `std::alloc` I think. As long as the pointer is coerced to a type which takes ownership of the memory, `Drop` can do the rest.

axic: There is no drop in `release_stack_result`, it uses `dealloc` directly.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -353,14 +353,28 @@ mod tests {
         assert!(r.get_create_address().is_none());
     }
 
+    // Test-specific helper to dispose of execution results in unit tests
+    extern "C" fn test_result_dispose(result: *const ffi::evmc_result) {
+        unsafe {
+            if !result.is_null() {
+                let owned = *result;
+                let owned_retbuf = Vec::from_raw_parts(
```
axic: Cannot all this just work with `drop(Box::from_raw(result))` ?

jakelang: A box taking ownership of a pointer will only drop the pointer.
You can do a boxed slice but that is more or less the same thing as `Vec` with extra steps.

PR #316, opened 2019-06-05
rust: unwind execution panics in evmc-declare

### LINKED ISSUE #310: Rust: unwind panics for the execute method
See https://github.com/ethereum/evmc/pull/262/files#r289002274.

axic:
Fixes #310.

### DIFF
--- a/bindings/rust/evmc-declare/src/lib.rs
+++ b/bindings/rust/evmc-declare/src/lib.rs
@@ -363,7 +363,15 @@ fn build_execute_fn(names: &VMNameSet) -> proc_macro2::TokenStream {
                 ::evmc_vm::EvmcContainer::<#type_name_ident>::from_ffi_pointer(instance)
             };
 
-            let result = container.execute(code_ref, &execution_context);
+            let result = ::std::panic::catch_unwind(|| {
+                container.execute(code_ref, &execution_context)
+            });
+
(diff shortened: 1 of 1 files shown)


bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -363,7 +363,15 @@ fn build_execute_fn(names: &VMNameSet) -> proc_macro2::TokenStream {
                 ::evmc_vm::EvmcContainer::<#type_name_ident>::from_ffi_pointer(instance)
             };
 
-            let result = container.execute(code_ref, &execution_context);
+            let result = ::std::panic::catch_unwind(|| {
```
jakelang: >It is not recommended to use this function for a general try/catch mechanism. The Result type is more appropriate to use for functions that can fail on a regular basis. Additionally, this function is not guaranteed to catch all panics, see the "Notes" section below.

https://doc.rust-lang.org/std/panic/fn.catch_unwind.html

axic: The main reason for this change is exactly listed in your link:
> It is currently undefined behavior to unwind from Rust code into foreign code, so this function is particularly useful when Rust is called from another language (normally C). This can run arbitrary Rust code, capturing a panic and allowing a graceful handling of the error.

PR #317, opened 2019-06-05
rust: container.execute to return a Result 

### LINKED ISSUE #309: Rust: change execute entry point to return a Result
See https://github.com/ethereum/evmc/pull/262#discussion_r281873596

axic:
Fixes #309.

### DIFF
--- a/bindings/rust/evmc-declare-tests/src/lib.rs
+++ b/bindings/rust/evmc-declare-tests/src/lib.rs
@@ -16,7 +16,7 @@ impl EvmcVm for FooVM {
         FooVM {}
     }
 
-    fn execute(&self, _code: &[u8], _context: &ExecutionContext) -> ExecutionResult {
-        ExecutionResult::success(1337, None)
+    fn execute(&self, _code: &[u8], _context: &ExecutionContext) -> Result<ExecutionResult, ()> {
+        Ok(ExecutionResult::success(1337, None))
     }
 }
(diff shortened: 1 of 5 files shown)


axic:
@jakelang should we have this?

jakelang:
@axic seeing as we handle failures via catching panics in the wrapper code, I don't really see a reason for it unless we refactor for error handling via the Result enum.

axic:
> @axic seeing as we handle failures via catching panics in the wrapper code, I don't really see a reason for it unless we refactor for error handling via the Result enum.

They serve two different purposes. The panic catching is to avoid messing up the stack for the caller.

jakelang:
Needs rebase

bindings/rust/evmc-declare-tests/src/lib.rs
```diff
@@ -16,7 +16,7 @@ impl EvmcVm for FooVM {
         FooVM {}
     }
 
-    fn execute(&self, _code: &[u8], _context: &ExecutionContext) -> ExecutionResult {
-        ExecutionResult::success(1337, None)
+    fn execute(&self, _code: &[u8], _context: &ExecutionContext) -> Result<ExecutionResult, ()> {
```
jakelang: Idea: Instead of having `()` as the err type, introduce a trait which the handler (the macro-generated code) uses to interpret the error type as an EVMC exit code. Add a trait-bounded associated type to the `EvmcVM` trait for the Error type parameter of this function's return.
That way, as long as the user implements the `EvmcError` trait on a type, that type can be used for more sophisticated error handling in a way that is still compatible with the EVMC API.

axic:
While this would be a more idiomatic Rust API, after having written to EVMC VMs using this API, I think requiring the manual creation of ExecutionResult, given its many different status codes, is better. Closing this.

PR #312, opened 2019-06-04
Add LLVM sanitizers to Rust CI

jakelang:
depends: #314

PR #323, opened 2019-06-19
circleci: use ASan for vmtester with rust

axic:
Follow up to #312.


axic:
```
rustc 1.35.0 (3c235d560 2019-05-20)
binary: rustc
commit-hash: 3c235d5600393dfe6c36eeed34042efad8d4f26e
commit-date: 2019-05-20
host: x86_64-unknown-linux-gnu
release: 1.35.0
LLVM version: 8.0
```

Yet it complains linking against llvm-8. I guess it may be that asan is 8.0.1 and Rust may use 8.0.0 ?

PR #324, opened 2019-06-19
rust: introduce ExecutionMessage wrapper

axic:


bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -21,10 +21,23 @@ pub struct ExecutionResult {
     create_address: Option<ffi::evmc_address>,
 }
 
+/// EVMC execution message structure.
+pub struct ExecutionMessage {
+    pub kind: ffi::evmc_call_kind,
```
axic: We could consider replacing pubs with getters and it that case we could just take a reference to `ffi::evmc_message`.

jakelang: That makes sense. the pub stuff belongs only in FFI IMO

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -339,6 +390,39 @@ extern "C" fn release_stack_result(result: *const ffi::evmc_result) {
...
+            flags: message.flags,
+            depth: message.depth,
+            gas: message.gas,
+            destination: message.destination,
+            sender: message.sender,
+            input: if message.input_data.is_null() {
+                assert!(message.input_size == 0);
+                None
+            } else {
+                // TODO: what to do if input_size is 0?
```
jakelang: panic.

PR #325, opened 2019-06-19
circleci: add MSan task for Rust

axic:


chfast:
I have some questions:
1. What kind of errors do you want to find with it and in what code?
2. MemorySanitizer is a subset of Valgrind Memcheck. The later would be much easier to use.

axic:
Well it already found an issue. We have already merge ASan support which also discovered some problems in the Rust FFI parts. We are handling memory in an unsafe manner, hence it is useful to enforce these checks.

chfast:
> Well it already found an issue.

Did MSan find any issue? It was never useful for me.

circle.yml
```diff
@@ -211,6 +205,27 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: Update environment
+          command: |
+            echo 'deb http://apt.llvm.org/stretch/ llvm-toolchain-stretch-8 main' >> /etc/apt/sources.list
+            apt-key adv --keyserver keyserver.ubuntu.com --no-tty --recv-keys 6084F3CF814B57C1CF12EFD515CF4D18AF4F7421 60C317803A41BA51845E371A1E9377A2BA9EF27F
+            apt -qq update
+            apt -yq  install llvm-8-dev clang-8
+            rustup toolchain install nightly-x86_64-unknown-linux-gnu
```
chfast: Why do you need to update rust if you use a docker image?

axic: This is about installing nightly, it is not available by default.

axic:
> > Well it already found an issue.
>
> Did MSan find any issue? It was never useful for me.

Check out the log for `bindings-rust-msan`. It is not fixed yet.

chfast:
> Check out the log for `bindings-rust-msan`.

I stopped at build warnings previously.

axic:
> > Check out the log for bindings-rust-msan.
>
> I stopped at build warnings previously.

There are no build warnings.

chfast:
> > > Check out the log for bindings-rust-msan.
> > 
> > 
> > I stopped at build warnings previously.
> 
> There are no build warnings.

```
   Compiling example-rust-vm v0.1.0 (/root/project/examples/example-rust-vm)
warning: unused variable: `code`
  --> bindings/rust/evmc-declare-tests/src/lib.rs:22:23
   |
22 |     fn execute(&self, code: &[u8], context: &ExecutionContext) -> ExecutionResult {
   |                       ^^^^ help: consider prefixing with an underscore: `_code`
```

axic:
Okay, you can ignore them, those are style warnings in tests.

circle.yml
chfast: Yes, I know. It would be good to make a custom image with LLVM and rust nightly or run different configuration builds as a sequence of steps, not independent jobs.

PR #327, opened 2019-06-25
Add support for 32-bit architectures

### LINKED ISSUE #326: Test 32-bit builds
Add CI jobs for:
- Visual Studio 32-bit,
- linux 32-bit.

chfast:
Closes #326.

### DIFF
--- a/include/evmc/helpers.hpp
+++ b/include/evmc/helpers.hpp
@@ -53,17 +53,43 @@ inline bool is_zero(const evmc_bytes32& x) noexcept
     return x == evmc_bytes32{};
 }
 
-/// FNV1a hash function with 64-bit result.
-inline uint64_t fnv1a_64(const uint8_t* ptr, size_t len)
+/// Parameters for the fnv1a hash function, specialized by the hash result size (size_t).
+///
+/// The values for the matching size are taken from
+/// https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function#FNV_hash_parameters.
+///
+/// @tparam size  The size of the hash result (size_t).
+template <size_t size>
+struct fnv1_params
 {
-    constexpr uint64_t prime = 1099511628211ULL;
-    constexpr uint64_t offset_basis = 14695981039346656037ULL;
+};
 
-    uint64_t ret = offset_basis;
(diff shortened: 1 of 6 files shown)


include/evmc/helpers.hpp
```diff
@@ -53,17 +53,35 @@ inline bool is_zero(const evmc_bytes32& x) noexcept
...
+{
+    static constexpr auto prime = 0x1000193;
+    static constexpr auto offset_basis = 0x811c9dc5;
+};
+
+template <>
+struct fnv1_params<8>
+{
+    static constexpr auto prime = 0x100000001b3;
+    static constexpr auto offset_basis = 0xcbf29ce484222325;
```
axic: Are these the same as the decimals in the previous version?

test/unittests/test_helpers.cpp
```diff
@@ -31,6 +31,13 @@ static constexpr size_t optionalDataSize =
 static_assert(optionalDataSize == sizeof(evmc_result_optional_storage), "");
 
 
+TEST(helpers, fnv1a)
```
axic: The commit log is wrong for this, it is about testing fnv1a.

chfast: `git rebase -i` is planned.

include/evmc/helpers.hpp
chfast: Yes. The `0x...` makes them `unsigned`.

axic: The decimal version had a `ULL` postfix which supposedly would make it unsigned.

test/vmtester/tests.cpp
```diff
@@ -110,7 +110,7 @@ TEST_F(evmc_vm_test, execute_create)
     }
     else
     {
-        EXPECT_NE(result.output_size, 0);
+        EXPECT_NE(result.output_size, size_t{0});
```
axic: Why are these needed?

include/evmc/helpers.hpp
chfast: Yes, but then you have to decide what type it has to be unless `u` postfix will make it `unsigned long` if big enough.

test/vmtester/tests.cpp
chfast: A warning in MSVC about comparing signed and unsigned types.

PR #329, opened 2019-06-25
ci: Refactor Circle CI config

chfast:
Use CircleCI 2.1 which allows creating groups of steps and reuse them in job specifications.
This change is mostly motivated by the #327 where I need to modify jobs for 32-bit builds but still want to use the common build steps.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -5,33 +5,8 @@ executors:
     docker:
       - image: ethereum/cpp-build-env:10
 
-jobs:
-
-  lint:
-    executor: linux
-    steps:
-    - checkout
-    - run:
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -219,7 +224,7 @@ workflows:
       - build-cxx17
       - build-cxx14-asan
       - build-gcc6
-      - build-clang3.8
+      - build-clang38
```
axic: Why change the name of this?

chfast: In CircleCI 2.1 you cannot have dots in the job names.

chfast:
Any questions left here?

PR #330, opened 2019-06-25
rust: implement Deref trait for EvmcContainer

axic:


bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -334,6 +334,8 @@ fn build_execute_fn(names: &VMNameSet) -> proc_macro2::TokenStream {
             code_size: usize
         ) -> ::evmc_vm::ffi::evmc_result
         {
+            use evmc_vm::EvmcVm;
```
jakelang: I would rather not put `use` statements inline with code.

PR #331, opened 2019-06-25
rust: use ExecutionMessage in ExecutionContext.call() 

### LINKED ISSUE #319: Rust: add abstraction for evmc_message and types
To avoid the need to deal with raw pointers in `evmc_message`.

axic:
Closes #319.

### DIFF
--- a/bindings/rust/evmc-vm/src/lib.rs
+++ b/bindings/rust/evmc-vm/src/lib.rs
@@ -731,6 +731,36 @@ mod tests {
         105023 as usize
     }
 
+    unsafe extern "C" fn execute_call(
+        _context: *mut ffi::evmc_context,
+        _msg: *const ffi::evmc_message,
+    ) -> ffi::evmc_result {
+        // Some dumb validation for testing.
+        let msg = *_msg;
+        let success = if msg.input_size != 0 && msg.input_data == std::ptr::null() {
+            false
+        } else if msg.input_size == 0 && msg.input_data != std::ptr::null() {
+            false
+        } else {
+            true
+        };
+
+        ffi::evmc_result {
+            status_code: if success {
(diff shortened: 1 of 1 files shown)


bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -320,7 +369,7 @@ impl From<ffi::evmc_result> for ExecutionResult {
     }
 }
 
-fn allocate_output_data(output: Option<Vec<u8>>) -> (*const u8, usize) {
+fn allocate_output_data(output: Option<&Vec<u8>>) -> (*const u8, usize) {
```
axic: Made this improvement here because first tried to use this in `call()`, but realised there's no need to duplicate memory.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -243,12 +267,37 @@ impl<'a> ExecutionContext<'a> {
         }
     }
 
-    pub fn call(&mut self, message: &ffi::evmc_message) -> ExecutionResult {
+    pub fn call(&mut self, message: &ExecutionMessage) -> ExecutionResult {
```
jakelang: I am a bit concerned about passing by reference here. It can be extremely weird if the host ends up mutating through a pointer to something that belonged to the `ExecutionMessage`.
Maybe it should just be consumed by value?

axic: Well it is passed as a `const` to the host so it should not mutate it, otherwise it violates the EVMC interface.

jakelang: Good point, should be fine then.

axic:
@jakelang this has a test now, should be ready for merging

PR #332, opened 2019-06-26
test: Rewrite an unit test to avoid UBSan warning

chfast:
Previously in the last iteration of the for loop the value EVMC_MAX_REVISION + 1 is assigned to rev of type evmc_revision what triggers an UBSan warning.

### DIFF
--- a/test/unittests/test_instructions.cpp
+++ b/test/unittests/test_instructions.cpp
@@ -113,9 +113,9 @@ TEST(instructions, constantinople_hard_fork)
 
 TEST(instructions, name_gas_cost_equivalence)
 {
-    for (auto rev = EVMC_FRONTIER; rev <= EVMC_MAX_REVISION;
-         rev = static_cast<evmc_revision>(rev + 1))
+    for (auto r = int{EVMC_FRONTIER}; r <= EVMC_MAX_REVISION; ++r)
     {
+        const auto rev = static_cast<evmc_revision>(r);
         const auto names = evmc_get_instruction_names_table(rev);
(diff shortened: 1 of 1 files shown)


test/unittests/test_instructions.cpp
```diff
@@ -113,9 +113,9 @@ TEST(instructions, constantinople_hard_fork)
 
 TEST(instructions, name_gas_cost_equivalence)
 {
-    for (auto rev = EVMC_FRONTIER; rev <= EVMC_MAX_REVISION;
-         rev = static_cast<evmc_revision>(rev + 1))
+    for (auto r = int{EVMC_FRONTIER}; r <= EVMC_MAX_REVISION; ++r)
     {
+        const auto rev = evmc_revision(r);
```
gumb0: why not `static_cast`, as before?

chfast: The convention I use:
- `static_cast` for expected loss of information,
- `T()` for cases where loss of information should not happen,
- `T{}` loss of information not possible (enforced by compiler), use as much as you can.

chfast: @gumb0 is this ok?

gumb0: Yeah it's not a big deal, it's just that it's only your convention, e.g. Core Guidelines recommend `static_cast` when cast is needed. (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es49-if-you-must-use-a-cast-use-a-named-cast)

PR #333, opened 2019-06-26
cpp: Extend API for creating results

### LINKED ISSUE #258: Improve C++ API for result creation
https://github.com/ethereum/evmc/blob/v6.2.0/examples/example_host.cpp#L98

chfast:
Closes #258.

Usage check:
- [x] Test in evmone: https://github.com/ethereum/evmone/pull/77
- [x] Test in aleth-interpreter (VM.cpp): https://github.com/ethereum/aleth/pull/5715
- [ ] ~~Test in Aleth (`EvmCHost::call()`, `EvmCHost::create()`)~~

### DIFF
--- a/include/evmc/evmc.hpp
+++ b/include/evmc/evmc.hpp
@@ -7,8 +7,6 @@
 #include <evmc/evmc.h>
 #include <evmc/helpers.h>
 
-#include <algorithm>
-#include <cstdlib>
 #include <functional>
 #include <initializer_list>
 #include <utility>
@@ -257,6 +255,10 @@ constexpr bytes32 operator"" _bytes32() noexcept
 
 using namespace literals;
 
+
+/// Alias for evmc_make_result().
+constexpr auto make_result = evmc_make_result;
+
 /// @copydoc evmc_result
 ///
 /// This is a RAII wrapper for evmc_result and objects of this type
@@ -283,19 +285,8 @@ public:
            int64_t _gas_left,
            const uint8_t* _output_data,
            size_t _output_size) noexcept
-      : evmc_result{_status_code, _gas_left, nullptr, _output_size, {}, {}, {}}
-    {
-        if (output_size != 0)
-        {
-            auto mem = static_cast<uint8_t*>(std::malloc(output_size));
-            std::copy_n(_output_data, output_size, mem);
-            output_data = mem;
-            release = [](const evmc_result* r) noexcept
-            {
-                std::free(const_cast<uint8_t*>(r->output_data));
-            };
-        }
-    }
+      : evmc_result{make_result(_status_code, _gas_left, _output_data, _output_size)}
+    {}
 
     /// Converting constructor from raw evmc_result.
     explicit result(evmc_result const& res) noexcept : evmc_result{res} {}
--- a/include/evmc/evmc.hpp
+++ b/include/evmc/evmc.hpp
@@ -7,6 +7,8 @@
 #include <evmc/evmc.h>
 #include <evmc/helpers.h>
 
+#include <algorithm>
+#include <cstdlib>
 #include <functional>
 #include <initializer_list>
 #include <utility>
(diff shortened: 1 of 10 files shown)


include/evmc/evmc.hpp
```diff
@@ -27,6 +29,33 @@ class result : private evmc_result
...
+    /// @param _output_size  The output size.
+    result(evmc_status_code _status_code,
+           int64_t _gas_left,
+           const uint8_t* _output_data,
+           size_t _output_size) noexcept
+      : evmc_result{_status_code, _gas_left, nullptr, _output_size, {}, {}, {}}
+    {
+        if (output_size != 0)
+        {
+            auto mem = static_cast<uint8_t*>(std::malloc(output_size));
```
axic: Apparently this can return `nullptr`. Why not use `new` and `delete` here instead?

chfast: Not practical. I want to keep VM implementation low profile and require only C standard library if possible.

But good point raised. How about converting it to `EVMC_INTERNAL_ERROR`?

axic: > How about converting it to EVMC_INTERNAL_ERROR?

I think that's a bad idea, because it masks what's going on. The caller would think the called contract failed.

chfast: I'm not sure what do you mean by a caller here but in any case it should not take EVMC internal errors as indicators for contract execution failures.

I believe we only have 2 options here: ignore it or report it as an internal error.

include/evmc/evmc.hpp
```diff
@@ -27,6 +31,22 @@ class result : private evmc_result
...
+    /// Creates the result from the provided arguments.
+    ///
+    /// The provided output is copied to memory allocated with malloc()
+    /// and the evmc_result::release function is set to one invoking free().
+    ///
+    /// @param _status_code  The status code.
+    /// @param _gas_left     The amount of gas left.
+    /// @param _output_data  The pointer to the output.
+    /// @param _output_size  The output size.
+    result(evmc_status_code _status_code,
```
gumb0: In other functions you don't use underscore prefix for parameter names

chfast: We have to somehow avoid field names shadowing here.

include/evmc/evmc.hpp
```diff
@@ -7,13 +7,17 @@
 #include <evmc/evmc.h>
 #include <evmc/helpers.h>
 
+#include <algorithm>
```
gumb0: Why is this needed?

chfast: Checking...

chfast: No needed.

include/evmc/helpers.h
```diff
@@ -106,6 +108,55 @@ static inline struct evmc_result evmc_execute(struct evmc_instance* instance,
...
+    struct evmc_result result;
+    memset(&result, 0, sizeof(result));
+
+    if (output_size != 0)
+    {
+        uint8_t* buffer = (uint8_t*)malloc(output_size);
+
+        if (!buffer)
+        {
+            result.status_code = EVMC_OUT_OF_MEMORY;
```
axic: Still not convinced this it the right way to do it, given it confuses status of the execution vs. helpers, but in the ned if OOM is hit there is a bigger underlying issue. Perhaps worth highlighting this in the description above?

chfast: > but if OOM is hit there is a bigger underlying issue

What?

I don't thing this is the best way of passing output buffer. This will rather be only useful for Go, so we can drop this PR.

axic: > > but if OOM is hit there is a bigger underlying issue
>
> What?

OOM = out of memory

Basically I'm ok with this change given this new specific error code is introduced.

chfast: No, I had issue with understanding the sentence after "OOM". The OOM itself is difficult to handle in general. E.g. it will never happen on Linux in practice as it will happily try to allocate 2 TB of memory and in the end kill the process.

Anyway, the new specific error code was introduced: https://github.com/ethereum/evmc/pull/333/files#diff-8e9caa43721b131d9f7897e2bfd98299R309.

axic: Yeah I know, that's the line I commented on. Just having this documented on `make_result` would be useful.

chfast:
I used this in Go: `Host.Call()`.

include/evmc/evmc.hpp
gumb0: Remove it then?

include/evmc/evmc.hpp
```diff
@@ -7,6 +7,8 @@
...
+#include <cstdlib>
```
gumb0: This also not needed I guess, helpers.h include it?

chfast:
All fixed.

PR #342, opened 2019-07-03
rust: support ParitalEq, Default, Hash on certain ffi types

axic:


bindings/rust/evmc-sys/src/lib.rs
```diff
@@ -8,3 +8,30 @@
 #![allow(non_snake_case)]
 
 include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+
+// TODO: with bindgen's interface improving these may be moved to
+// bindgen configuration
+
+impl PartialEq for evmc_address {
```
jakelang: Implementing traits for bindgen-declared types is not the best approach IMO.

axic: It is explained in the comment above and the commit log.

PR #206, opened 2019-03-13
rust: add enum aliases

axic:
Part of #168.

PR #343, opened 2019-07-03
rust: add type aliases for Address, Bytes32, Uint256

axic:
Part of #319.

Enums are handled in #206.

### DIFF
--- a/bindings/rust/evmc-vm/src/types.rs
+++ b/bindings/rust/evmc-vm/src/types.rs
@@ -1,0 +1,37 @@
+use evmc_sys as ffi;
+
+/// EVMC address
+pub type Address = ffi::evmc_address;
+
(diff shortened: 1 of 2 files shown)


bindings/rust/evmc-vm/src/types.rs
```diff
@@ -0,0 +1,34 @@
+use evmc_sys as ffi;
+
+pub type Address = ffi::evmc_address;
+
+pub type Bytes32 = ffi::evmc_bytes32;
+
+pub type Uint256be = ffi::evmc_uint256be;
```
axic: Not happy with this name, perhaps Uint256 is enough?

axic:
Need to squash this prior to merge.

axic:
That was original version, but that requires from traits for conversion, IIRC.

PR #349, opened 2019-07-04
rust: add more tests to EvmcContainer

axic:


bindings/rust/evmc-vm/src/container.rs
```diff
@@ -61,6 +62,20 @@ mod tests {
         }
     }
 
+    unsafe extern "C" fn get_dummy_tx_context(
```
jakelang: This seems to be a duplicate of the one in lib.rs. Perhaps we could update the one in lib.rs and include it?

bindings/rust/evmc-vm/src/container.rs
```diff
@@ -61,6 +62,20 @@ mod tests {
...
+        _context: *mut evmc_sys::evmc_context,
+    ) -> evmc_sys::evmc_tx_context {
+        evmc_sys::evmc_tx_context {
+            tx_gas_price: Uint256::default(),
```
jakelang: This is ok but personally I'm not a huge fan of `Default`. Would rather have methods with descriptive names like `Uint256::zero()`, `Uint256::max()`, etc.

bindings/rust/evmc-vm/src/container.rs
```diff
@@ -74,10 +89,48 @@ mod tests {
             set_option: None,
         };
 
+        let code = [0u8; 0];
```
jakelang: Maybe leave a note here that this is just for a dummy call.

bindings/rust/evmc-vm/src/container.rs
axic: It is not an exact copy though.

bindings/rust/evmc-vm/src/container.rs
axic: Yeah, well it can be changed in a separate PR, but for tests I think it is definitely fine.

bindings/rust/evmc-vm/src/container.rs
axic: Well its a test, isn't it? Happy to add a comment just let me know what exactly.

axic:
Merging this so we can get the release out sooner.

PR #350, opened 2019-07-04
rust: make ExecutionContext optional in EvmcVm.execute()

### LINKED ISSUE #318: Rust bindings are incompatible with EIP-2003
[EIP-2003](https://github.com/ethereum/EIPs/pull/2003) states that `code` and (host) `context` is not mandatory for precompiles and can be null, but the Rust bindings assert for the presence of both.

axic:
Fixes #318.

### DIFF
--- a/bindings/rust/evmc-declare-tests/src/lib.rs
+++ b/bindings/rust/evmc-declare-tests/src/lib.rs
@@ -22,7 +22,7 @@ impl EvmcVm for FooVM {
         _revision: evmc_sys::evmc_revision,
         _code: &[u8],
         _message: &ExecutionMessage,
-        _context: &mut ExecutionContext,
+        _context: Option<&mut ExecutionContext>,
     ) -> ExecutionResult {
         ExecutionResult::success(1337, None)
     }
--- a/bindings/rust/evmc-vm/src/container.rs
+++ b/bindings/rust/evmc-vm/src/container.rs
@@ -69,7 +69,7 @@ mod tests {
             _revision: evmc_sys::evmc_revision,
             _code: &[u8],
             _message: &ExecutionMessage,
-            _context: &mut ExecutionContext,
+            _context: Option<&mut ExecutionContext>,
         ) -> ExecutionResult {
             ExecutionResult::failure()
         }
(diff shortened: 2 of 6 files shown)


bindings/rust/evmc-declare-tests/src/lib.rs
```diff
@@ -16,7 +17,12 @@ impl EvmcVm for FooVM {
         FooVM {}
     }
 
-    fn execute(&self, _code: &[u8], _context: &ExecutionContext) -> ExecutionResult {
+    fn execute(
+        &self,
+        _code: &[u8],
```
jakelang: I figured out recently that the identifiers for the struct members and the arguments won't conflict, therefore we don't need to prefix with an underscore.

examples/example-rust-vm/src/lib.rs
```diff
@@ -14,8 +14,13 @@ impl EvmcVm for ExampleRustVM {
         ExampleRustVM {}
     }
 
-    fn execute(&self, _code: &[u8], context: &ExecutionContext) -> ExecutionResult {
-        let is_create = context.get_message().kind() == evmc_sys::evmc_call_kind::EVMC_CREATE;
+    fn execute(
+        &self,
+        _code: &[u8],
```
jakelang: Let's get rid of this underscore to make the argument formatting consistent.

bindings/rust/evmc-vm/src/container.rs
```diff
@@ -49,14 +49,19 @@ where
...
 
     struct TestVm {}
     impl EvmcVm for TestVm {
         fn init() -> Self {
             TestVm {}
         }
-        fn execute(&self, _code: &[u8], _context: &ExecutionContext) -> ExecutionResult {
+        fn execute(
+            &self,
+            _code: &[u8],
```
jakelang: Drop the underscores here as well

bindings/rust/evmc-declare-tests/src/lib.rs
axic: These are in examples and are prefixed for making clippy happy since they are not used.

axic:
> I feel like this would be cleaner if ExecutionContext would make its context member optional instead

The problem with that is every single feature of `ExecutionContext` depends on that field being present.

chfast:
Can this be merged?

axic:
@jakelang rebased, but not sure I like it.

@chfast we can also consider precompiles support for 6.4.

PR #354, opened 2019-07-15
Add separate Istanbul instruction tables

chfast:
It is highly likely that EVM instructions are going to be changes or added in Istanbul. Copy the table from Constantinople for easier EIP prototype implementations.

This is non-functional change.


gumb0:
Maybe some unit test checking that they're equivalent to Constantinople?

chfast:
> Maybe some unit test checking that they're equivalent to Constantinople?

Added.

PR #355, opened 2019-07-15
 Add instructions for EIP-663 variant C

chfast:


axic:
@chfast I think we can close this now.

PR #357, opened 2019-07-22
C++: Add basic types based on C types

chfast:
The basic types `address` and `bytes32` have received their C++ wrappers to assure they are always initialized. They also have convenient operator overloadings for comparison and usage as keys in standard containers.

Closes https://github.com/ethereum/evmc/issues/352

PR #358, opened 2019-07-22
Deprecate helpers.hpp

chfast:
Depends on #357

### DIFF
--- a/include/evmc/helpers.hpp
+++ b/include/evmc/helpers.hpp
@@ -18,39 +18,47 @@
 #include <functional>
 
 /// The comparator for std::map<evmc_address, ...>.
+EVMC_DEPRECATED
 inline bool operator<(const evmc_address& a, const evmc_address& b)
 {
     return std::memcmp(a.bytes, b.bytes, sizeof(a.bytes)) < 0;
 }
 
 /// The comparator for std::map<evmc_bytes32, ...>.
+EVMC_DEPRECATED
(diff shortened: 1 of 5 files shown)


include/evmc/helpers.hpp
```diff
@@ -101,6 +109,7 @@ template <>
 struct hash<evmc_address>
 {
     /// Hash operator using FNV1a.
+    EVMC_DEPRECATED
```
gumb0: Can't it be applied to the entire `struct hash<evmc_address>`?

chfast: It can be.

test/unittests/test_helpers.cpp
```diff
@@ -27,27 +26,6 @@ static constexpr size_t optionalDataSize =
...
 
-
-TEST(helpers, fnv1a)
-{
-    const uint8_t text[] = {'E', 'V', 'M', 'C'};
-    const auto h = fnv1a(text, sizeof(text));
-    EXPECT_EQ(h, sizeof(size_t) == 8 ? 0x15e05d6d22fed89a : 0xffaa6a9a);
-}
-
-TEST(helpers, is_zero)
```
axic: I don't think tests should be removed with deprecation. We still need to ensure nobody breaks the deprecated features.

axic: That being said, I think the test removed in https://github.com/ethereum/evmc/commit/279c1cd6e75b69947f33b33a95348939d3c72206 should also be brought back.

PR #359, opened 2019-07-22
C++: Literals for basic types

chfast:


test/unittests/test_cpp.cpp
```diff
@@ -10,11 +10,221 @@
...
+    static_assert(hash1.bytes[10] == 0xa1, "");
+    static_assert(hash1.bytes[31] == 0xd2, "");
+    static_assert(zero_address == evmc::address{}, "");
+    static_assert(zero_hash == evmc::bytes32{}, "");
+#endif
+
+    EXPECT_EQ(0_addr, evmc::address{});
+    EXPECT_EQ(0_b32, evmc::bytes32{});
+
+    auto a1 = 0xa0a1a2a3a4a5a6a7a8a9d0d1d2d3d4d5d6d7d8d9_addr;
```
axic: Since these are long literals anyway, I see no benefit shortening `address` to `addr` and `bytes32` to `b32`.

include/evmc/evmc.hpp
```diff
@@ -150,6 +150,108 @@ constexpr bytes32::operator bool() const noexcept
...
+struct concatenate<byte_sequence<Bytes1...>, byte_sequence<Bytes2...>>
+{
+    using type = byte_sequence<Bytes1..., Bytes2...>;
+};
+
+
+constexpr uint8_t parse_hex_digit(uint8_t d) noexcept
+{
+    return static_cast<uint8_t>(
+        (d >= '0' && d <= '9') ? d - '0' : (d >= 'a' && d <= 'f') ? d - 'a' + 10 : d - 'A' + 10);
```
axic: Any chance to abort on invalid hex digits? E.g. `0x123kkx_address` ?

axic:
Need to apply renaming to changelog.

include/evmc/evmc.hpp
chfast: This will be C++ parser error. But I've found in C++17 you can have literals like `0x0.000000000000000000000000000000000000p0_address` so I added additional checks.

include/evmc/evmc.hpp
```diff
@@ -150,6 +150,110 @@ constexpr bytes32::operator bool() const noexcept
     return !is_zero(*this);
 }
 
+namespace literals
+{
+namespace internal
```
gumb0: Should internal stuff better go into separate header?

include/evmc/evmc.hpp
axic: I reckon it is not easy to add a test case (e.g. file which would not compile)?

chfast: Not without creating custom testing framework in CMake.

include/evmc/evmc.hpp
chfast: So far I'm ok with having single header for C++ (now it's even better when helpers.hpp is gone).

PR #361, opened 2019-07-23
rust: add ExecutionTxContext type alias

axic:


axic:
Merging this as it is simple and working.

PR #362, opened 2019-07-23
examples: Move example_vm into the examples/example_vm subdirectory

axic:


axic:
@chfast the goal here was to visually decouple `example_vm` and bring it to the "same level" as `example_precompiles_vm`.

Eventually I think the `example.c` should move on to use the loader (as that is the default behaviour we encourage) and `example_vm` could be rewritten in C++ (as we don't expect many pure C VMs).

PR #365, opened 2019-07-23
circleci: also test example.c from examples

axic:


axic:
@chfast any idea why example is failing?

examples/example.c
```diff
@@ -34,9 +34,11 @@ int main()
     msg.depth = 0;
     struct evmc_result result = evmc_execute(vm, ctx, EVMC_HOMESTEAD, &msg, code, code_size);
     printf("Execution result:\n");
+    int exit_code = 0;
     if (result.status_code != EVMC_SUCCESS)
     {
         printf("  EVM execution failure: %d\n", result.status_code);
+        exit_code = -1;
```
chfast: It returns this.

axic: But it shouldn't :)

axic: Based on the inputs in this file, this case should be triggered: https://github.com/ethereum/evmc/blob/master/examples/example_vm/example_vm.c#L119

chfast: `code_size` is wrong.

PR #366, opened 2019-07-23
Add support for EVMC module configuration string in evmc-vmtester

chfast:


test/vmtester/vmtester.cpp
```diff
@@ -137,8 +119,12 @@ int main(int argc, char* argv[])
 
         const auto& evmc_module = cli.arguments[0];
         std::cout << "Testing " << evmc_module << "\n";
+
         evmc_loader_error_code ec;
-        create_fn = evmc_load(evmc_module.c_str(), &ec);
+        evmc_vm_test::vm = evmc_load_and_configure(evmc_module.c_str(), &ec);
+
+        // The safe C++ wrapper is used to make sure VM is destroyed properly.
+        static const evmc::vm loaded_vm{evmc_vm_test::vm};
```
gumb0: Couldn't you declare `evmc_vm_test::vm` as `evmc::vm`, then move-assign here?

test/vmtester/vmtester.hpp
```diff
@@ -1,23 +1,18 @@
...
 
-evmc_instance* get_vm_instance();
-
 class evmc_vm_test : public ::testing::Test
 {
-protected:
-    evmc_instance* vm = nullptr;
+public:
+    /// The raw pointer to the loaded VM instance.
+    /// The C API is used to allow more sophisticated unit tests.
```
chfast: As mentioned here, we want to access to C API for better coverage.

gumb0: Still maybe better to declare `evmc::vm` as a static member of the class close to `vm`? Then it's clearer that they're coupled.

gumb0: Or maybe I would just not use `vm` at all then, and destroy it manually in `evmc_vm_test` destructor

chfast: I moved them close together. Using the destructor is not an option because destructor is called for every test case.

test/vmtester/vmtester.cpp
```diff
@@ -3,29 +3,18 @@
 // Licensed under the Apache License, Version 2.0.
 
 #include "vmtester.hpp"
-
 #include <evmc/loader.h>
-
+#include <evmc/evmc.hpp>
```
gumb0: (includes look not ordered, "evmc" < "loader")

chfast: I noticed, but clang-format prefers `.h` over `.hpp` in current configuration.

PR #375, opened 2019-07-30
Support for EIP-1344 CHAINID opcode

gumb0:
https://eips.ethereum.org/EIPS/eip-1344


examples/example_host.cpp
```diff
@@ -129,6 +129,8 @@ class ExampleHost : public evmc::Host
         (void)topics;
         (void)topics_count;
     }
+
+    evmc_bytes32 get_chain_id() noexcept final { return 0_bytes32; }
```
axic: I think this should go into the `tx_context` header.

gumb0: My line of thought was: 1. it's probably not going to be used very often, so why copy it with every `get_tx_context` call; 2. it's not "transaction context" conceptually, but rather more global chain constant.
But I don't insist, can change.

gumb0:
Not sure what happens to `bindings-rust-asan-combined` build.

gumb0:
@axic I'm bumping the version here, is that ok?

test/unittests/test_instructions.cpp
```diff
@@ -251,7 +251,21 @@ TEST(instructions, istanbul_hard_fork)
...
+            EXPECT_EQ(i[op], p[op]) << op;
+            EXPECT_STREQ(in[op], pn[op]) << op;
+            break;
+        }
     }
+
+    EXPECT_EQ(i[OP_CHAINID].gas_cost, 2);
+    EXPECT_EQ(i[OP_CHAINID].num_stack_arguments, 0);
+    EXPECT_EQ(i[OP_CHAINID].num_stack_returned_items, 1);
+    EXPECT_EQ(p[OP_CHAINID].gas_cost, -1);
```
axic: -1?

include/evmc/evmc.h
```diff
@@ -153,6 +153,7 @@ struct evmc_tx_context
     int64_t block_timestamp;         /**< The block timestamp. */
     int64_t block_gas_limit;         /**< The block gas limit. */
     evmc_uint256be block_difficulty; /**< The block difficulty. */
+    evmc_uint256be chain_id;         /**< The blockchain's ChainID. */
```
axic: This part is contentious. There is a long discussion on https://ethereum-magicians.org/t/eip-1344-add-chain-id-opcode/1131/56

However many say it should be a hash, e.g. bytes32, however the calculation done the transaction format requires it to be a number.

Geth uses 256-bit, Parity 64-bit and Pantheon 32-bit numbers for chain id.

test/unittests/test_instructions.cpp
gumb0: This mimics other tests in this file, but I guess it should better use `UNDEFINED` macro

axic: Ah didn't notice it was `p` (petersburg). These short variable names are easy to mistake :)

gumb0:
Rebased.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -759,6 +759,7 @@ mod tests {
             block_timestamp: 235117,
             block_gas_limit: 105023,
             block_difficulty: Uint256 { bytes: [0xaa; 32] },
+            chain_id: Uint256 { bytes: [0u8; 32] },
```
axic: Can use `Uint256::default()` here, the other used the in-line declaration to set a non-zero value.

gumb0: I've changed it, though `tx_gas_price`, `tx_origin`, `block_coinbase` are zeroes here, too.

gumb0:
Rebased.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ Documentation of all notable changes to the **EVMC** project.
 The format is based on [Keep a Changelog],
 and this project adheres to [Semantic Versioning].
 
+## [7.0.0] - unreleased
+- Added: [[#375](https://github.com/ethereum/evmc/pull/375)]
```
chfast: Please update to new format.

chfast:
@axic Can you check Rust CI? It is  failing from some days.

CHANGELOG.md
gumb0: Updated.

axic:
> Can you check Rust CI? It is failing from some days.

@chfast I tried to fix it (#408), but it seems the unstable distro link has broken packages right now, and msan needs llvm8.

PR #376, opened 2019-08-05
Deprecate tracing API

### LINKED ISSUE #260: Deprecate the tracing API
The tracing API is attached in wrong place so it has to be changed. We should deprecate the current API as it is planned to be changed.

chfast:
Closes #260.

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -859,7 +859,11 @@ typedef uint32_t evmc_capabilities_flagset;
  */
 typedef evmc_capabilities_flagset (*evmc_get_capabilities_fn)(struct evmc_instance* instance);
 
-/** The opaque type representing a Client-side tracer object. */
+/**
+ * The opaque type representing a Client-side tracer object.
+ *
+ * @deprecated Deprecated since EVMC 6.3, see evmc_instance::set_tracer().
+ */
 struct evmc_tracer_context;
 
 /**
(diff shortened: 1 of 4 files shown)


include/evmc/evmc.h
```diff
@@ -991,7 +991,12 @@ struct evmc_instance
      * Optional pointer to function setting the EVM instruction tracer.
      *
      * If the EVM does not support this feature the pointer can be NULL.
+     *
+     * @deprecated
+     * The tracing API from EVMC 6 has been deprecated as there have been some
+     * design flaws discovered. New API is expected to be introduced in future.
      */
+    EVMC_DEPRECATED
```
chfast: We cannot put [[deprecated]] attribute here because it makes the `evmc_instance` non-constuctable in some compilers: https://circleci.com/gh/ethereum/evmc/13261.

include/evmc/evmc.h
```diff
@@ -991,6 +991,10 @@ struct evmc_instance
      * Optional pointer to function setting the EVM instruction tracer.
      *
      * If the EVM does not support this feature the pointer can be NULL.
+     *
+     * @deprecated
+     * The tracing API from EVMC 6 has been deprecated as there have been some
```
gumb0: It's not deprecated in 6.2, should this say `from EVMC 6.3` ?

gumb0:
What about `evmc_set_tracer_fn`, `evmc_trace_callback`, `evmc_tracer_context`, they stay?

chfast:
> What about `evmc_set_tracer_fn`, `evmc_trace_callback`, `evmc_tracer_context`, they stay?

We can only use `@deprecated` in doxygen there, because otherwise we will generate warnings in other places.

PR #377, opened 2019-08-05
cpp: Use new address and bytes32 wrappers in HostInterface

chfast:


include/evmc/evmc.hpp
```diff
@@ -636,7 +634,8 @@ inline void emit_log(evmc_context* h,
                      const evmc_bytes32 topics[],
                      size_t num_topics) noexcept
 {
-    static_cast<Host*>(h)->emit_log(*addr, data, data_size, topics, num_topics);
+    static_cast<Host*>(h)->emit_log(*addr, data, data_size, static_cast<const bytes32*>(topics),
```
axic: Is this a safe cast?

chfast: This is a kind of downcast. It is safe provided the size of `evmc::bytes32` is the same as `evmc_bytes32`. I can add `static_cast` here.

axic: Whatever is better.

PR #378, opened 2019-08-05
cmake: Configure source package

chfast:


axic:
@chfast this for 6.3?

chfast:
> @chfast this for 6.3?

Well, it works so we can include it. But no CI is producing the source package. I expected more from it in general (the ability to specify components that goes into the source package) but this is the most we can get. Still you can build source package on your own without git.

I plan to push them to GitHub releases. For 6.3 we can do it manually. But maybe I will just store them as CI artifacts.

Any comments about naming convention?

axic:
> Any comments about naming convention?

`-source` is fine.

CMakeLists.txt
```diff
@@ -88,11 +88,25 @@ endif()
...
+
+set(
+    CPACK_SOURCE_IGNORE_FILES
+    /\\\\.git/
+    /\\\\.idea/
+    /build/
+    /cmake-build-.*/
+    /bindings/
+    /Cargo.toml
+)
```
axic: Also add `/target` to ignore everything Rust related.

CMakeLists.txt
```diff
@@ -88,11 +88,26 @@ endif()
...
 set(CPACK_INCLUDE_TOPLEVEL_DIRECTORY FALSE)
+
+set(
+    CPACK_SOURCE_IGNORE_FILES
+    /\\\\.git/
+    /\\\\.idea/
+    /build/
+    /cmake-build-.*/
+    /bindings/
+    /Cargo.toml
```
axic: Sorry, `Cargo.lock` too, see the contents of `.gitignore`.

PR #383, opened 2019-08-08
examples: extend example to support dynamic loading

axic:
Follow up of #365.

The reason I've started doing these changes is two fold:
- to give a better starter example for anyone implementing evmc host (e.g. learning from Solidity's trial and error)
- to test Rust example on CI

### DIFF
--- a/examples/example.c
+++ b/examples/example.c
@@ -4,18 +4,34 @@
  */
 
 #include "example_host.h"
+#ifdef STATICALLY_LINKED_EXAMPLE
 #include "example_vm/example_vm.h"
+#endif
 
 #include <evmc/helpers.h>
+#include <evmc/loader.h>
 
 #include <inttypes.h>
 #include <stdio.h>
 
 int main()
 {
+#ifdef STATICALLY_LINKED_EXAMPLE
     struct evmc_instance* vm = evmc_create_example_vm();
+    if (!vm)
+        return EVMC_LOADER_INSTANCE_CREATION_FAILURE;
     if (!evmc_is_abi_compatible(vm))
-        return -1;
+        return EVMC_LOADER_ABI_VERSION_MISMATCH;
+#else
+    enum evmc_loader_error_code error_code;
+    struct evmc_instance* vm = evmc_load_and_configure("example-vm.so", &error_code);
+    if (!vm)
+    {
+        printf("Loading error: %d\n", error_code);
+        // NOTE: the values are small enough for casting
+        return (int)error_code;
+    }
+#endif
 
     // EVM bytecode goes here. This is one of the examples.
     const uint8_t code[] = "\x30\x60\x00\x52\x59\x60\x00\xf3";
--- a/examples/example.c
+++ b/examples/example.c
@@ -14,17 +14,20 @@
 #include <inttypes.h>
 #include <stdio.h>
 
-int main()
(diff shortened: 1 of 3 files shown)


examples/example.c
```diff
@@ -4,18 +4,33 @@
...
 {
+#ifdef STATICALLY_LINKED_EXAMPLE
     struct evmc_instance* vm = evmc_create_example_vm();
+    if (!vm)
+        return -1;
     if (!evmc_is_abi_compatible(vm))
         return -1;
+#else
+    enum evmc_loader_error_code error_code;
+    struct evmc_instance *vm = evmc_load_and_configure("example-vm.so", &error_code);
```
axic: @chfast can we extend the loader to check for dll extensions (`.so`, `.dylib`, `.dll`) so we can avoid declaring them in the load string?

axic: Or perhaps pass the path in `argv[1]` here.

chfast: `argv[1]` sounds good.

examples/example.c
```diff
@@ -4,18 +4,36 @@
...
+#ifdef STATICALLY_LINKED_EXAMPLE
+    (void)argc;
+    (void)argv;
     struct evmc_instance* vm = evmc_create_example_vm();
+    if (!vm)
+        return -1;
     if (!evmc_is_abi_compatible(vm))
         return -1;
+#else
+    char* config_string = (argc > 1) ? argv[1] : "example-vm.so";
```
chfast:
```suggestion
    const char* config_string = (argc > 1) ? argv[1] : "./example-vm.so";
```

examples/example.c
```diff
@@ -4,18 +4,36 @@
...
+    enum evmc_loader_error_code error_code;
+    struct evmc_instance* vm = evmc_load_and_configure(config_string, &error_code);
+    if (!vm)
+    {
+        printf("Loading error: %d\n", error_code);
+        return -1;
```
chfast:
```suggestion
        return error_code;
```

axic: `error_code` are positive, ok with that? (should be fine as it is non-zero).

axic:
@chfast updated

examples/example.c
```diff
@@ -4,18 +4,36 @@
...
 
-int main()
+int main(int argc, char* argv[])
 {
+#ifdef STATICALLY_LINKED_EXAMPLE
+    (void)argc;
+    (void)argv;
     struct evmc_instance* vm = evmc_create_example_vm();
+    if (!vm)
+        return -1;
```
chfast:
```suggestion
        return EVMC_LOADER_INSTANCE_CREATION_FAILURE;
```

examples/example.c
```diff
@@ -4,18 +4,36 @@
...
     if (!evmc_is_abi_compatible(vm))
         return -1;
```
chfast:
```suggestion
        return EVMC_LOADER_ABI_VERSION_MISMATCH;
```

examples/example.c
chfast: Success is 0. The rest is positive. I think this is ok from OS point of view. Even better than negative I believe, but not expert here.

examples/example.c
axic: Ok, added them.

PR #388, opened 2019-08-08
examples: improve tx_context handling and extend examples

axic:


examples/example_host.cpp
```diff
@@ -25,8 +25,12 @@ struct account
 class ExampleHost : public evmc::Host
 {
     std::map<evmc::address, account> accounts;
+    evmc_tx_context tx_context = evmc_tx_context{};
 
 public:
+    ExampleHost() = default;
+    ExampleHost(evmc_tx_context& _tx_context) : tx_context(_tx_context){};
```
chfast:
```suggestion
    explicit ExampleHost(evmc_tx_context& _tx_context) noexcept : tx_context{_tx_context}{};
```

examples/example_host.cpp
```diff
@@ -25,8 +25,12 @@ struct account
 class ExampleHost : public evmc::Host
 {
     std::map<evmc::address, account> accounts;
+    evmc_tx_context tx_context = evmc_tx_context{};
```
chfast:
```suggestion
    evmc_tx_context tx_context{};
```

examples/example_host.cpp
axic: Fixed.

examples/example_host.cpp
```diff
@@ -25,8 +25,12 @@ struct account
 class ExampleHost : public evmc::Host
 {
     std::map<evmc::address, account> accounts;
+    evmc_tx_context tx_context{};
 
 public:
+    explicit ExampleHost() = default;
```
chfast: The `explicit` is only needed for constructors with single argument.

axic: Fixed.

PR #391, opened 2019-08-08
rust: implement the save_return_block_number example in example-rust-vm

axic:
Follow up of #388. Depends on #389.

### DIFF
--- a/examples/example-rust-vm/src/lib.rs
+++ b/examples/example-rust-vm/src/lib.rs
@@ -21,12 +21,33 @@ impl EvmcVm for ExampleRustVM {
         message: &ExecutionMessage,
         _context: &mut ExecutionContext,
     ) -> ExecutionResult {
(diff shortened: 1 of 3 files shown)


examples/example-rust-vm/src/lib.rs
```diff
@@ -21,12 +21,33 @@ impl EvmcVm for ExampleRustVM {
...
+
+        assert!(tx_context.block_number <= 255);
+        let block_number = tx_context.block_number as u8;
+
+        let storage_key = Bytes32::default();
+        let mut storage_value = Bytes32::default();
+        storage_value.bytes[31] = block_number;
+        _context.set_storage(&message.destination(), &storage_key, &storage_value);
+
+        let ret = format!("{}", block_number).into_bytes();
```
jakelang: Why do we need to convert the blocknum to a human-readable string?

axic: To match example-vm in C.

PR #393, opened 2019-08-09
rust: catch exceptions in the entire execute() wrapper

axic:


axic:
@chfast @jakelang this was a mistake and should be reverted. In case of an inner panic, the container ownership is not returned.

PR #400, opened 2019-08-10
Revert "rust: catch exceptions in the entire execute() wrapper"

axic:
Reverts #393.

### DIFF
--- a/bindings/rust/evmc-declare/src/lib.rs
+++ b/bindings/rust/evmc-declare/src/lib.rs
@@ -337,40 +337,34 @@ fn build_execute_fn(names: &VMNameSet) -> proc_macro2::TokenStream {
         {
             use evmc_vm::EvmcVm;
 
+            assert!(!instance.is_null());
+            // TODO: context is optional in case of the "precompiles" capability
+            assert!(!context.is_null());
+            assert!(!msg.is_null());
+
+            let execution_message: ::evmc_vm::ExecutionMessage = unsafe {
+                msg.as_ref().expect("EVMC message is null").into()
+            };
+
+            let empty_code = [0u8;0];
+            let code_ref: &[u8] = if code.is_null() {
+                assert_eq!(code_size, 0);
+                &empty_code
+            } else {
+                unsafe {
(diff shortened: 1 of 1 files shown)


axic:
> With the additional commit, this much more than revert.

Yes, it is better than a revert. Achieves the original goal :wink:

bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -337,40 +341,40 @@ fn build_execute_fn(names: &VMNameSet) -> proc_macro2::TokenStream {
...
+            // TODO: context is optional in case of the "precompiles" capability
+            if instance.is_null() || context.is_null() || msg.is_null() || (code.is_null() && code_size != 0) {
+                // These are irrecoverable errors that violate the EVMC spec.
+                std::process::abort();
+            }
 
-                let execution_message: ::evmc_vm::ExecutionMessage = unsafe {
-                    msg.as_ref().expect("EVMC message is null").into()
-                };
+            assert!(!instance.is_null());
```
jakelang: The above conditional should handle all three of these asserts.

bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -337,40 +341,40 @@ fn build_execute_fn(names: &VMNameSet) -> proc_macro2::TokenStream {
...
+            let execution_message: ::evmc_vm::ExecutionMessage = unsafe {
+                msg.as_ref().expect("EVMC message is null").into()
+            };
 
-                let container = unsafe {
-                    ::evmc_vm::EvmcContainer::<#type_name_ident>::from_ffi_pointer(instance)
-                };
+            let empty_code = [0u8;0];
+            let code_ref: &[u8] = if code.is_null() {
+                assert_eq!(code_size, 0);
```
jakelang: Also handled by the first conditional.

bindings/rust/evmc-declare/src/lib.rs
axic: I know, though I saw no harm keeping the asserts.

PR #401, opened 2019-08-10
rust: change ownership api in container 

axic:
Depends on #400.

@jakelang this is a cleanup I think it makes it easier to understand.


bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -314,7 +317,12 @@ fn build_destroy_fn(names: &VMNameSet) -> proc_macro2::TokenStream {
 
     quote! {
         extern "C" fn __evmc_destroy(instance: *mut ::evmc_vm::ffi::evmc_instance) {
+            if instance.is_null() {
+                // This is an irrecoverable error that violates the EVMC spec.
+                std::process::abort();
+            }
             unsafe {
+                // Acquire ownership from EVMC. This will deallocate it also and the end of the scope.
```
jakelang: 2nd sentence in this comment looks a bit funky.

axic: "it also at the end"

PR #404, opened 2019-08-16
go: add LoadAndConfigure method

axic:


axic:
> Should this be included in 6.3.1? If yes, target release/6.3 branch. I believe we should have it in the EVMC PR in geth.

Semantic versioning wise 6.4.0 would be more appropriate because it is a new feature, but I'm fine with 6.3.1

chfast:
> Semantic versioning wise 6.4.0 would be more appropriate because it is a new feature, but I'm fine with 6.3.1

Let's have in 6.3.1 then.

axic:
> Let's have in 6.3.1 then.

@chfast done.

PR #406, opened 2019-08-17
cpp: Fallback to evmc::is_zero() in case the deprecated ::is_zero() is used

chfast:
The deprecated is_zero() is removed, but evmc.hpp is included in helpers.hpp in a way that allows using is_zero() as before. This is more user-friendly solution because now evmc::is_zero() can be used also for C types. Before, the deprecated overloading was selected by the compiler.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ and this project adheres to [Semantic Versioning].
 
 - Added `LoadAndConfigure` method to the Go API.
   [[#404](https://github.com/ethereum/evmc/pull/404)]
+  
+### Deprecated
+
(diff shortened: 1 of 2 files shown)


axic:
I think it should still remain deprecated if you want users' to move to the other header. However, if 7.0 just breaks compatibility without this warning, then this can be merged.

chfast:
This change makes that less changes are needed in Hera, but it is not a big deal. On the other hand, we cannot deprecate a header in a compiler.

chfast:
@gumb0 @axic any final opinions?

axic:
I think either way is fine, evmc 7 will be a breaking change and can remove all these.

chfast:
Ok, I'm merging this then.

PR #408, opened 2019-08-20
Attempt to fix Rust CI

axic:


chfast:
Fixed in https://github.com/ethereum/evmc/pull/409.

PR #417, opened 2019-09-02
Gas refunded

### LINKED ISSUE #413: Calculate refunds on VM side
It should be VM's responsibility to calculate and aggregate refunds.

- [ ] Add "refund" filed to `evmc_result`.

chfast:
Closes #413

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -315,6 +315,12 @@ enum evmc_status_code
     EVMC_OUT_OF_MEMORY = -3
 };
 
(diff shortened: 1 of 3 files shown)


chfast:
Hey @axic, I'm not sure if we discussed this, but having "destructed accounts" list is also not enough. `SELFDESTRUCT` also performs ETH transfer, and it has to happen immediately.

axic:
I'm just pushed a piece of commit dated from when we last discussed, before it gets lost. That's all.

PR #420, opened 2019-09-09
Change ChainID width to 64 bit

gumb0:
https://github.com/ethereum/EIPs/pull/2263

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ and this project adheres to [Semantic Versioning].
 ### Added
 
 - Support for Istanbul EIP-1344 (CHAINID opcode). `chain_id` added to `evmc_tx_context` struct.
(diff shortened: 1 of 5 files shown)


gumb0:
Rebased.

chfast:
Can this be closed now?

gumb0:
Well there's a separate EIP for this now https://github.com/ethereum/EIPs/issues/2294

It might or might not be included in Berlin :man_shrugging:

chfast:
Can we close this now? The 256-bit variant is the current EIP spec.

axic:
We actually still want to apply this through EIP-1985.

chfast:
> We actually still want to apply this through EIP-1985.

Right, but the change is trivial. I don't think it's worth to keep the PR open. But we can still link to it from the EIP.

PR #421, opened 2019-09-13
cmake: Upgrade GTest to 1.8.1

chfast:
This is also fix builds with GCC9.

### DIFF
--- a/test/unittests/test_cpp.cpp
+++ b/test/unittests/test_cpp.cpp
@@ -89,24 +89,24 @@ TEST(cpp, std_maps)
 {
     std::map<evmc::address, bool> addresses;
     addresses[{}] = true;
-    ASSERT_EQ(addresses.size(), 1);
+    ASSERT_EQ(addresses.size(), size_t{1});
     EXPECT_EQ(addresses.begin()->first, evmc::address{});
 
     std::unordered_map<evmc::address, bool> unordered_addresses;
     unordered_addresses.emplace(*addresses.begin());
     addresses.clear();
-    EXPECT_EQ(unordered_addresses.size(), 1);
+    EXPECT_EQ(unordered_addresses.size(), size_t{1});
     EXPECT_FALSE(unordered_addresses.begin()->first);
 
     std::map<evmc::bytes32, bool> storage;
     storage[{}] = true;
-    ASSERT_EQ(storage.size(), 1);
+    ASSERT_EQ(storage.size(), size_t{1});
     EXPECT_EQ(storage.begin()->first, evmc::bytes32{});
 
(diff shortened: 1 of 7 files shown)


test/unittests/test_cpp.cpp
```diff
@@ -89,24 +89,24 @@ TEST(cpp, std_maps)
...
     storage[{}] = true;
-    ASSERT_EQ(storage.size(), 1);
+    ASSERT_EQ(storage.size(), size_t{1});
     EXPECT_EQ(storage.begin()->first, evmc::bytes32{});
 
     std::unordered_map<evmc::bytes32, bool> unordered_storage;
     unordered_storage.emplace(*storage.begin());
     storage.clear();
-    EXPECT_EQ(unordered_storage.size(), 1);
+    EXPECT_EQ(unordered_storage.size(), size_t{1});
```
axic: Should this be assert though?

test/unittests/test_cpp.cpp
```diff
@@ -89,24 +89,24 @@ TEST(cpp, std_maps)
...
     addresses[{}] = true;
-    ASSERT_EQ(addresses.size(), 1);
+    ASSERT_EQ(addresses.size(), size_t{1});
     EXPECT_EQ(addresses.begin()->first, evmc::address{});
 
     std::unordered_map<evmc::address, bool> unordered_addresses;
     unordered_addresses.emplace(*addresses.begin());
     addresses.clear();
-    EXPECT_EQ(unordered_addresses.size(), 1);
+    EXPECT_EQ(unordered_addresses.size(), size_t{1});
```
axic: Same here.

test/unittests/test_cpp.cpp
chfast: Fixed.

test/unittests/test_cpp.cpp
```diff
@@ -89,24 +91,24 @@ TEST(cpp, std_maps)
 {
     std::map<evmc::address, bool> addresses;
     addresses[{}] = true;
-    ASSERT_EQ(addresses.size(), 1);
+    ASSERT_EQ(addresses.size(), size_t{1});
```
gumb0: It's annoying to do the cast every time, is it fixing the warning?
Do you not like `1u` better?

chfast: Somehow combination of GTest 1.8.1 and Visual Studio 2015 produces warnings about signed/unsigned comparison. Hopefully, we could drop VS 2015 at some point.

It could have been `1u`, I'm just not used to using them because `ul` / `unsigned long` has different size on VS and GCC/clang.

chfast:
@axic final word?

PR #422, opened 2019-09-13
Officially drop C90 standard support

chfast:
The support for C90 was accidentally broken in v6 in https://github.com/ethereum/evmc/commit/280ff9936308c2d07687bdf833485c5c63098d6f - C90 does not have _Bool type.
I don't see it as a problem in practice.


axic:
Should this be merged into `release/6.3` too ?

chfast:
> Should this be merged into `release/6.3` too ?

It only causes issues when building compilation tests in clang 10. So _no_ for now.

PR #423, opened 2019-09-13
Clang tidy config

chfast:
This adds first revision of clang-tidy config. No automatic checking is planned at this point. This is rather for experimenting / IDE integration. And to be exported to other projects.


axic:
@chfast is this OK to merge?

chfast:
> @chfast is this OK to merge?

Ok just for having a first version of clang-tidy config. Only small portion of warnings is fixed and some other adjustments to the config might be needed.

test/vmtester/vmtester.cpp
```diff
@@ -65,7 +65,8 @@ class cli_parser
                 arguments.emplace_back(std::move(arg));
                 continue;
             }
-            else if (x <= 2)
+
+            if (x <= 2)
```
axic: The `else if` version is more readable, though the `continue` statement above makes sure the new version is semantically equivalent.

test/unittests/test_cpp.cpp
```diff
@@ -254,7 +254,7 @@ TEST(cpp, result)
 
         EXPECT_EQ(release_called, 1);
     }
-    EXPECT_EQ(release_called, 1);
+    EXPECT_EQ(release_called, 1);  // NOLINT(clang-analyzer-unix.Malloc)
```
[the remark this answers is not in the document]
chfast: I thinks it is just because this is the last line in the scope. Maybe if put next line it will also work.

examples/example_precompiles_vm/example_precompiles_vm.cpp
```diff
@@ -52,8 +52,8 @@ static evmc_result execute(evmc_vm* /*unused*/,
                            evmc_host_context* /*unused*/,
                            enum evmc_revision rev,
                            const evmc_message* msg,
-                           const uint8_t*,
-                           size_t)
+                           const uint8_t* /*unused*/,
```
chfast: Is this fine or too verbose?

test/unittests/test_cpp.cpp
```diff
@@ -502,7 +502,7 @@ TEST(cpp, result_raii)
 {
     static auto release_called = 0;
     release_called = 0;
-    auto release_fn = [](const evmc_result*) noexcept { ++release_called; };
+    auto release_fn = [](const evmc_result* /*unused*/) noexcept { ++release_called; };
```
chfast: Is this fine or too verbose?

examples/example_precompiles_vm/example_precompiles_vm.cpp
gumb0: I think too verbose. I see some value in commenting what is exactly unused, like `/*code*/` (probably not possible with clang-tidy), but `unused` doesn't add any new info.

test/vmtester/vmtester.cpp
chfast: This is this check: https://clang.llvm.org/extra/clang-tidy/checks/readability-else-after-return.html.

test/unittests/test_cpp.cpp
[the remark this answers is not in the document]
chfast: Not needed any more.

examples/example_precompiles_vm/example_precompiles_vm.cpp
chfast: Apparently, you can add any comment, even `/**/`.

gumb0: cool

gumb0: Ah it just allows these manually typed comments... For some reason I thought it can insert them automatically

chfast: No. It will insert `/*unused*/` as a fix.

chfast:
It is ok to land it, or you have more suggestions?

examples/example_precompiles_vm/example_precompiles_vm.cpp
```diff
@@ -75,32 +75,14 @@ static evmc_result execute(evmc_vm* /*unused*/,
     switch (id)
     {
     case 0x0001:  // ECDSARECOVER
-        return not_implemented();
```
gumb0: Why do you execute identity now for addresses 0x1-0x3 ?

examples/example_precompiles_vm/example_precompiles_vm.cpp
```diff
@@ -117,9 +99,9 @@ evmc_vm* evmc_create_example_precompiles_vm()
         EVMC_ABI_VERSION,
         "example_precompiles_vm",
         PROJECT_VERSION,
-        [](evmc_vm*) {},
+        [](evmc_vm* /*vm*/) {},
```
gumb0: these and the ones with `evmc_result` are kind of tautological... not sure if better just to allow no comment

test/unittests/test_cpp.cpp
```diff
@@ -353,8 +353,9 @@ TEST(cpp, vm_null)
 TEST(cpp, vm_move)
 {
     static int destroy_counter = 0;
-    const auto template_vm = evmc_vm{
-        EVMC_ABI_VERSION, "", "", [](evmc_vm*) { ++destroy_counter; }, nullptr, nullptr, nullptr};
+    const auto template_vm =
+        evmc_vm{EVMC_ABI_VERSION, "",      "",     [](evmc_vm* /*vm*/) { ++destroy_counter; },
```
gumb0: I'm not a fan of this kind of indentation :)

examples/example_precompiles_vm/example_precompiles_vm.cpp
chfast: clang-tidy is complaining that many switch cases have the same code.

test/unittests/test_cpp.cpp
chfast: This is up to clang-format config. I actually was considering changing this to be aligned with other C++ projects.

examples/example_precompiles_vm/example_precompiles_vm.cpp
chfast: I agree here.

examples/example_precompiles_vm/example_precompiles_vm.cpp
gumb0: Should be
```
    case 0x0001:  // ECDSARECOVER
    case 0x0002:  // SHA256
    case 0x0003:  // RIPEMD160
        return not_implemented();
    case 0x0004:  // Identity
        return execute_identity(msg);
```

test/unittests/test_cpp.cpp
chfast: I cannot find the option responsible for it, if any exists.

PR #424, opened 2019-09-13
Drop 'const' from the parameter type of evmc_result::release()

### LINKED ISSUE #278: Remove const from parameter to evm_result::release

chfast:
Closes #278.


chfast:
This looks to be useless change - a const cast for `free()` is still needed and gives no benefits to C++ nor Rust.

PR #425, opened 2019-09-19
instructions: Set gas cost for undefined instructions to 0

### LINKED ISSUE #356: Instructions: consider definding undefined instruction's cost as 0 instead of -1
https://github.com/ethereum/evmc/blob/22ed657d0c49e0bd1c50f86277e911e253bc196a/lib/instructions/instruction_metrics.c#L9

`-1` might lead to some weird results in VM implementations, like block's gas cost going down when counted in pre-execution analysis, as seen in https://github.com/ethereum/evmone/pull/93

chfast:
Closes #356.

### DIFF
--- a/lib/instructions/instruction_metrics.c
+++ b/lib/instructions/instruction_metrics.c
@@ -5,8 +5,13 @@
 
 #include <evmc/instructions.h>
 
-/** Marks an instruction as undefined. */
-#define UNDEFINED -1
+/**
+ * Marks an instruction as undefined.
+ *
+ * The gas cost for undefined instructions is 0 because this is the cost of executing them
+ * in practice in EVM implementations.
+ */
+#define UNDEFINED 0
 
 /**
  * Gas price tiers, names from Yellow Paper.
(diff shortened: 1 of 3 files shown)


lib/instructions/instruction_metrics.c
```diff
@@ -277,7 +277,7 @@ static struct evmc_instruction_metrics istanbul_metrics[256] = {
     /*                = 0xfc */ {UNDEFINED, 0, 0},
     /*         REVERT = 0xfd */ {ZERO, 2, 0},
     /*        INVALID = 0xfe */ {ZERO, 0, 0},
-    /*        SUICIDE = 0xff */ {5000, 1, 0},
+    /*   SELFDESTRUCT = 0xff */ {5000, 1, 0},
```
axic: Wow this wasn't done yet? :)

chfast: Only in comments. This was created by transforming Aleth's table / enum.

PR #427, opened 2019-09-20
Move evmc_host_interface out of evmc_host_context

chfast:
On the C API level (and also ABI) the `host_interface` (think: an object vtable) and `host_context` (think: pointer to an object's data) are separated. This actually makes creating language bindings easier - no need for workarounds as in https://github.com/ethereum/evmc/blob/master/bindings/go/evmc/evmc.go#L26-L30.

Language bindings can still hide this by providing single OOP interface for Host (both for Client and VM side).

### DIFF
--- a/bindings/rust/evmc-declare/src/lib.rs
+++ b/bindings/rust/evmc-declare/src/lib.rs
@@ -346,14 +346,14 @@ fn build_execute_fn(names: &VMNameSet) -> proc_macro2::TokenStream {
             use evmc_vm::EvmcVm;
 
             // TODO: context is optional in case of the "precompiles" capability
-            if instance.is_null() || host.is_null() || context.is_null() || msg.is_null() || (code.is_null() && code_size != 0) {
+            if instance.is_null() || host.is_null() || msg.is_null() || (code.is_null() && code_size != 0) {
                 // These are irrecoverable errors that violate the EVMC spec.
                 std::process::abort();
             }
 
             assert!(!instance.is_null());
-            // TODO: context is optional in case of the "precompiles" capability
-            assert!(!context.is_null());
+            // TODO: host is optional in case of the "precompiles" capability
+            assert!(!host.is_null());
             assert!(!msg.is_null());
 
             let execution_message: ::evmc_vm::ExecutionMessage = unsafe {
@@ -379,7 +379,7 @@ fn build_execute_fn(names: &VMNameSet) -> proc_macro2::TokenStream {
                 let mut execution_context = unsafe {
                     ::evmc_vm::ExecutionContext::new(
                         host.as_ref().expect("EVMC host is null"),
-                        context.as_mut().expect("EVMC context is null")
+                        context,
                     )
                 };
                 container.execute(revision, code_ref, &execution_message, &mut execution_context)
--- a/bindings/rust/evmc-declare/src/lib.rs
+++ b/bindings/rust/evmc-declare/src/lib.rs
@@ -335,6 +335,7 @@ fn build_execute_fn(names: &VMNameSet) -> proc_macro2::TokenStream {
     quote! {
         extern "C" fn __evmc_execute(
             instance: *mut ::evmc_vm::ffi::evmc_vm,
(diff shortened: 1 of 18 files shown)


chfast:
@axic Do you want me to finish Rust part here?

chfast:
@axic @jakelang can you finish Rust changes here, please?

axic:
@chfast fixed compilation, but haven't reviewed the changes yet.

bindings/rust/evmc-vm/src/container.rs
```diff
@@ -132,7 +132,7 @@ mod tests {
             get_block_hash: None,
             emit_log: None,
         };
-        let mut host_context = ::evmc_sys::evmc_host_context { _unused: [] };
+        let mut host_context = ::evmc_sys::evmc_host_context::default();
```
chfast: Is it possible to pass a null pointer instead of creating host_context object?

chfast:
Ok, thanks a lot. I will clean up git history and put it for review.

bindings/rust/evmc-vm/src/container.rs
axic: Yeah, that should be an option too.

axic: Just tried it and it feels like to be simple leaving it as is, for now.

include/evmc/evmc.h
```diff
@@ -156,6 +156,16 @@ struct evmc_tx_context
     evmc_uint256be chain_id;         /**< The blockchain's ChainID. */
 };
 
+/**
+ * @struct evmc_host_context
+ * Execution context managed by the Host.
+ *
+ * The Host MUST pass the pointer to the execution context to ::evmc_execute_fn.
+ * The VM MUST pass the same pointer back to the Host in every callback function.
+ * The context MUST contain at least the function table defining
```
axic: Why is this requirement and how can it be enforced? :)

include/evmc/evmc.h
```diff
@@ -802,6 +796,7 @@ enum evmc_revision
  * @return           The execution result.
```
axic: Need to add host to the param list. Also mentioned it may be null.

bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -345,7 +346,7 @@ fn build_execute_fn(names: &VMNameSet) -> proc_macro2::TokenStream {
             use evmc_vm::EvmcVm;
 
             // TODO: context is optional in case of the "precompiles" capability
-            if instance.is_null() || context.is_null() || msg.is_null() || (code.is_null() && code_size != 0) {
+            if instance.is_null() || host.is_null() || context.is_null() || msg.is_null() || (code.is_null() && code_size != 0) {
```
axic: Depending on https://github.com/ethereum/evmc/pull/427#discussion_r342330665 we may need to remove `context.is_null()` here.

bindings/rust/evmc-vm/src/container.rs
chfast: I tried it too, but feels like null pointers is Rust's advanced feature.

include/evmc/evmc.h
chfast: Sorry, this has to be updated. Now:

> Host MAY pass the pointer...
> VM MUST pass the same pointer back...

The last one is not needed any more, as the "context callback interface" has been separated.

bindings/rust/evmc-declare/src/lib.rs
chfast: The `context` may always be null. VM never needs to dereference it.
Now the TODO comment should be applied to the `host` as in the case of "precompiles" it is not useful and may be null — this can be discussed further in https://github.com/ethereum/evmc/pull/350.

axic: Fixed the `context` part. Lets leave `host` for #350.

axic:
@chfast simplified it to deal with null pointers. Also allows `context` to be null now.

PR #428, opened 2019-09-24
test: Rework cpp.result unit test

chfast:


gumb0:
Out of curiosity, what needs this `evmc_result_optional_storage`?

chfast:
> Out of curiosity, what needs this `evmc_result_optional_storage`?

Aleth is using this to store `std::vector`.

PR #430, opened 2019-09-24
Rename evmc_instance -> evmc_vm

chfast:


lib/loader/loader.c
```diff
@@ -207,29 +206,29 @@ struct evmc_instance* evmc_load_and_create(const char* filename,
...
     enum evmc_loader_error_code ec = EVMC_LOADER_SUCCESS;
 
-    struct evmc_instance* instance = create_fn();
-    if (!instance)
+    struct evmc_vm* vm = create_fn();
+    if (!vm)
     {
-        ec = set_error(EVMC_LOADER_INSTANCE_CREATION_FAILURE,
-                       "creating EVMC instance of %s has failed", filename);
+        ec = set_error(EVMC_LOADER_VM_CREATION_FAILURE, "creating EVMC VM of %s has failed",
```
chfast: I'm not sure about this error message.

include/evmc/evmc.hpp
```diff
@@ -408,12 +408,12 @@ class VM
     }
 
 private:
-    evmc_instance* m_instance = nullptr;
+    evmc_vm* m_instance = nullptr;
```
chfast: I have left the name `m_instance` because `m_vm` would be worse. This can also be `m_handle` (as in Go) or `m_vm_ptr`.

lib/loader/loader.c
gumb0: would be great if it could mention the name of the function it called, but it doesn't look easy

axic:
What was the reasoning behind this?

chfast:
> What was the reasoning behind this?

Looks better. Instance is like `class Class` and comes from old days when this project was about VM concept only.

PR #432, opened 2019-09-25
ci: Docker images upgrades

chfast:


circle.yml
```diff
@@ -109,20 +110,27 @@ jobs:
...
+    docker:
+      - image: ethereum/cpp-build-env:12-gcc-8
     environment:
-      CC: gcc-8
-      CXX: g++-8
       CMAKE_OPTIONS: -DTOOLCHAIN=cxx17-pic
     steps:
       - build_and_test
 
+  build-cxx17-ubsan:
```
gumb0: Maybe it would make sense to always include compiler in the name, so `build-gcc9-cxx17-ubsan` ?

PR #433, opened 2019-09-26
ci: Add VS 2019 build on AppVeyor

### LINKED ISSUE #415: Add Visual Studio 2019 to CI

### LINKED ISSUE #416: EVMC does not compile in Visual Studio 2019
The problem is in C++ user-defined literals.
Reproduced here: https://ci.appveyor.com/project/chfast/evmc/builds/27017625/job/uhjsf9gs5b7gc95s
Compiler bug: https://developercommunity.visualstudio.com/content/problem/553839/vs-2019-incorrect-partial-template-specialization.html

chfast:
Closes #415
Closes #416.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,7 +10,8 @@ configuration:
   - Release
 environment:
   matrix:
-    - VS: 2017
+    - VS: 2019
+      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
     - VS: 2017-32bit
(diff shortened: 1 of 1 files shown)


axic:
> > Do we still need the 2015 and 2017 stuff in here? Or is that boilerplate?
> 
> We need it to check if it still builds with older versions. Portability is important for this project.

Oh wow there's this new fancy feature of quote reply? :)

PR #435, opened 2019-09-27
Change values in instruction metrics

chfast:
This updates instruction metrics with changed values - they are easier to explain and represent number directly usable by VM implementations (will benefit aleth-interepreter).

This is tested against evmone's table: https://github.com/ethereum/evmone/pull/191.

The modification can be easily verified by following the rule:
```
{..., a, b} -> {..., a, b - a}
```

Continuation of https://github.com/ethereum/evmc/issues/356.


include/evmc/instructions.h
```diff
@@ -188,11 +188,11 @@ struct evmc_instruction_metrics
     /** The instruction gas cost. */
     int16_t gas_cost;
 
-    /** The number of items the instruction pops from the EVM stack before execution. */
-    int8_t num_stack_arguments;
+    /** The number of the EVM stack items required for the instruction. */
```
gumb0: "the minimum number" maybe?

include/evmc/instructions.h
```diff
@@ -188,11 +188,11 @@ struct evmc_instruction_metrics
...
+    int8_t stack_height_required;
 
-    /** The number of items the instruction pushes to the EVM stack after execution. */
-    int8_t num_stack_returned_items;
+    /** The EVM stack height change caused by the instruction execution. */
```
gumb0: I'd clarify that it's stack after execution - stack before

chfast: Added.

include/evmc/instructions.h
chfast: Changed.

PR #442, opened 2019-11-04
helpers: Return bool from is_abi_compatible()

chfast:


chfast:
> `evmc_account_exists_fn` already returns a bool, so if there were any issues with it, we should have seen it.

Correct. The main `evmc.h` is using C `_Bool` / `<stdbool.h>` for some time. Also `helpers.h` is using `bool` in other places.

PR #445, opened 2019-11-06
Tweaks to C++ API

chfast:
- [x] Add unit tests

### DIFF
--- a/include/evmc/evmc.hpp
+++ b/include/evmc/evmc.hpp
@@ -596,8 +596,8 @@ class Host : public HostInterface
 {
 public:
     /// Provides access to the global host interface.
-    /// @returns  Pointer to the host interface object.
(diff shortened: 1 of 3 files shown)


include/evmc/evmc.hpp
```diff
@@ -408,6 +410,12 @@ class VM
         return result{m_instance->execute(m_instance, &host, ctx, rev, &msg, code, code_size)};
     }
 
+    inline result execute(Host& host,
```
gumb0: The rest of the methods here are inline in the class, why define this outside of class?

gumb0: Should it have some doxygen comment?

chfast: Because the implementation requires Host and Host is defined after VM. Maybe reordering the definitions can help, but that will require another PR to have diff readable.

Issue #447, opened 2019-11-06
Add "runner" tool to execute VMs


axic:
Basically the equivalent of "evmbin" many clients have.

This could start simple with only accepted two inputs:
- bytecode to run
- calldata

And hardcoding everything else. Later it could be extended to accept more settings or even store the state in a yaml/json file (aka genesis).

This would reuse `examples/example_host.c` and I'd call the tool `evmc-runner` and place it into `test/runner` or `tools/runner`.

chfast:
This sounds good, but I'm not sure I want to add more dependencies to this project. And also it duplicates some work at least from `evmone-bench`.

axic:
I'd need this for Hera and it is awkward getting "EVMC tools" from evmone.

PR #448, opened 2019-11-06
Use ethash::keccak in example_host

axic:
Part of #447.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -10,11 +10,14 @@ if(NOT CMAKE_CXX_STANDARD)
 endif()
 set(CMAKE_DEBUG_POSTFIX "")
 
+hunter_add_package(ethash)
+find_package(ethash CONFIG REQUIRED)
+
 add_subdirectory(example_vm)
 add_subdirectory(example_precompiles_vm)
 
 add_library(evmc-example-host STATIC example_host.cpp)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -58,6 +58,9 @@ if(EVMC_TESTING OR EVMC_TEST_TOOLS)
 endif()
 
 if(EVMC_TESTING OR EVMC_EXAMPLES)
+    hunter_add_package(ethash)
```
axic: Had this in `examples/CMakeLists.txt` first but hoped this would fix the linking issue. Too bad there doesn't seem to be a way to statically link with ethash via `target_link_libraries`.

@chfast any simple solutions?

chfast: You need ethash 0.5.0: https://github.com/chfast/ethash/blob/master/CHANGELOG.md#050---2019-06-07.

chfast:
Why are you doing this?

axic:
> Why are you doing this?

For #447 to have a working host?

CMakeLists.txt
axic: Doesn't hunter have 0.5.0 ?

axic: Ah, hunter has 0.5.0, but evmc still uses an old hunter.

axic:
Obsoleted by #456.

PR #454, opened 2019-11-16
Fix comment on evmc_get_code_hash_fn

jrhea:
Was working on Java bindings and noticed in `include/evmc/evmc.h`:

Comment on `evmc_get_code_hash_fn` was
```
Get code size callback function.
```

changed to 

```
Get code hash callback function.
```


jrhea:
@chfast thanks for the review and approval! i'll state the obvious and mention that I dont have write access so i cannot merge.  I'm 99% sure that you were just waiting for CI to run, but I wanted to point it out just in case.

PR #456, opened 2019-11-18
Add mocked_host library

chfast:
It takes the MockedHost implementation from evmone and creates an interface library out of it.

Part of #447.

### DIFF
--- a/include/evmc/mocked_host.hpp
+++ b/include/evmc/mocked_host.hpp
@@ -5,7 +5,6 @@
 
 #include <evmc/evmc.hpp>
 #include <algorithm>
-#include <iterator>
 #include <string>
 #include <unordered_map>
 #include <vector>
@@ -76,9 +75,7 @@ public:
         /// Equal operator.
         bool operator==(const log_record& other) const noexcept
         {
-            return creator == other.creator && data == other.data &&
-                   topics.size() == other.topics.size() &&
-                   std::equal(topics.begin(), topics.end(), other.topics.begin());
+            return creator == other.creator && data == other.data && topics == other.topics;
         }
     };
 
@@ -188,21 +185,19 @@ protected:
         if (old.value == value)
             return EVMC_STORAGE_UNCHANGED;
 
-        evmc_storage_status status;
+        evmc_storage_status status{};
+        if (!old.dirty)
         {
-            if (!old.dirty)
-            {
-                old.dirty = true;
-                if (!old.value)
-                    status = EVMC_STORAGE_ADDED;
-                else if (value)
-                    status = EVMC_STORAGE_MODIFIED;
-                else
-                    status = EVMC_STORAGE_DELETED;
-            }
+            old.dirty = true;
+            if (!old.value)
+                status = EVMC_STORAGE_ADDED;
+            else if (value)
+                status = EVMC_STORAGE_MODIFIED;
             else
-                status = EVMC_STORAGE_MODIFIED_AGAIN;
+                status = EVMC_STORAGE_DELETED;
         }
+        else
+            status = EVMC_STORAGE_MODIFIED_AGAIN;
 
         old.value = value;
         return status;
@@ -311,10 +306,7 @@ protected:
                   const bytes32 topics[],
                   size_t topics_count) noexcept override
     {
-        recorded_logs.push_back({addr, {data, data_size}, {}});
-        auto& record_topics = recorded_logs.back().topics;
-        record_topics.reserve(topics_count);
-        std::copy_n(topics, topics_count, std::back_inserter(record_topics));
+        recorded_logs.push_back({addr, {data, data_size}, {topics, topics + topics_count}});
     }
 };
 }  // namespace evmc
--- a/include/evmc/mocked_host.hpp
+++ b/include/evmc/mocked_host.hpp
@@ -1,0 +1,320 @@
+// EVMC: Ethereum Client-VM Connector API.
+// Copyright 2019 The EVMC Authors.
+// Licensed under the Apache License, Version 2.0.
+#pragma once
+
+#include <evmc/evmc.hpp>
+#include <algorithm>
+#include <iterator>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace evmc
+{
+/// The string of bytes.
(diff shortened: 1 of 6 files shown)


include/evmc/mocked_host.hpp
```diff
@@ -0,0 +1,258 @@
...
+                                    const bytes32& value) noexcept override
+    {
+        record_account_access(addr);
+        const auto it = accounts.find(addr);
+        if (it == accounts.end())
+            return EVMC_STORAGE_UNCHANGED;
+
+        auto& old = it->second.storage[key];
+
+        // Follow https://eips.ethereum.org/EIPS/eip-1283 specification.
```
axic: 2200 ?

chfast: No.

include/evmc/mocked_host.hpp
```diff
@@ -0,0 +1,320 @@
...
+
+        /// The log topics.
+        std::vector<bytes32> topics;
+
+        /// Equal operator.
+        bool operator==(const log_record& other) const noexcept
+        {
+            return creator == other.creator && data == other.data &&
+                   topics.size() == other.topics.size() &&
+                   std::equal(topics.begin(), topics.end(), other.topics.begin());
```
gumb0: wouldn't `topics == other.topics` work here?

include/evmc/mocked_host.hpp
```diff
@@ -0,0 +1,320 @@
...
+        auto& old = it->second.storage[key];
+
+        // Follow https://eips.ethereum.org/EIPS/eip-1283 specification.
+        // WARNING! This is not complete implementation as refund is not handled here.
+
+        if (old.value == value)
+            return EVMC_STORAGE_UNCHANGED;
+
+        evmc_storage_status status;
+        {
```
gumb0: what's the point of this block?

include/evmc/mocked_host.hpp
```diff
@@ -0,0 +1,320 @@
...
+        if (n > 0)
+            std::copy_n(&code[code_offset], n, buffer_data);
+        return n;
+    }
+
+    /// Selfdestruct the account (EVMC host method).
+    void selfdestruct(const address& addr, const address& beneficiary) noexcept override
+    {
+        record_account_access(addr);
+        recorded_selfdestructs.push_back({addr, beneficiary});
```
gumb0: could be `emplace_back`

include/evmc/mocked_host.hpp
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    /// Emit LOG (EVMC host method).
+    void emit_log(const address& addr,
+                  const uint8_t* data,
+                  size_t data_size,
+                  const bytes32 topics[],
+                  size_t topics_count) noexcept override
+    {
+        recorded_logs.push_back({addr, {data, data_size}, {}});
```
gumb0: `emplace_back`?

gumb0: Also probably you could just initialize topics vector inline with something like `std::vector<log_record>(&topics[0], topics + topics_count)`

include/evmc/mocked_host.hpp
chfast: 🤷

include/evmc/mocked_host.hpp
chfast: `emplace_back` does not work with `{}` — I believe it needs to know the type.

include/evmc/mocked_host.hpp
chfast: I think it should. I missed the fact `std::vector` has `==`.

axic: Will you change this?

include/evmc/mocked_host.hpp
```diff
@@ -0,0 +1,320 @@
...
+
+    /// The record of all block numbers for which get_block_hash() was called.
+    std::vector<int64_t> recorded_blockhashes;
+
+    /// The record of all account accesses.
+    std::vector<address> recorded_account_accesses;
+
+    /// The maximum number of entries in recorded_account_accesses record.
+    /// This is arbitrary value useful in fuzzing when we don't want the record to explode.
+    static constexpr auto max_recorded_account_accesses = 200;
```
axic: Why not power of 2?

include/evmc/mocked_host.hpp
axic: Well 1283 is kind of irrelevant, only existed in Constantinople before Petersburg.

I know technically it is the same except the stipend, but still.

include/evmc/mocked_host.hpp
```diff
@@ -0,0 +1,320 @@
...
+            recorded_calls.emplace_back(msg);
+            auto& call_msg = recorded_calls.back();
+            if (call_msg.input_size > 0)
+            {
+                m_recorded_calls_inputs.emplace_back(call_msg.input_data, call_msg.input_size);
+                const auto& input_copy = m_recorded_calls_inputs.back();
+                call_msg.input_data = input_copy.data();
+            }
+        }
+        return result{call_result};
```
axic: If there is a single `call_result` it is not possible to deal with precompiles? Well, this function can be overriden. Have a look at the precompile stuff soltest is doing.

include/evmc/mocked_host.hpp
chfast: This is changed.

include/evmc/mocked_host.hpp
chfast: Because this value is for humans.

include/evmc/mocked_host.hpp
chfast: This PR is for moving MockedHost from evmone. I will handle solidity later.

include/evmc/mocked_host.hpp
axic: I might be blind, but it still uses `std::equal` with iterators?

chfast: Yes, but these comments are attached to the outdated code fragment.

chfast:
> I think it shouldn't refer to 1283 alone, but also to 2200.

Referencing EIP-2200 needs additional work to check its spec and it is still a draft.

PR #462, opened 2019-11-26
rust: update dependencies

axic:


bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -268,8 +267,7 @@ fn build_static_data(names: &VMNameSet, metadata: &VMMetaData) -> proc_macro2::T
 
 /// Takes a capabilities flag and builds the evmc_get_capabilities callback.
 fn build_capabilities_fn(capabilities: u32) -> proc_macro2::TokenStream {
-    let capabilities_literal =
-        LitInt::new(capabilities as u64, IntSuffix::U32, capabilities.span());
+    let capabilities_literal = LitInt::new(&capabilities.to_string(), capabilities.span());
```
axic: I wonder if `LitInt` copies the string or at least `quote` will just print it below and we don't need to take care about lifetime.

axic: Will at least make the string its own variable to ensure its lifetime.

axic:
> could use semantic versioning

cargo does a weird mix of semantic versioning, `1.0` equals to `^1.0`

bindings/rust/evmc-declare/src/lib.rs
jakelang: I am unsure because it takes a reference to an owned type but it also wouldn't make sense in this crate to expect the user to keep the members in scope.

PR #463, opened 2019-11-26
Refactoring and CMake changes around vmtester

chfast:
This is preparation for moving vmtester to `tools/` dir.
- CMake option renamed: `EVMC_TEST_TOOLS` -> `EVMC_TOOLS`.
- `vmtester` decoupled from examples: it uses `MockedHost` now.
- CMake option `EVMC_TESTING` explicitly means "build everything". The `EVMC_TOOLS` and `EVMC_EXAMPLES` are force to be `ON` in this case.


CMakeLists.txt
```diff
@@ -53,7 +57,7 @@ if(EVMC_TESTING)
     enable_testing()
 endif()
 
-if(EVMC_TESTING OR EVMC_TEST_TOOLS)
+if(EVMC_TESTING OR EVMC_TOOLS)
```
chfast: Here `EVMC_TOOLS` check would be enough as `EVMC_TESTING` implies `EVMC_TOOLS`. I just left it for clarity. The end goal is:

```
if(EVMC_TOOLS)
    add_subdirectory(tools)
endif()

if(EVMC_TESTING)
    add_subdirectory(test)
endif()
```

PR #465, opened 2019-11-27
cpp: Add VM::has_capability() method

chfast:


test/unittests/test_cpp.cpp
```diff
@@ -316,6 +316,8 @@ TEST(cpp, vm)
     EXPECT_EQ(res.status_code, EVMC_FAILURE);
 
     EXPECT_TRUE(vm.get_capabilities() & EVMC_CAPABILITY_EVM1);
```
axic: Can you add the ewasm test here too?

PR #472, opened 2019-11-28
CMake improvements 2

chfast:
This removes mandatory usage of CMake toolchain files. Instead it is now specified that C++ libraries require C++11, but the C++ standard is not forced.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -153,7 +153,7 @@ jobs:
     docker:
       - image: ethereum/cpp-build-env:12-gcc-8
     environment:
-      CMAKE_OPTIONS: -DTOOLCHAIN=cxx17-pic
+      CMAKE_OPTIONS: -DCMAKE_CXX_STANDARD=17
     steps:
       - build_and_test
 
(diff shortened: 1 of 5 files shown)


circle.yml
```diff
@@ -153,23 +153,23 @@ jobs:
     docker:
       - image: ethereum/cpp-build-env:12-gcc-8
     environment:
-      CMAKE_OPTIONS: -DTOOLCHAIN=cxx17-pic
+      CMAKE_OPTIONS: -DCMAKE_CXX_STANDARD=17
```
axic: Is this PIC by default?

chfast: Yes. All the EVMC libs are built as PIC. I made this change some time ago. None of this code is performance critical and this is convenient if someone what to include them in a shared library.

axic: I mean the cable toolchain was explicitly PIC, but I wasn't sure cmake sets PIC on by default.

chfast: CMake does not set PIC by default.
- It will set PIC when building a shared library, but only for files directly included in the library (linked in static library will not become PIC magically).
- We are enabling PIC in EVMC static libs: https://github.com/ethereum/evmc/blob/master/lib/instructions/CMakeLists.txt#L15.
- You can always enable that with `-DCMAKE_POSITION_INDEPENDENT_CODE=TRUE`.

PR #474, opened 2019-12-03
cpp: Add all comparison operators for address and bytes32

axic:


chfast:
At this point, I think we should add all other comparison operators.

axic:
@chfast is this what you meant?

Need to squash them before merging. And need to include them in the unit tests. @chfast do you want to take this over?

chfast:
> @chfast is this what you meant?
> 
> Need to squash them before merging. And need to include them in the unit tests. @chfast do you want to take this over?

Yes. And yes, I will finish this.

include/evmc/evmc.hpp
```diff
@@ -50,7 +50,6 @@ using uint256be = bytes32;
 /// Loads 64 bits / 8 bytes of data from the given @p bytes array in big-endian order.
 constexpr inline uint64_t load64be(const uint8_t* bytes) noexcept
 {
-    // TODO: Report bug in clang incorrectly optimizing this with AVX2 enabled.
```
chfast: Reported and fixed: https://bugs.llvm.org/show_bug.cgi?id=42708

axic:
Looks good to me!

PR #478, opened 2019-12-04
EVMC runner tool

### LINKED ISSUE #447: Add "runner" tool to execute VMs

chfast:
A first version of the EVMC runner tool.

1. We want a single tool `evmc` with subcommands. This one adds the main tool with `run` command.
2. Later we will move the `evmc-vmtester` here probably as `evmc check-vm`.
3. This uses CLI11 library for parsing command line options.
4. I tried to move the `run` command to separate file, but it gets messy a bit. We may try it later, but this might not be good idea in general. Moreover, building every file including `CLI.hpp` takes ~2s.
5. We should move hex helpers from evmone here. But C++17 would be nice for it to have `string_view`. There is a way to specify that only some CMake targets require C++17.

Closes #447.

### DIFF
--- a/tools/evmc/main.cpp
+++ b/tools/evmc/main.cpp
@@ -17,9 +17,14 @@ int main(int argc, const char** argv)
 
     std::string vm_config;
     std::string code_hex;
+    evmc_message msg{};
+    msg.gas = 1000000;
+
     auto& run_cmd = *app.add_subcommand("run", "Execute EVM bytecode");
-    run_cmd.add_option("--vm", vm_config)->required();
-    run_cmd.add_option("code", code_hex)->required();
+    run_cmd.add_option("code", code_hex, "Hex-encoded bytecode")->required();
+    run_cmd.add_option("--vm", vm_config, "EVMC VM module")->required();
+    run_cmd.add_option("--gas", msg.gas, "Execution gas limit", true)
+        ->check(CLI::Range(0, 1000000000));
 
     try
     {
@@ -49,8 +54,6 @@ int main(int argc, const char** argv)
             }
 
             MockedHost host;
-            evmc_message msg{};
-            msg.gas = 10000000;
 
             std::cout << "Executing on Istanbul with " << msg.gas << " gas limit\n"
                       << "in " << vm_config << "\n";
--- a/tools/evmc/main.cpp
+++ b/tools/evmc/main.cpp
@@ -3,11 +3,23 @@
 // Licensed under the Apache License, Version 2.0.
 
 #include <CLI/CLI.hpp>
+#include <evmc/loader.h>
+#include <evmc/mocked_host.hpp>
+
+#include "utils.hpp"
 
 int main(int argc, const char** argv)
 {
+    using namespace evmc;
+
     CLI::App app{"EVMC tool"};
     const auto& version_flag = *app.add_flag("--version", "Print version information");
+
+    std::string vm_config;
+    std::string code_hex;
+    auto& run_cmd = *app.add_subcommand("run", "Execute EVM bytecode");
+    run_cmd.add_option("--vm", vm_config)->required();
+    run_cmd.add_option("code", code_hex)->required();
 
     try
     {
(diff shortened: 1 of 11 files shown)


tools/evmc/main.cpp
```diff
@@ -20,14 +76,53 @@ int main(int argc, const char** argv)
...
+                else
+                    std::cerr << "Loading error " << ec << "\n";
+                return static_cast<int>(ec);
+            }
+
+            evmc::MockedHost host;
+            evmc_message msg{};
+            msg.gas = 10000000;
+
+            auto result = vm.execute(host, EVMC_ISTANBUL, msg, code.data(), code.size());
```
axic: Maybe an output message here for: `"Executing on Istanbul with " << msg.gas << "gas"`

tools/evmc/main.cpp
```diff
@@ -0,0 +1,128 @@
...
+        // Handle the --version flag first and exit when present.
+        if (version_flag)
+        {
+            std::cout << "EVMC tool " PROJECT_VERSION "\n";
+            return 0;
+        }
+
+        if (run_cmd)
+        {
+            auto code = evmc::from_hex(code_hex);
```
gumb0: const

tools/evmc/main.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+            std::cout << "Output: " << output_hex << "\n"
+                      << "Gas used: " << gas_used << "\n";
+
+            return 0;
+        }
+
+        return 0;
+    }
+    catch (const CLI::ParseError& e)
```
gumb0: This should be better only around `app.parse` call, nothing else throws `ParseError`

chfast: It can be, but isn't the point of exceptions to have single place for handling them and having the main code flow uninterrupted?

gumb0: Ok true, maybe my intuition was wrong

tools/evmc/main.cpp
```diff
@@ -0,0 +1,89 @@
...
+    std::string vm_config;
+    std::string code_hex;
+    evmc_message msg{};
+    msg.gas = 1000000;
+    auto rev = EVMC_ISTANBUL;
+
+    auto& run_cmd = *app.add_subcommand("run", "Execute EVM bytecode");
+    run_cmd.add_option("code", code_hex, "Hex-encoded bytecode")->required();
+    run_cmd.add_option("--vm", vm_config, "EVMC VM module")->required()->envname("EVMC_VM");
+    run_cmd.add_option("--gas", msg.gas, "Execution gas limit", true)
```
gumb0: Maybe print that 1000000 is default

tools/evmc/main.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+    std::string vm_config;
+    std::string code_hex;
+    evmc_message msg{};
+    msg.gas = 1000000;
+    auto rev = EVMC_ISTANBUL;
+
+    auto& run_cmd = *app.add_subcommand("run", "Execute EVM bytecode");
+    run_cmd.add_option("code", code_hex, "Hex-encoded bytecode")->required();
+    run_cmd.add_option("--vm", vm_config, "EVMC VM module")->required()->envname("EVMC_VM");
```
gumb0: Does it take the value from environment variable by default? Also maybe say this in the help message if it doesn't do it automatically.

chfast: It uses the env var if `--vm` not provided.

tools/evmc/main.cpp
chfast: This is print by automatically generated help message.

gumb0: ok nice

axic:
Why disable codecov reporting?

chfast:
> Why disable codecov reporting?

Only the PR status reporting - the PR will not fail because of the coverage drop, etc. The comment with the status will be posted.

axic:
Forgot to document this in the README & docs.

PR #482, opened 2019-12-06
Coverage

chfast:
This creates LCOV code coverage report using clang AST-based coverage tools. The report is visible at CircleCI as artifact: https://20298-66214407-gh.circle-artifacts.com/0/coverage/index.html
The same report is also send to codecov.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+comment:
+  layout: "diff"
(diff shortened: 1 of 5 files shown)


codecov.yml
```diff
@@ -0,0 +1,4 @@
+comment:
+  layout: "diff"
+ignore:
+  - "test/.*"
```
chfast: This should be removed.

PR #484, opened 2019-12-10
tools: Load bytes from hex string or file

chfast:
I'm just dumping this because I cannot find a way to force CLI11 to work correctly.


axic:
@chfast is this superseded now?

chfast:
Replaced with #574.

PR #486, opened 2019-12-12
go: Add Go modules support

### LINKED ISSUE #450: Add support for Go modules

chfast:
Go modules support by @mandrigin.

Closes #450

### DIFF
--- a/bindings/go/evmc/evmc.go
+++ b/bindings/go/evmc/evmc.go
@@ -5,7 +5,7 @@
 package evmc
 
 /*
-#cgo CFLAGS:  -I${SRCDIR}/.. -Wall -Wextra
+#cgo CFLAGS: -I${SRCDIR}/../../../include -Wall -Wextra
 #cgo !windows LDFLAGS: -ldl
 
 #include <evmc/evmc.h>
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/ethereum/evmc/v7
+
+go 1.11
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,15 @@ and this project adheres to [Semantic Versioning].
 
 ## [7.3.0] — unreleased
 
+### Added
+
+- Support for **Go modules**.
+  [#486](https://github.com/ethereum/evmc/pull/486)
+
 ### Changed
 
+- The minimum **Go version** supported bumped to **1.11** (Go modules are required).
+  [#486](https://github.com/ethereum/evmc/pull/486)
 - Removed dependency on go-ethereum in Go bindings by introducing own `Address` and `Hash` types.
   [#513](https://github.com/ethereum/evmc/pull/513)
 
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Please visit the [documentation].
 | ----------------------------- | --------------------- | ------------------------------ | -------------------
 | **C**                         | C99, C11              | GCC 6+, clang 3.8+, MSVC 2015+ | Host- and VM-side
 | **C++**                       | C++11, C++14, C++17   | GCC 6+, clang 3.8+, MSVC 2015+ | Host- and VM-side
-| **Go** _(bindings)_           | 1.9 - 1.12            |                                | Host-side only
+| **Go** _(bindings)_           | 1.11 - 1.14 (modules) |                                | Host-side only
 | **Rust** _(bindings)_[¹](#n1) | 2018 edition          | 1.37.0 and newer               | VM-side only
 | **Java** _(bindings)_         | 11                    |                                | Host-side only
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -39,11 +39,6 @@ build_script:
       gcc --version
       go env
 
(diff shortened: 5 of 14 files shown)


go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/ethereum/evmc
+
+go 1.13
+
+require github.com/ethereum/go-ethereum v1.9.9
```
axic: Is this needed for testing? Does it mean that anyone using evmc as a god module also will depend on geth?

chfast: I believe so. We are only using `github.com/ethereum/go-ethereum/common` but go-ethereum only defines single module for whole source repo. Maybe there is a way to define the `github.com/ethereum/go-ethereum/common` module there?

chfast: I'm also not sure if go-ethereum will be able to use evmc module with this.

axic: This is solved by #499.

go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/ethereum/evmc/v7
+
+go 1.14
```
axic, line 3: The readme say "Go (bindings) | 1.9 - 1.12". Is this changing that?

mandrigin: you can still build it on the old golang, but go modules will require go 1.14

I can adjust README

mandrigin:
turbo-geth code that uses this PR: https://github.com/ledgerwatch/turbo-geth/tree/evmc-upstream/core

go.mod
mandrigin: updated readme

go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/ethereum/evmc/v7
```
chfast, line 1: So it is enough to only declare it as "v7" here but use dir structure unchanged?

bindings/go/evmc/evmc.go
```diff
@@ -15,6 +15,8 @@ package evmc
 #include <stdlib.h>
 #include <string.h>
 
+#include "loader/loader.c"
```
chfast, line 18: I think this should be a problem because the C comment in go files are compiled twice if any Go function is exported. I think it only works because there is nothing exported here.
I may have modified version of this, will try to push it there.

go.mod
mandrigin: yeah, it seems so

bindings/go/evmc/evmc.go
mandrigin: It misses some definitions w/o that import for some reason... At least on a mac.

chfast:
I pushed some additions, including an integration test for Go module.

bindings/go/evmc/evmc.go
chfast: I changed that.

go.mod
chfast: I'm not sure what this mean, but this works in Go 1.11 (the first version that supports modules).

chfast: I changed that to `1.11` as some random comment on internet this is minimum Go version of the module.

go.mod
axic, line 1: Why is there a need for v7 here? If we bump to ABI 8 and this is changed to v8, will go modules still be able to download v7 things?

Or the only benefit is knowing it is v7 even if someone uses a commithash after it?

chfast: This is required by `go mod`. The `vX` is required if you version is `vX.Y.Z`. Can be only omitted if `X==1 || X==0`.

Versions are marked with git tags, and v7 series will continue to work (because e.g. at `v7.3.0` this line will be still `module github.com/ethereum/evmc/v7`.

The usage is something like this:

```
go get github.com/ethereum/evmc/v7@v7.3.0
go get github.com/ethereum/evmc/v7@v7.3.1
go get github.com/ethereum/evmc/v8@v8.0.0
```

And also 
```go
import "github.com/ethereum/evmc/v7"
import "github.com/ethereum/evmc/v8"
```

So users need to update their import directives when upgrading major version number. In case of EVMC where vX matches ABI version this is reasonable IMHO.

axic: So it is just a convention required by `go mod`, it won't magically make `v7` work when we move on to `v8`, because it still just downloads from the repo master, right?

chfast: 1. `go get github.com/ethereum/evmc/v7@v8.0.0` will fail due to mismatch.
2. `go get github.com/ethereum/evmc/v7@master` will also fail if the most recent version is `v8.Y.Z`. `go mod` uses a kind of `git describe` strategy to figure out the "current" version.

axic: Sounds good.

axic, line 1: Maybe in a future PR, one could update `bumpversion` to also update this line.

axic:
Do we have documentation on the bindings? I don't think we have them yet, but maybe we should, where example uses could be explained.

chfast:
> Do we have documentation on the bindings? I don't think we have them yet, but maybe we should, where example uses could be explained.

We don't have any Go documentation. The standard Go packages documentation hosting is https://pkg.go.dev/. For general documentation migration to readthedocs would be nice.

PR #488, opened 2019-12-12
C++: Add uint64_t -> {bytes32, address} constructors

### LINKED ISSUE #487: Add constructor with uint64_t input for evmc::uint256

chfast:
Closes #487

### DIFF
--- a/include/evmc/evmc.hpp
+++ b/include/evmc/evmc.hpp
@@ -25,6 +25,33 @@ struct address : evmc_address
     /// Initializes bytes to zeros if not other @p init value provided.
     constexpr address(evmc_address init = {}) noexcept : evmc_address{init} {}
 
+    /// Converting constructor from unsigned integer value.
+    ///
+    /// This constructor assigns the @p v value to the last 8 bytes [12:19]
+    /// in big-endian order.
+    constexpr explicit address(uint64_t v) noexcept
+      : evmc_address{{0,
+                      0,
+                      0,
+                      0,
+                      0,
+                      0,
+                      0,
+                      0,
+                      0,
+                      0,
+                      0,
+                      0,
+                      static_cast<uint8_t>(v >> 56),
+                      static_cast<uint8_t>(v >> 48),
+                      static_cast<uint8_t>(v >> 40),
+                      static_cast<uint8_t>(v >> 32),
+                      static_cast<uint8_t>(v >> 24),
+                      static_cast<uint8_t>(v >> 16),
+                      static_cast<uint8_t>(v >> 8),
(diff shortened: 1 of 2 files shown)


include/evmc/evmc.hpp
```diff
@@ -15,6 +15,8 @@
 /// @ingroup cpp
 namespace evmc
 {
+#define B(V, INDEX) static_cast<uint8_t>((V) >> (8 * (INDEX)))
```
gumb0: why not a constexpr function?

chfast: The name must be short to get formatting reasonable. And I can undef it later.

axic:
I'm not sure we need it for `bytes32` and `address`.

include/evmc/evmc.hpp
axic: I'm not a fan of macros with really short names. Also to me the original long formatted version with `static_cast`s doesn't look that bad to warrant a bad macro :)

gumb0: Well if someone has another `#define B` before `#include <evmc/evmc.hpp>`, this will break it.

You could maybe group lines by 4-5 bytes to make formatting work with longer name.


Also lint complains about missing documentation for it.

chfast:
The last commit adds macro `B()`. You can see how it looked before.

The constructor for `evmc::address` will be useful in solidity: https://github.com/ethereum/solidity/blob/develop/test/EVMHost.cpp#L124-L139

axic:
> The constructor for evmc::address will be useful in solidity: https://github.com/ethereum/solidity/blob/develop/test/EVMHost.cpp#L124-L139

I don't think it will be used though, this is one is okay for Solidity.

axic:
> The last commit adds macro B(). You can see how it looked before.

Yeah the one before the macro looked fine to me.

test/unittests/cpp_test.cpp
```diff
@@ -330,6 +330,45 @@ TEST(cpp, literals)
     EXPECT_EQ(h1, f1);
 }
 
+TEST(cpp, bytes32_from_uint)
+{
+    using evmc::bytes32;
+    using evmc::operator""_bytes32;
+
+    static_assert(bytes32{0} == bytes32{}, "");
```
gumb0: VS 2015 fails at this line

gumb0:
Ugly formatting is better than a macro for me, too.

chfast:
Ok, removed commits with macro and address type.

axic:
The address type is fine, I don't mind it, just that probably Solidity will not use it.

chfast:
Added address back and fixed VS 2015.

PR #489, opened 2019-12-12
java: fix crash with evmc_last_error_msg

axic:


bindings/java/c/evmc-vm.c
```diff
@@ -24,7 +24,7 @@ JNIEXPORT jobject JNICALL Java_org_ethereum_evmc_EvmcVm_init(JNIEnv* jenv,
         {
             const char* error_msg = evmc_last_error_msg();
             jclass jclazz = (*jenv)->FindClass(jenv, "java/lang/AssertionError");
-            (*jenv)->ThrowNew(jenv, jclazz, error_msg);
+            (*jenv)->ThrowNew(jenv, jclazz, error_msg ? error_msg : "Loading EVMC VM failed");
```
axic: `ThrowNew` expects a valid pointer and `evmc_last_error_msg` can return null.

axic:
Have bunch of other refactorings too in the pipeline.

bindings/java/c/evmc-vm.c
chfast: Is there really a crash happening?

axic: https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html

The documentation is not explicit about it, but says:
> message: the message used to construct the java.lang.Throwable object. The string is encoded in modified UTF-8.

PR #490, opened 2019-12-12
Java bindings refactoring

axic:


bindings/java/c/host.c
```diff
@@ -18,11 +18,20 @@ static JNIEnv* attach()
     return jenv;
 }
 
+static jbyteArray copyDataToJava(JNIEnv* jenv, const void* ptr, size_t size)
```
chfast: Seems to be inconsistent with other function names. I not sure we should aim for consistent naming convention here.

bindings/java/c/host.c
```diff
@@ -66,6 +68,7 @@ static evmc_bytes32 get_storage_fn(struct evmc_host_context* context,
     evmc_bytes32 result;
     const char java_method_name[] = "get_storage";
     const char java_method_signature[] = "(I[B[B)Ljava/nio/ByteBuffer;";
+    assert(context != NULL);
```
chfast: There are makefile targets: `build` and `debug`. I'm not sure this is intentional, but asserts are enabled in both. I think this is good at least at this phase of development.

bindings/java/c/host.c
axic: It is somewhat consistent with the JNI helpers though. Perhaps follow those?

bindings/java/c/host.c
axic: I think it is intentional currently.

atoulme:
Sorry, I can't really review C code, not my specialty. I would love to try this code out at some point.

bindings/java/c/evmc-vm.c
```diff
@@ -7,6 +7,12 @@
 #include "evmc/loader.h"
 #include "host.h"
 
+static void throw_java_assert(JNIEnv* jenv, const char* msg)
+{
+    jclass jclazz = (*jenv)->FindClass(jenv, "java/lang/AssertionError");
```
chfast, line 12:
```suggestion
    jclass jcls = (*jenv)->FindClass(jenv, "java/lang/AssertionError");
```

Generally using variable names as `jxxx` causes troubles because they clash if JNI defined type names. E.g. every usage of `jvalue` clashes with `jvalue` type.

Maybe name variables `j_class` or `jClass`. But I'm also fine with having file-scope consistent mapping: `jclass -> jcls`.

bindings/java/c/evmc-vm.c
```diff
@@ -15,23 +21,16 @@ JNIEXPORT jobject JNICALL Java_org_ethereum_evmc_EvmcVm_init(JNIEnv* jenv,
...
-        (*jenv)->ReleaseStringUTFChars(jenv, jfilename, filename);
-    }
-    else
+    const char* filename = (*jenv)->GetStringUTFChars(jenv, jfilename, NULL);
+    if (filename == NULL)
+        throw_java_assert(jenv, "JNI Error: filename cannot be NULL");
+    enum evmc_loader_error_code loader_error;
+    evm = evmc_load_and_create(filename, &loader_error);
+    (*jenv)->ReleaseStringUTFChars(jenv, jfilename, filename);
+    if (evm == NULL || loader_error != EVMC_LOADER_SUCCESS)
```
chfast, line 30:
```suggestion
    if (loader_error != EVMC_LOADER_SUCCESS)
```
Don't check `evm` for null, checking error code is enough.

bindings/java/c/host.c
chfast: It will not be when we move it under CMake jurisdiction.

bindings/java/c/evmc-vm.c
axic: This was in the original version of the code.

chfast: I can see that.

bindings/java/c/host.c
axic: What do you suggest to use?

bindings/java/c/evmc-vm.c
axic: Let me know a naming scheme and we can apply it to the file.

bindings/java/c/host.c
axic: Looking at the extent of changes I have a preference to merge this first and then the cmake PR also replacing the asserts.

chfast: `assert` is fine.

bindings/java/c/evmc-vm.c
chfast: `jcls`.

axic: Was hard to see on the phone :)

bindings/java/c/evmc-vm.c
axic: Fixed.

axic:
@chfast addressed, squashed, should be ready?

PR #491, opened 2019-12-13
cpp: Replace literals implementation with constexpr functions

chfast:


include/evmc/evmc.hpp
```diff
@@ -255,108 +255,72 @@ namespace literals
...
-template <typename T, char... Literal>
-constexpr T parse() noexcept
+template <typename T>
+constexpr T from_literal(const char* s)
 {
-    return parse_literal<T, char_sequence<Literal...>>::get();
+    return (s[0] == '0' && s[1] == '\0') ?
+               T{} :
+               !(s[0] == '0' && s[1] == 'x') ?
+               throw "literal must be in hexadecimal notation" :
```
gumb0: How does throw in constexpr work? This message will be in compiler output?

chfast:
This is only to simplify the implementation, no functionality has been altered. This version is also more compiler-friendly. In previous version, the number of template instances was proportional to number of characters in all literals used. This version is also easier to understand and modify. 

The `throw` in constexpr do not have special meaning - it works as any expression that is _not_ evaluable in constexpr context : compiler will protest if this branch is reached during compile-time evaluation. But it will also highlight the expression in the error message so it is possible to figure out what went wrong. Example:

```
In file included from /home/chfast/Projects/ethereum/evmc/test/unittests/cpp_test.cpp:12:
/home/chfast/Projects/ethereum/evmc/include/evmc/evmc.hpp: In member function ‘virtual void cpp_literals_Test::TestBody()’:
/home/chfast/Projects/ethereum/evmc/test/unittests/cpp_test.cpp:303:9:   in ‘constexpr’ expansion of ‘evmc::literals::operator""_bytes32(((const char*)"0x01020304050607080910a1a2a3a4a5a6a7a8a9b0c1c2c3c4c5c6c7c8c9d0d1d20"))’
/home/chfast/Projects/ethereum/evmc/include/evmc/evmc.hpp:318:45:   in ‘constexpr’ expansion of ‘evmc::literals::internal::from_literal<evmc::bytes32>(s)’
/home/chfast/Projects/ethereum/evmc/include/evmc/evmc.hpp:304:57: error: expression ‘<throw-expression>’ is not a constant expression
  304 |                (length(s + 2) != sizeof(T) * 2) ? throw "literal must match the result type size" :
      |                                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
A short article about it: http://ericniebler.com/2014/09/27/assert-and-constexpr-in-cxx11/

I thought this trick is only possible in C++14, that's why I used templates and static_assert() previously. It turned out I was wrong - that's why I'm proposing new implementation. However, using static_assert() is nicer for reporting errors to users.

The design for these literals requires to specify "all" bytes explicitly, including leading zero ones. So you cannot have:

```
0xa1a2a3a4a5a6a7a8a9d0d1d2d3d4d5d6d7d8d9_address
```

Did you notice that the first byte of the address is `00`?

One exception is zero value as `0_address`.

I'm also considering to extend the allowed formats for "short" literals (let's say up to 4 bytes). E.g.:

```
0x9_address
0xffff_address
```

This refactoring makes it much easier to extend by such features.

Or maybe checksum addresses? ;)

gumb0:
Awesome! I wish you'd write the explanations for PRs more often :)

chfast:
I will try to improve here. I also noticed I usually explain myself very too late.

PR #492, opened 2019-12-13
cpp: Add VM::get_raw_pointer()

chfast:


include/evmc/evmc.hpp
```diff
@@ -727,6 +727,13 @@ class VM
             m_instance->execute(m_instance, nullptr, nullptr, rev, &msg, code, code_size)};
     }
 
+    /// Returns the pointer to C EVMC struct representing the VM.
+    ///
+    /// Gives access to the C EVMC VM struct to allow advanced interaction with the VM not supported
+    /// by the C++ interface. Use as the last resort.
+    /// The returned pointer MAY be null and this object is still owning it.
```
gumb0: If I understand it right, it should better be two separate sentences, something like "This object still owns the VM after returning the pointer. The returned pointer MAY be null. "

chfast: Yes, I will change it.

Issue #471, opened 2019-11-28
Bump CMake requirement to 3.13


chfast:

axic:
Why is this needed? To clean up the makefiles a bit?

chfast:
This allows some simplifications in CMake scripts. E.g. you can `install()` targets from a single location in the main CMakeLists.txt file.

a1love:
Bump CMake requirement to 3.13 #472

PR #495, opened 2019-12-13
cmake: Bump minimum required version to 3.10

chfast:
- We are using GoogleTest module introduced in 3.9.
- None of Debian / Ubuntu distributions versions provides CMake 3.9. The 3.10 is selected because this is the version provided in Ubuntu 18.04 LTS.


axic:
Is this closing #471?

chfast:
> Is this closing #471?

No. It simply documents the current status (CMake 3.9 is required currently if you want to build everything). This might be good enough for EVMC 7.2.

PR #497, opened 2019-12-23
Fix documentation comment for evmc_result::gas_left

chfast:


include/evmc/evmc.h
```diff
@@ -345,8 +345,8 @@ struct evmc_result
     /**
      * The amount of gas left after the execution.
      *
-     *  If evmc_result::code is not ::EVMC_SUCCESS nor ::EVMC_REVERT
-     *  the value MUST be 0.
+     * If evmc_result::status_code is not ::EVMC_SUCCESS nor ::EVMC_REVERT
```
gumb0:
```suggestion
     * If evmc_result::status_code is neither ::EVMC_SUCCESS nor ::EVMC_REVERT
```

gumb0: Why do you write status codes with double colon?

chfast: Doxygen will make them links.

chfast: https://evmc.ethereum.org/structevmc__result.html#af8478c93dbcc3cb2876037c5a5afd4c0

PR #498, opened 2020-01-06
fix less compare for address and bytes32

kjx98:
"load64be(&a.bytes[8]) == load64be(&b.bytes[8])" requires "load64be(&a.bytes[0]) == load64be(&b.bytes[0]" as well, check if load64be(&a.bytes[0]) > load64be(&b.bytes[0] and "load64be(&a.bytes[8]) == load64be(&b.bytes[8])"


chfast:
Do you have unit tests for these cases?

kjx98:
add unit test to test/unittests/cpp_test.cpp
z_max stands 0x00000000 00000000 (64 bits) leading 0 for address/bytes32
max_z tailing with 64 bits 0

PR #501, opened 2020-05-06
java: Build update

chfast:
This is Java build system maintenance work and to fix my local build based on OpenJDK 13. I believe Gradle 6 is needed for OpenJDK 13.

There are some additional questions:
- The `build/evmc.jar` (produced by gradle?) is empty.
- I can build `evmc.so` with JNI using CMake. That would be preferable than using the custom Makefile.
- I can also build `evmc.jar` with CMake. So we may only need gradle to do testing (get JUnit5) and "spotless"?

### DIFF
--- a/bindings/java/wrapper.gradle
+++ b/bindings/java/wrapper.gradle
@@ -1,4 +1,4 @@
 task setup(type: Wrapper) {
     description = "Download the gradle wrapper and requisite files. Overwrites existing wrapper files."
-    gradleVersion = "5.0"
+    gradleVersion = "6.3"
 }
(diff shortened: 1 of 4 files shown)


bindings/java/wrapper.gradle
```diff
@@ -1,4 +1,4 @@
 task setup(type: Wrapper) {
     description = "Download the gradle wrapper and requisite files. Overwrites existing wrapper files."
-    gradleVersion = "5.0"
+    gradleVersion = "6.3"
```
axic, line 3: Do we need to require newer gradle? Does it mean projects using gradle 5.x can't use the binding?

chfast: I don't think gradle is dependecy on its own. And Gradle 6 is needed for Java 13 support. Gradle 6.3 for Java 14 support.

jrhea: Shouldn't be a problem. Projects using an older version of Gradle can still use the binding.  They can just build this separately and then reference the jar in their build.

PR #502, opened 2020-05-06
readme: mention evmc-js, rust-ssvm, ssvm, core-geth and turbo-geth

axic:


axic:
Should we also mention [aleth-precompiles](https://github.com/ethereum/aleth/pull/5855) ?

gumb0:
> Should we also mention aleth-precompiles ?

probably no, until/unless it's merged

PR #503, opened 2020-05-06
Java CMake build

chfast:
This adds:
- `evmc-java.so` shared library with JNI (optional with `-DEVMC_JAVA=ON`),
- `evmc-java.jar` built by CMake.

But for testing Java we still use gradle + Makefile, so `evmc-java.jar` is not directly tested.


axic:
Would this deprecate the `bindings/java/Makefile` file?

axic:
Do you want to split out the first commit (C fixes) and adding the license header into a new PR and leave this PR for cmake only?

axic:
Probably should remove `bindings/java/java/README.md` as it contains outdated info (old gradle changed by #501, `make debug` removed by #507)

axic:
> Would this deprecate the bindings/java/Makefile file?

If so, can it be removed?

In case this question was hidden.

chfast:
> > Would this deprecate the bindings/java/Makefile file?
> 
> If so, can it be removed?
> 
> In case this question was hidden.

This is covered now.

bindings/java/Makefile
```diff
@@ -1,42 +1,31 @@
...
 	./gradlew --no-daemon test
 
+gradlew:
+	gradle --no-daemon setup
+
+$(OUT_DIR)/lib/libevmc-java.so:
+	cmake -B $(OUT_DIR)/_cmake_build -S $(SOURCE_DIR) -DCMAKE_INSTALL_PREFIX=$(OUT_DIR) -DEVMC_JAVA=ON -DJAVA_HOME=$(JAVA_HOME) -DEVMC_EXAMPLES=ON
+	cmake --build $(OUT_DIR)/_cmake_build --target install
+
 format:
```
axic, line 25: I think this is not important.

chfast:
> Looks okay to me, but still a bit confused why we need the makefile which invokes the cmakefile.

So Java dev know how to actually build it.

PR #504, opened 2020-05-06
Upgrade Cable to 0.5.0

chfast:


cmake/cable/CableCompilerSettings.cmake
```diff
@@ -109,9 +137,6 @@ macro(cable_configure_compiler)
             else()
                 add_compile_options(-mtune=native -march=native)
             endif()
-        elseif(NOT MSVC)
-            # Tune for currently most common CPUs.
-            cable_add_cxx_compiler_flag_if_supported(-mtune=generic)
```
chfast, line 114: Pulled Cable update to have this change.

axic:
Does cable 0.5.0 contains those commits found in fizzy?

chfast:
> @chfast can you also update cable in fizzy? (with a subtree?)

There is not much to update there: https://github.com/wasmx/fizzy/pull/303.

PR #512, opened 2020-05-12
java: prefix C exports with evmc_java_

axic:
Similarly to how we prefix with evmc_go_ in the go bindings.

See https://github.com/ethereum/evmc/blob/master/bindings/go/evmc/host.c#L16.

Probably it doesn't matter too much as long as the `bindings/java/c` directory is a compiled to a single shared library, which I think it is.


chfast:
They should not be exported anyway.

PR #499, opened 2020-05-04
golang: remove dependencies on go-ethereum

mandrigin:
This PR introduces `Hash` and `Address` types for Golang binding and removed dependencies on `go-ethereum/common` package.

PR #513, opened 2020-05-14
go: Remove dependencies on go-ethereum

chfast:
Continuation of #499.

I added a change not to define `{address,hash}Length` constants. To me this looks fine.
Because of the change I would just move the `Address` and `Hash` types to `evmc.go`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog],
 and this project adheres to [Semantic Versioning].
 
 
+## [7.3.0] — unreleased
+
+### Changed
+
+- Removed dependency on go-ethereum in Go bindings by introducing own `Address` and `Hash` types.
+  [#513](https://github.com/ethereum/evmc/pull/513)
+
+
 ## [7.2.0] — 2020-05-13
 
 ### Added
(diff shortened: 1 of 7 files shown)


bindings/go/evmc/types.go
```diff
@@ -0,0 +1,11 @@
+// EVMC: Ethereum Client-VM Connector API.
+// Copyright 2020 The EVMC Authors.
+// Licensed under the Apache License, Version 2.0.
+
+package evmc
+
+// Hash represents the 32 bytes of arbitrary data (e.g. the result of Keccak256 hash).
```
axic, line 7: Is byteorder important here?

axic:
@mandrigin @chfast Should `Error` (https://github.com/ethereum/evmc/blob/remove-geth-types/bindings/go/evmc/evmc.go#L71) be also moved to types?

chfast:
> Should `Error` (https://github.com/ethereum/evmc/blob/remove-geth-types/bindings/go/evmc/evmc.go#L71) be also moved to types?



> Because of the change I would just move the `Address` and `Hash` types to `evmc.go`.

axic:
Argh just noticed, couldn't these things be removed from CI: https://github.com/ethereum/evmc/pull/486/files#diff-180360612c6b8c4ed830919bbb4dd459L41

chfast:
What do you mean?

axic:
Never mind, this PR did remove those lines.

PR #519, opened 2020-06-23
cpp: Use custom error("msg") to report invalid literals

chfast:


axic:
Can you show what the compiler error message would be like?

chfast:
```
error: call to non-‘constexpr’ function ‘T evmc::literals::internal::error(const char*) [with T = evmc::address]’
  309 |                !(s[0] == '0' && s[1] == 'x') ? error<T>("literal must be in hexadecimal notation") :
      |                                                ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

include/evmc/evmc.hpp
```diff
@@ -294,15 +294,22 @@ constexpr address from_hex<address>(const char* s) noexcept
...
+template <typename T>
+constexpr T from_literal(const char* s) noexcept
 {
     return (s[0] == '0' && s[1] == '\0') ?
                T{} :
-               !(s[0] == '0' && s[1] == 'x') ?
-               throw "literal must be in hexadecimal notation" :
-               (length(s + 2) != sizeof(T) * 2) ? throw "literal must match the result type size" :
-                                                  from_hex<T>(s + 2);
+               !(s[0] == '0' && s[1] == 'x') ? error<T>("literal must be in hexadecimal notation") :
```
axic, line 309: Can somehow this be done via "static_assert" ?

chfast:
I had to add `std::abort` in the implementation because it looks clang in Debug want's to parse literals in runtime -> we get linker errors. Empty implementation will also do. Not sure what is better.

include/evmc/evmc.hpp
chfast: It might be possible in C++11 but after long night hours I still don't have it fully right. It should be much simpler in C++14 where you can just use variables in `constexpr` functions.

chfast:
Closed in favor of #520.

PR #520, opened 2020-06-24
Rework literals implementation to use static_assert

chfast:
Closes #519.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning].
 
 - The C++ standard required increased to **C++14**.
   [#521](https://github.com/ethereum/evmc/pull/521)
+- The C++ literals for `address` and `bytes32` types have been reimplemented to
+  use `static_assert` to report errors. This makes C++ bindings exception-free
+  (_no-throw guarantee_ level of exception safety).
+  [#520](https://github.com/ethereum/evmc/pull/520)
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning].
 
 - The C++ standard required increased to **C++14**.
   [#521](https://github.com/ethereum/evmc/pull/521)
+- The C++ literals for `address` and `bytes32` types have been reimplemented
+  to use `static_assert` to report errors (instead of exceptions).
```
axic, line 15: Maybe highlight this makes the C++ bindings exception free?

test/unittests/cpp_test.cpp
```diff
@@ -111,10 +111,8 @@ TEST(cpp, std_hash)
 #pragma warning(disable : 4307 /* integral constant overflow */)
 #pragma warning(disable : 4309 /* 'static_cast': truncation of constant value */)
 
-#if !defined(_MSC_VER) || (_MSC_VER >= 1910 /* Only for Visual Studio 2017+ */)
     static_assert(std::hash<evmc::address>{}({}) == static_cast<size_t>(0xd94d12186c0f2fb7), "");
     static_assert(std::hash<evmc::bytes32>{}({}) == static_cast<size_t>(0x4d25767f9dce13f5), "");
```
axic, line 115: Why have both `static_assert` and `EXPECT_EQ` for the same cases here?

chfast: The `constexpr` functions only mean "may be executed in constexpr context". But they can also be executed with non-constexpr arguments at runtime. Additional runtime executions also contribute to code coverage.

PR #521, opened 2020-06-24
cpp: Require C++14

chfast:
Bump C++ standard required from C++11 to C++14. This also drops support
for Visual Studio 2015.


axic:
Checked all clients/VMs listed in the README and it seems all of them are at least C++14 **if** they use the C++ headers.

Issue #476, opened 2019-12-03
Implement Rust evmc-host crate to consume VMs


axic:

axic:
An implementation exists here: https://github.com/second-state/rust-ssvm

CaptainVincent:
> An implementation exists here: https://github.com/second-state/rust-ssvm

Below things are being processed and we will update this issue as we make progress.
1. We will try to keep EVMC separate from rust-ssvm make them more independent at first.
2. Upgrade our codebase EVMC version from 6.x to 7.x.

CaptainVincent:
Stage 1 is completed. We separate evmc-client with evmc ABI version 6 from rust-ssvm.
> fork official repository with tag [v6.3.1-rust-evmc-client](https://github.com/second-state/evmc/releases/tag/v6.3.1-rust-evmc-client)

Now we will migration evmc-client to ABI version 7 and upstream it.

axic:
Sounds good, thanks!

CaptainVincent:
Progress:
We already migrate our evmc-client in rust to evmc v7.3.0.
https://github.com/second-state/evmc/tree/rust-binding-client

I will try more local test, add some unit test cases and update licence meet official requirement in next stage before PR it.

PR #524, opened 2020-06-24
rust: turn evmc_host_context into a void type

axic:
This may be needed for #476.

### DIFF
--- a/bindings/rust/evmc-sys/src/lib.rs
+++ b/bindings/rust/evmc-sys/src/lib.rs
@@ -8,6 +8,11 @@
 
 include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 
+// Defining evmc_host_context here, because bindgen cannot create a useful declaration yet.
+
+/// This is a void type given host context is an opaque pointer. Functions allow it to be a null ptr.
+pub type evmc_host_context = ::std::os::raw::c_void;
+
 // TODO: add `.derive_default(true)` to bindgen instead?
 
 impl Default for evmc_address {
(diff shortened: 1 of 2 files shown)


bindings/rust/evmc-sys/src/lib.rs
```diff
@@ -6,10 +6,15 @@
 #![allow(non_upper_case_globals)]
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
-#![allow(clippy::derive_hash_xor_eq)]
 
 include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 
+// Defining evmc_host_context here, because bindgen cannot create a useful declaration yet.
+
+pub type evmc_host_context = ::std::os::raw::c_void;
```
axic, line 14: Could also just implement `Default` on this.

bindings/rust/evmc-sys/src/lib.rs
```diff
@@ -6,10 +6,15 @@
...
+
+//pub struct evmc_host_context { }
```
axic, line 16: Could also just implement Default on this.

bindings/rust/evmc-sys/src/lib.rs
axic, line 14: Perhaps this is the best option.

CaptainVincent:
> This may be needed for #476.

Yes, #476 need evmc_host_context instantiable. I will replace my workaround after this PR merged. Thank you for informing me.

axic:
@CaptainVincent do you want to make a PR so we can discuss it?

CaptainVincent:
> @CaptainVincent do you want to make a PR so we can discuss it?

Done. #527.

axic:
@vorot93 this is an old outstanding change, but do you have any suggestion?

What bindgen generates is the following:
```rust
#[repr(C)]
#[derive(Debug, Copy, Clone, Hash, PartialEq)]
pub struct evmc_host_context {
    _unused: [u8; 0],
}
```

We want something what can be "instantiated".

chfast (approved):

PR #526, opened 2020-06-24
rust: upgrade bindgen to 0.54

axic:


bindings/rust/evmc-sys/build.rs
```diff
@@ -26,6 +26,8 @@ fn gen_bindings() {
         .whitelist_type("evmc_.*")
         .whitelist_function("evmc_.*")
         .whitelist_var("EVMC_ABI_VERSION")
+        // TODO: consider removing this
+        .size_t_is_usize(true)
```
axic, line 30: This was the default <0.52.0 and our bindings rely on this. It was disabled by default in 0.53.0 but this option exists in 0.54.0.

We have to consider changing the bindings, but this update here with this setting will produce identical output so no risk.

PR #527, opened 2020-06-26
Add rust binding client

CaptainVincent:
1. Add rust client binding base on EVMC 7.3.0.
2. f8466c7 is a temporary workaround for evmc_host_context cannot be instantiated.
3. Update all files' license to Apache 2.0.

I am not yet add unit test for this module but it work fine with Hera.
https://github.com/second-state/rust-ssvm/tree/upgrade-evmc-client
Use this branch follow original [example](https://github.com/second-state/rust-ssvm/tree/upgrade-evmc-client#example) section.  
(I just replace submodule and build.rs to build and link libhera.so)


CaptainVincent:
I saw some ci issues come from my evmc_host_context workaround as below.

1. /home/circleci/project/include/evmc/evmc.h:164:8: error: struct has no members [-Werror=pedantic]
2. error: empty struct is a GNU extension [-Werror,-Wgnu-empty-struct]
> struct evmc_host_context {};
3. Code format
```diff
-struct evmc_host_context {};
+struct evmc_host_context
+{
+};
```
4. bindings-go-latest fail reason unknown.

bindings/rust/evmc-client/src/types.rs
```diff
@@ -0,0 +1,18 @@
...
+pub const BYTES32_LENGTH: usize = 32;
+pub type Address = [u8; ADDRESS_LENGTH];
+pub type Bytes32 = [u8; BYTES32_LENGTH];
+pub type Bytes = [u8];
+
+pub type StorageStatus = ffi::evmc_storage_status;
+pub type CallKind = ffi::evmc_call_kind;
+pub type Revision = ffi::evmc_revision;
+pub type StatusCode = ffi::evmc_status_code;
+pub type Capabilities = ffi::evmc_capabilities;
```
axic, line 18: Why not use `evmc-vm::types` instead of duplicating? We could later refactor to pull them out somewhere.

bindings/rust/evmc-client/src/host.rs
```diff
@@ -0,0 +1,315 @@
...
+        depth: i32,
+        is_static: bool,
+    ) -> (Vec<u8>, i64, Address, StatusCode);
+}
+
+struct Callback {
+    host_interface: ::std::option::Option<Box<dyn HostInterface>>,
+}
+
+static mut CALLBACK: Callback = Callback {
```
axic, line 45: This makes this not thread-safe. Why did you choose to use this and not use `evmc_host_context`, which by design is there to avoid having these globals.

bindings/rust/evmc-client/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+        is_static: bool,
+        depth: i32,
+        gas: i64,
+        destination: &Address,
+        sender: &Address,
+        input: &Bytes,
+        value: &Bytes32,
+        code: &Bytes,
+        create2_salt: &Bytes32,
+    ) -> (&Bytes, i64, StatusCode) {
```
axic, line 86: It would be nice to use `evmc-vm::ExecutionResult` here.

bindings/rust/evmc-client/src/loader.rs
```diff
@@ -0,0 +1,74 @@
...
+    EvmcLoaderAbiVersionMismatch = 5,
+
+    /** The VM option is invalid. */
+    EvmcLoaderInvalidOptionName = 6,
+
+    /** The VM option value is invalid. */
+    EvmcLoaderInvalidOptionValue = 7,
+}
+
+pub fn evmc_load_and_create(fname: &str) -> (*mut ffi::evmc_vm, EvmcLoaderErrorCode) {
```
axic, line 40: Why did you chose to reimplement this in Rust and not just use the C loader which already exists?

bindings/rust/evmc-client/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+pub mod types;
+pub use self::loader::EvmcLoaderErrorCode;
+
+use crate::loader::evmc_load_and_create;
+use crate::types::*;
+use evmc_sys as ffi;
+use std::ffi::CStr;
+
+extern "C" {
+    fn evmc_create() -> *mut ffi::evmc_vm;
```
axic, line 17: I think as a first step it would be better to only support dynamic loading.

Are you using this static linking in SSVM?

bindings/rust/evmc-client/src/types.rs
CaptainVincent: Yes, there is needless. I don't notice main line is already added alias types after I fork it. 
Fixed in 7dbcfe0

bindings/rust/evmc-client/src/host.rs
CaptainVincent: Yes, you are right. The original implementation was limited by how much I familiar with Rust. I have rewritten it more like go-binding implementation in 5e150ea.

bindings/rust/evmc-client/src/lib.rs
CaptainVincent: The struct is incompatible with my idea. I try to hide ffi layer's nest struct to user make it more user friendly.
ex. in bindings.rs
```rust
pub struct evmc_address {
    #[doc = " The 20 bytes of the hash."]
    pub bytes: [u8; 20usize],
}
```
evmc-vm expose those nest structures directly like Address, Bytes32 ...etc and ExecutionResult include one Address field.
But I usually prefer use the simply types that re-defined in emc-client as below.
```rust
pub const ADDRESS_LENGTH: usize = 20;
pub const BYTES32_LENGTH: usize = 32;
pub type Address = [u8; ADDRESS_LENGTH];
pub type Bytes32 = [u8; BYTES32_LENGTH];
```
I need confirm which way is better.

bindings/rust/evmc-client/src/loader.rs
CaptainVincent: I reimplement it is because I use this function (loader) outside evmc module at first. I want to avoid move partial evmc's C code to my codebase. But I agree it's better and more consistent to use the C loader after I move my evmc relative code into evmc-client now. I will try to replace it use the C loader, but maybe this won't finish too soon.

bindings/rust/evmc-client/src/lib.rs
CaptainVincent: I am not sure this is relative with static linking. I still use dynamic link to lib<EvmcVM>.so and the statement just tell rustc how to resolve symbol after cargo link the library.
https://github.com/second-state/rust-ssvm/blob/master/build.rs#L25
This is useful in my case that I don't need to know where is my evmc dynamic-link library (especially the target path is not always the same).

bindings/rust/evmc-client/src/loader.rs
CaptainVincent: Finish 5b2b4f2, btw I need cmake in ci environment.

bindings/rust/evmc-client/src/lib.rs
axic: Sorry for the delay, but `evmc-vm::ExecutionResult` is not ffi, it has high-level types.

axic: Also regarding the type aliases, the goal was to eventually replace them with nicer types, but to do it incrementally.

bindings/rust/evmc-client/src/host.rs
```diff
@@ -0,0 +1,237 @@
...
+    ));
+}
+
+pub unsafe extern "C" fn call(
+    context: *mut ffi::evmc_host_context,
+    msg: *const ffi::evmc_message,
+) -> ffi::evmc_result {
+    let msg = *msg;
+    let (output, gas_left, create_address, status_code) =
+        (*(context as *mut ExtendedContext)).hctx.call(
```
axic, line 212: This entire function could be reduced to `hctx.call().into()` is using `ExecutionResult` from `evmc-vm/lib.rs` because that implements the translation and memory handling already.

bindings/rust/evmc-client/src/lib.rs
axic: If it does an extern import of `evmc_create` that is statically linking against an EVMC VM, IIRC, otherwise this would be a lingering reference.

PR #529, opened 2020-07-09
ci: Skip gomod integration test in "PRs from forks"

chfast:
This disables running gomod integration test in PRs from forked repos. It was expected this test not to work as it would require to pull the commit from the fork as the version of `github.com/ethereum/evmc` module.

The same commit from withing `ethereum/evmc` repo: https://circleci.com/gh/ethereum/evmc/24915.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -268,6 +268,8 @@ jobs:
           name: "Go module integration test"
           working_directory: test/gomod
           command: |
+            test -z "$CIRCLE_PR_USERNAME" || (echo "Skip for PRs from forks" && exit 0)
+
             V=$CIRCLE_TAG
             if [ -z $V ]; then
               V=$CIRCLE_SHA1


chfast:
> Would be nice to actually make this working, otherwise we need to double check after a fork is merged.

I tried some tricks, but this is rather impossible. This line https://github.com/ethereum/evmc/blob/master/go.mod#L1 says this is `github.com/ethereum/evmc` module and it cannot be installed from `github.com/<some_fork>/evmc`.

circle.yml
```diff
@@ -268,6 +268,8 @@ jobs:
           name: "Go module integration test"
           working_directory: test/gomod
           command: |
+            test -z "$CIRCLE_PR_USERNAME" || (echo "Skip for PRs from forks" && exit 0)
```
axic, line 271: Hmm, isn't `exit 0` exiting the `()` subshell there?

PR #530, opened 2020-07-23
Bundle shared libs in jar, so it becomes easier to install

atoulme:
This PR allows to search for libevmc in paths, and will install locally a libevmc as a temporary file if needed if the .so file is not found in the path.
This PR also resolves the test example-vm.so file to the Java tests so they don't depend on the user dir.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,3 @@
 /.idea
 Cargo.lock
 /target
-bindings/java/java/src/main/resources
-bindings/java/java/src/test/resources
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 /.idea
 Cargo.lock
 /target
+bindings/java/java/src/main/resources
+bindings/java/java/src/test/resources
--- a/bindings/java/c/host.c
+++ b/bindings/java/c/host.c
@@ -41,28 +41,26 @@ static bool account_exists_fn(struct evmc_host_context* context, const evmc_addr
     assert(context != NULL);
     JNIEnv* jenv = attach();
 
-        jclass host_class;
-        jmethodID method;
-        jint jcontext_index;
-        jbyteArray jaddress;
+    jclass host_class;
+    jmethodID method;
+    jint jcontext_index;
+    jbyteArray jaddress;
 
-        // get java class
-        host_class = (*jenv)->FindClass(jenv, "org/ethereum/evmc/Host");
-        assert(host_class != NULL);
+    // get java class
+    host_class = (*jenv)->FindClass(jenv, "org/ethereum/evmc/Host");
+    assert(host_class != NULL);
 
-        // get java method
-        method =
-            (*jenv)->GetStaticMethodID(jenv, host_class, java_method_name, java_method_signature);
-        assert(method != NULL);
+    // get java method
+    method = (*jenv)->GetStaticMethodID(jenv, host_class, java_method_name, java_method_signature);
+    assert(method != NULL);
 
-        // set java method params
-        jcontext_index = context->index;
-        jaddress = CopyDataToJava(jenv, address, sizeof(struct evmc_address));
+    // set java method params
+    jcontext_index = context->index;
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -3,3 +3,5 @@
 /.idea
 Cargo.lock
 /target
+bindings/java/java/src/main/resources
+bindings/java/java/src/test/resources
```
axic, line 7: These should go into `bindings/java/.gitignore`.

bindings/java/c/host.c
```diff
@@ -40,31 +40,28 @@ static bool account_exists_fn(struct evmc_host_context* context, const evmc_addr
     const char java_method_signature[] = "(I[B)I";
     assert(context != NULL);
     JNIEnv* jenv = attach();
-    if (jenv != NULL)
```
axic, line 43: Why did you remove these?

atoulme: cmake fails to compile with this. Also, the behavior is surprising - if JNI fails to attach, we should fail, not avoid executing the operation.

.gitignore
atoulme: done

atoulme:
Closing this in favor of #531

bindings/java/c/host.c
axic: Can you show an error log?

atoulme:
```/Users/antoinetoulme/w/evmc/bindings/java/c/host.c:370:9: error: variable 'result' is used uninitialized whenever 'if' condition is false [-Werror,-Wsometimes-uninitialized]
    if (jenv != NULL)
        ^~~~~~~~~~~~
/Users/antoinetoulme/w/evmc/bindings/java/c/host.c:402:12: note: uninitialized use occurs here
    return result;
           ^~~~~~
/Users/antoinetoulme/w/evmc/bindings/java/c/host.c:370:5: note: remove the 'if' if its condition is always true
    if (jenv != NULL)
    ^~~~~~~~~~~~~~~~~
/Users/antoinetoulme/w/evmc/bindings/java/c/host.c:365:5: note: variable 'result' is declared here
    struct evmc_result result;
    ^
/Users/antoinetoulme/w/evmc/bindings/java/c/host.c:412:9: error: variable 'result' is used uninitialized whenever 'if' condition is false [-Werror,-Wsometimes-uninitialized]
    if (jenv != NULL)
        ^~~~~~~~~~~~
/Users/antoinetoulme/w/evmc/bindings/java/c/host.c:439:12: note: uninitialized use occurs here
    return result;
           ^~~~~~
/Users/antoinetoulme/w/evmc/bindings/java/c/host.c:412:5: note: remove the 'if' if its condition is always true
    if (jenv != NULL)
    ^~~~~~~~~~~~~~~~~
/Users/antoinetoulme/w/evmc/bindings/java/c/host.c:407:5: note: variable 'result' is declared here
    struct evmc_tx_context result;
    ^
/Users/antoinetoulme/w/evmc/bindings/java/c/host.c:450:9: error: variable 'result' is used uninitialized whenever 'if' condition is false [-Werror,-Wsometimes-uninitialized]
    if (jenv != NULL)
        ^~~~~~~~~~~~
/Users/antoinetoulme/w/evmc/bindings/java/c/host.c:481:12: note: uninitialized use occurs here
    return result;
           ^~~~~~
/Users/antoinetoulme/w/evmc/bindings/java/c/host.c:450:5: note: remove the 'if' if its condition is always true
    if (jenv != NULL)
    ^~~~~~~~~~~~~~~~~
/Users/antoinetoulme/w/evmc/bindings/java/c/host.c:445:5: note: variable 'result' is declared here```

axic: I think the preference would be to individually fix these issues, in separate PRs. I agree this extra conditional isn't the nicest, and probably we have multiple options:
1) Move the declarations around so the compiler won't complain
2) Instead of the conditional, throw a Java exception?
3) Instead of conditional, add an assertion (does an abort).

PR #531, opened 2020-07-27
Create a way to run the linux build with docker (for mac), adjust the code

atoulme:
Scaled down version of #530 : just fix the C code so cmake compiles, removing the absolute paths in the code and `System.exit(1)`.

Add a Maven build so the gradle build can generate Maven artifacts.

### DIFF
--- a/bindings/java/Makefile
+++ b/bindings/java/Makefile
@@ -8,7 +8,9 @@ SOURCE_DIR = $(realpath $(CURDIR)/../..)
 OUT_DIR = $(CURDIR)/c/build
 BUILD_DIR = $(OUT_DIR)/_cmake_build
 
-build: gradlew $(OUT_DIR)/lib/libevmc-java.so
+all:$(CURDIR)/java/src/test/resources/libevmc.so
+
+build: gradlew
 	mkdir -p ./java/build
 	./gradlew --no-daemon clean spotlessApply build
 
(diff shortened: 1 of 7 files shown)


bindings/java/Makefile
```diff
@@ -18,7 +20,16 @@ test: build
 gradlew:
 	gradle --no-daemon setup
 
-$(OUT_DIR)/lib/libevmc-java.so:
+docker-linux: docker/linux.Dockerfile
```
jrhea, line 23: Trying to understand the need for a docker-linux rule in the Makefile.  iirc, the CI job uses a linux container to build.   I'm guessing you were developing on MacOs and needed a way to generate .so files for linux? 

Is the end goal here to provide support for MacOs and Linux?  If so, instead of Making an explicit docker-linux rule...I would detect the OS in the Makefile and modify a variable that contains the flags/includes/paths in a conditional and leave the rule alone.  I think the original PR to add Java-Bindings had this feature, but I imagine it was removed when getting it to work with the CMAKE build process.  

@axic / @chfast can @atoulme conform to your build process and add support for producing shared libs for different platforms (Linux, MacOs, Windows, etc)?

chfast:
Having maven support, can we just use maven for builds?

atoulme:
Maven support - this is just to publish to a Maven repository.

PR #532, opened 2020-08-03
Essential fixes for EVMC Java bindings

atoulme:
A set of critical fixes for the Java bindings, removing the cruft of other fixes I added elsewhere.

### DIFF
--- a/bindings/java/Makefile
+++ b/bindings/java/Makefile
@@ -22,6 +22,9 @@ $(OUT_DIR)/lib/libevmc-java.so:
 	mkdir -p $(BUILD_DIR)
 	(cd $(BUILD_DIR) && cmake $(SOURCE_DIR) -DCMAKE_INSTALL_PREFIX=$(OUT_DIR) -DEVMC_JAVA=ON -DJAVA_HOME=$(JAVA_HOME) -DEVMC_EXAMPLES=ON)
 	cmake --build $(OUT_DIR)/_cmake_build --target install
+	mkdir -p $(CURDIR)/java/src/test/resources
+	cp $(OUT_DIR)/lib/libevmc-java.so $(CURDIR)/java/src/test/resources/libevmc.so
+	cp $(OUT_DIR)/lib/libexample-vm.so $(CURDIR)/java/src/test/resources/libexample-vm.so
 
 clean:
 	rm -rf $(OUT_DIR)
--- a/bindings/java/c/evmc-vm.c
+++ b/bindings/java/c/evmc-vm.c
@@ -131,6 +131,15 @@ JNIEXPORT jint JNICALL Java_org_ethereum_evmc_EvmcVm_set_1option(JNIEnv* jenv,
     return (jint)option_result;
 }
 
+JNIEXPORT jlong JNICALL Java_org_ethereum_evmc_EvmcVm_address(JNIEnv* jenv,
+                                                              jclass jcls,
+                                                              jobject buf)
+{
+    (void)jcls;
+    void* p = (*jenv)->GetDirectBufferAddress(jenv, buf);
+    return (jlong)p;
+}
+
 JNIEXPORT jint JNICALL Java_org_ethereum_evmc_EvmcVm_get_1result_1size(JNIEnv* jenv, jclass jcls)
 {
     (void)jenv;
--- a/bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
+++ b/bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
@@ -6,7 +6,6 @@ package org.ethereum.evmc;
 import org.ethereum.evmc.EvmcLoaderException;
 
 import java.nio.ByteBuffer;
-import java.util.Objects;
 
 /**
  * The Java interface to the evm instance.
@@ -14,30 +13,18 @@ import java.util.Objects;
  * <p>Defines the Java methods capable of accessing the evm implementation.
  */
 public final class EvmcVm implements AutoCloseable {
-  private static EvmcVm evmcVm;
-  private static boolean isEvmcLibraryLoaded = false;
   private ByteBuffer nativeVm;
+
   /**
    * This method loads the specified evm shared library and loads/initializes the jni bindings.
    *
+   * @param evmcPath the path to the evmc shared library
    * @param filename /path/filename of the evm shared object
    * @throws EvmcLoaderException
    */
-  public static EvmcVm create(String filename) throws EvmcLoaderException {
-    if (!EvmcVm.isEvmcLibraryLoaded) {
-      try {
-        // load so containing the jni bindings to evmc
-        System.load(System.getProperty("user.dir") + "/../c/build/lib/libevmc-java.so");
-        EvmcVm.isEvmcLibraryLoaded = true;
-      } catch (UnsatisfiedLinkError e) {
-        System.err.println("Native code library failed to load.\n" + e);
-        System.exit(1);
-      }
-    }
-    if (Objects.isNull(evmcVm)) {
-      evmcVm = new EvmcVm(filename);
-    }
-    return evmcVm;
+  public static EvmcVm create(String evmcPath, String filename) {
+    System.load(evmcPath);
+    return new EvmcVm(filename);
   }
 
   private EvmcVm(String filename) throws EvmcLoaderException {
@@ -153,14 +140,15 @@ public final class EvmcVm implements AutoCloseable {
   private static native int get_result_size();
 
   /**
-   * This method cleans up resources
-   *
-   * @throws Exception
+   * Utility method to get a bytebuffer address
+   * @param buffer the byte buffer to consider
+   * @return the byte buffer address
    */
+  native long address(ByteBuffer buffer);
+
+  /** This method cleans up resources */
   @Override
-  public void close() throws Exception {
+  public void close() {
     destroy(nativeVm);
-    isEvmcLibraryLoaded = false;
-    evmcVm = null;
   }
 }
--- a/bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
+++ b/bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
@@ -144,7 +144,7 @@ public final class EvmcVm implements AutoCloseable {
    * @param buffer the byte buffer to consider
    * @return the byte buffer address
    */
-  native long address(ByteBuffer buffer);
+  public native long address(ByteBuffer buffer);
 
   /** This method cleans up resources */
   @Override
--- a/bindings/java/java/src/main/java/org/ethereum/evmc/HostContext.java
+++ b/bindings/java/java/src/main/java/org/ethereum/evmc/HostContext.java
@@ -145,4 +145,9 @@ public interface HostContext {
    * @param topicCount The number of the topics. Valid values are between 0 and 4 inclusively.
    */
   void emitLog(byte[] address, byte[] data, int dataSize, byte[][] topics, int topicCount);
+
+  /**
+   * Closes the context and frees all underlying resources.
+   */
+  default void close() {};
 }
--- a/bindings/java/java/src/test/java/org/ethereum/evmc/EvmcTest.java
+++ b/bindings/java/java/src/test/java/org/ethereum/evmc/EvmcTest.java
@@ -9,20 +9,21 @@ import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 final class EvmcTest {
+  private static final String libEvmcPath = EvmcTest.class.getResource("/libevmc.so").getFile();
   private static final String exampleVmPath =
-      System.getProperty("user.dir") + "/../c/build/lib/libexample-vm.so";
+      EvmcTest.class.getResource("/libexample-vm.so").getFile();
 
   @Test
   void testInitCloseDestroy() throws Exception {
     Assertions.assertDoesNotThrow(
         () -> {
-          try (EvmcVm vm = EvmcVm.create(exampleVmPath)) {}
+          try (EvmcVm vm = EvmcVm.create(libEvmcPath, exampleVmPath)) {}
         });
   }
 
   @Test
   void testAbiVersion() throws Exception {
-    try (EvmcVm vm = EvmcVm.create(exampleVmPath)) {
+    try (EvmcVm vm = EvmcVm.create(libEvmcPath, exampleVmPath)) {
       int abiVersion = vm.abi_version();
       assert (abiVersion > 0);
     }
@@ -30,7 +31,7 @@ final class EvmcTest {
 
   @Test
   void testName() throws Exception {
-    try (EvmcVm vm = EvmcVm.create(exampleVmPath)) {
+    try (EvmcVm vm = EvmcVm.create(libEvmcPath, exampleVmPath)) {
       String name = vm.name();
       assert (name.length() > 0);
       assert (name.equals("example_vm"));
@@ -39,7 +40,7 @@ final class EvmcTest {
 
   @Test
   void testVersion() throws Exception {
-    try (EvmcVm vm = EvmcVm.create(exampleVmPath)) {
+    try (EvmcVm vm = EvmcVm.create(libEvmcPath, exampleVmPath)) {
       String version = vm.version();
       assert (version.length() >= 5);
     }
@@ -47,7 +48,7 @@ final class EvmcTest {
 
   @Test
   void testExecute_returnAddress() throws Exception {
-    try (EvmcVm vm = EvmcVm.create(exampleVmPath)) {
+    try (EvmcVm vm = EvmcVm.create(libEvmcPath, exampleVmPath)) {
       HostContext context = new TestHostContext();
       int BYZANTIUM = 4;
       int EVMC_CALL = 0;
@@ -77,7 +78,7 @@ final class EvmcTest {
   /** Tests callbacks: get_storage_fn & set_storage_fn */
   @Test
   void testExecute_counter() throws Exception {
-    try (EvmcVm vm = EvmcVm.create(exampleVmPath)) {
+    try (EvmcVm vm = EvmcVm.create(libEvmcPath, exampleVmPath)) {
       HostContext context = new TestHostContext();
       int BYZANTIUM = 4;
       int EVMC_CALL = 0;
@@ -107,7 +108,7 @@ final class EvmcTest {
   /** Tests callbacks: get_tx_context_fn */
   @Test
   void testExecute_returnBlockNumber() throws Exception {
-    try (EvmcVm vm = EvmcVm.create(exampleVmPath)) {
+    try (EvmcVm vm = EvmcVm.create(libEvmcPath, exampleVmPath)) {
       HostContext context = new TestHostContext();
       int BYZANTIUM = 4;
       int EVMC_CALL = 0;
@@ -137,7 +138,7 @@ final class EvmcTest {
   /** Tests callbacks: get_tx_context_fn & set_storage_fn */
   @Test
   void testExecute_saveReturnBlockNumber() throws Exception {
-    try (EvmcVm vm = EvmcVm.create(exampleVmPath)) {
+    try (EvmcVm vm = EvmcVm.create(libEvmcPath, exampleVmPath)) {
       HostContext context = new TestHostContext();
       int BYZANTIUM = 4;
       int EVMC_CALL = 0;
@@ -169,7 +170,7 @@ final class EvmcTest {
   /** Tests callbacks: call_fn */
   @Test
   void testExecute_makeCall() throws Exception {
-    try (EvmcVm vm = EvmcVm.create(exampleVmPath)) {
+    try (EvmcVm vm = EvmcVm.create(libEvmcPath, exampleVmPath)) {
       HostContext context = new TestHostContext();
       int BYZANTIUM = 4;
       int EVMC_CALL = 0;
@@ -207,7 +208,7 @@ final class EvmcTest {
 
   @Test
   void testExecute_EVMC_CREATE() throws Exception {
-    try (EvmcVm vm = EvmcVm.create(exampleVmPath)) {
+    try (EvmcVm vm = EvmcVm.create(libEvmcPath, exampleVmPath)) {
       HostContext context = new TestHostContext();
       int BYZANTIUM = 4;
       int EVMC_CREATE = 3;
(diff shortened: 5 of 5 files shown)


bindings/java/Makefile
```diff
@@ -22,6 +22,9 @@ $(OUT_DIR)/lib/libevmc-java.so:
 	mkdir -p $(BUILD_DIR)
 	(cd $(BUILD_DIR) && cmake $(SOURCE_DIR) -DCMAKE_INSTALL_PREFIX=$(OUT_DIR) -DEVMC_JAVA=ON -DJAVA_HOME=$(JAVA_HOME) -DEVMC_EXAMPLES=ON)
 	cmake --build $(OUT_DIR)/_cmake_build --target install
+	mkdir -p $(CURDIR)/java/src/test/resources
+	cp $(OUT_DIR)/lib/libevmc-java.so $(CURDIR)/java/src/test/resources/libevmc.so
+	cp $(OUT_DIR)/lib/libexample-vm.so $(CURDIR)/java/src/test/resources/libexample-vm.so
```
atoulme, line 27: Copy files over to test folder so it gets picked up by tests.

bindings/java/c/evmc-vm.c
```diff
@@ -138,6 +138,18 @@ JNIEXPORT jint JNICALL Java_org_ethereum_evmc_EvmcVm_set_1option(JNIEnv* jenv,
     return (jint)option_result;
 }
 
+JNIEXPORT jlong JNICALL Java_org_ethereum_evmc_EvmcVm_address(JNIEnv* jenv,
```
atoulme, line 141: Add a function to get a bytebuffer address - used to store input data.

bindings/java/c/host.c
```diff
@@ -117,159 +117,155 @@ static enum evmc_storage_status set_storage_fn(struct evmc_host_context* context
...
-        jobject jresult =
-            (*jenv)->CallStaticObjectMethod(jenv, host_class, method, jcontext_index, jaddress);
-        assert(jresult != NULL);
-
-        evmc_uint256be* result_ptr =
-            (evmc_uint256be*)(*jenv)->GetDirectBufferAddress(jenv, jresult);
-        assert(result_ptr != NULL);
-        result = *result_ptr;
-
-        (*jenv)->ReleaseByteArrayElements(jenv, jaddress, (jbyte*)address, 0);
```
atoulme, line 194: This was segfaulting when running against vm reference tests. Removed that line.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
```diff
@@ -7,37 +7,24 @@
...
-        System.exit(1);
-      }
-    }
-    if (Objects.isNull(evmcVm)) {
-      evmcVm = new EvmcVm(filename);
-    }
-    return evmcVm;
+  public static EvmcVm create(String evmcPath, String filename) {
+    System.load(evmcPath);
+    return new EvmcVm(filename);
```
atoulme, line 27: allow to reuse a VM, so remove the static singleton, remove System.exit, and just load the evmc with a file path to the evmc native library.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
```diff
@@ -156,14 +143,15 @@ public int set_option(String name, String value) {
...
 
   /**
-   * This method cleans up resources
-   *
-   * @throws Exception
+   * Utility method to get a bytebuffer address
+   * @param buffer the byte buffer to consider
+   * @return the byte buffer address
    */
+  native long address(ByteBuffer buffer);
```
atoulme, line 150: Adds a method to get the native address of a direct bytebuffer. It's not exposed in Java.

bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
```diff
@@ -77,12 +77,12 @@ static ByteBuffer copy_code(int context_index, byte[] address, int code_offset)
             "HostContext does not exist for context_index: " + context_index);
     byte[] code = context.getCode(address).array();
 
-    if (code != null && code_offset > 0 && code_offset < code.length) {
+    if (code_offset < code.length) {
       int length = code.length - code_offset;
-      return ByteBuffer.wrap(code, 0, length);
+      return ByteBuffer.allocateDirect(length).put(code, code_offset, length);
```
atoulme, line 82: Remove wrap - that uses a heap buffer. Use a direct buffer instead.

bindings/java/java/src/main/java/org/ethereum/evmc/HostContext.java
```diff
@@ -145,4 +145,9 @@
    * @param topicCount The number of the topics. Valid values are between 0 and 4 inclusively.
    */
   void emitLog(byte[] address, byte[] data, int dataSize, byte[][] topics, int topicCount);
+
+  /**
+   * Closes the context and frees all underlying resources.
+   */
+  default void close() {};
```
atoulme, line 152: Add a way to allow context impls to release whatever resources they hang on to.

bindings/java/java/src/test/java/org/ethereum/evmc/EvmcTest.java
```diff
@@ -9,28 +9,29 @@
 import org.junit.jupiter.api.Test;
 
 final class EvmcTest {
+  private static final String libEvmcPath = EvmcTest.class.getResource("/libevmc.so").getFile();
   private static final String exampleVmPath =
-      System.getProperty("user.dir") + "/../c/build/lib/libexample-vm.so";
+      EvmcTest.class.getResource("/libexample-vm.so").getFile();
```
atoulme, line 14: Load from classpath so this works without involving relative home paths.

bindings/java/java/src/test/java/org/ethereum/evmc/TestHostContext.java
```diff
@@ -23,7 +23,7 @@ public int setStorage(byte[] address, byte[] key, byte[] value) {
 
   @Override
   public ByteBuffer getBalance(byte[] address) {
-    return ByteBuffer.allocateDirect(64).put(new byte[64]);
+    return ByteBuffer.allocateDirect(32).put(new byte[32]);
```
atoulme, line 26: Balance should be 32 bytes big endian.

bindings/java/Makefile
jrhea: To clarify, this change was made in support of this:


https://github.com/ethereum/evmc/pull/532/files/0c9eb8e30b24e18e2910aef98c8a84c33d1360a8#r464218230

right?

bindings/java/c/host.c
jrhea: We should open an issue to develop a more complete set of unit tests to catch this.  

If this causes a segfault, then I would think these functions would have the same problem:

in `get_code_hash_fn()`:
https://github.com/ethereum/evmc/blob/0c9eb8e30b24e18e2910aef98c8a84c33d1360a8/bindings/java/c/host.c#L268

in `copy_code_fn()`:
https://github.com/ethereum/evmc/blob/0c9eb8e30b24e18e2910aef98c8a84c33d1360a8/bindings/java/c/host.c#L317

bindings/java/c/host.c
```diff
@@ -35,80 +35,80 @@ static jbyteArray CopyDataToJava(JNIEnv* jenv, const void* ptr, size_t size)
 
 static bool account_exists_fn(struct evmc_host_context* context, const evmc_address* address)
 {
+#ifdef DEBUG
+    printf("Entering account_exists_fn\n");
+#endif
```
jrhea, line 40: Debug printouts should probably be removed

bindings/java/c/host.c
```diff
@@ -35,80 +35,80 @@ static jbyteArray CopyDataToJava(JNIEnv* jenv, const void* ptr, size_t size)
...
     return result;
 }
 
 static evmc_bytes32 get_storage_fn(struct evmc_host_context* context,
                                    const evmc_address* address,
                                    const evmc_bytes32* key)
 {
+#ifdef DEBUG
+    printf("Entering get_storage_fn\n");
+#endif
```
jrhea, line 77: Debug printouts should probably be removed

bindings/java/c/host.c
```diff
@@ -117,159 +117,155 @@ static enum evmc_storage_status set_storage_fn(struct evmc_host_context* context
                                                const evmc_bytes32* key,
                                                const evmc_bytes32* value)
 {
+#ifdef DEBUG
+    printf("Entering set_storage_fn\n");
+#endif
```
jrhea, line 122: Debug printouts should probably be removed

bindings/java/c/host.c
```diff
@@ -117,159 +117,155 @@ static enum evmc_storage_status set_storage_fn(struct evmc_host_context* context
...
+    result = (enum evmc_storage_status)jresult;
+
     return result;
 }
 
 static evmc_uint256be get_balance_fn(struct evmc_host_context* context, const evmc_address* address)
 {
+#ifdef DEBUG
+    printf("Entering get_balance_fn\n");
+#endif
```
jrhea, line 162: Debug printouts should probably be removed

bindings/java/c/host.c
```diff
@@ -117,159 +117,155 @@ static enum evmc_storage_status set_storage_fn(struct evmc_host_context* context
...
+    result = *result_ptr;
+
     return result;
 }
 
 static size_t get_code_size_fn(struct evmc_host_context* context, const evmc_address* address)
 {
+#ifdef DEBUG
+    printf("Entering get_code_size_fn\n");
+#endif
```
jrhea, line 201: Debug printouts should probably be removed

bindings/java/c/host.c
```diff
@@ -117,159 +117,155 @@ static enum evmc_storage_status set_storage_fn(struct evmc_host_context* context
...
+    result = (size_t)jresult;
+
     return result;
 }
 
 static evmc_bytes32 get_code_hash_fn(struct evmc_host_context* context, const evmc_address* address)
 {
+#ifdef DEBUG
+    printf("Entering get_code_hash_fn\n");
+#endif
```
jrhea, line 235: Debug printouts should probably be removed

bindings/java/c/host.c
```diff
@@ -280,84 +276,82 @@ static size_t copy_code_fn(struct evmc_host_context* context,
                            size_t buffer_size)
 {
     (void)buffer_size;  // FIXME: buffer_size suspiciously unused.
+#ifdef DEBUG
+    printf("Entering copy_code_fn");
+#endif
```
jrhea, line 281: Debug printouts should probably be removed

bindings/java/c/host.c
```diff
@@ -280,84 +276,82 @@ static size_t copy_code_fn(struct evmc_host_context* context,
...
     return result;
 }
 
 static void selfdestruct_fn(struct evmc_host_context* context,
                             const evmc_address* address,
                             const evmc_address* beneficiary)
 {
+#ifdef DEBUG
+    printf("Entering selfdestruct");
+#endif
```
jrhea, line 328: Debug printouts should probably be removed

bindings/java/c/evmc-vm.c
```diff
@@ -138,6 +138,18 @@ JNIEXPORT jint JNICALL Java_org_ethereum_evmc_EvmcVm_set_1option(JNIEnv* jenv,
...
+                                                              jclass jcls,
+                                                              jobject buf)
+{
+    (void)jcls;
+#ifdef DEBUG
+    printf("Entering address\n");
+#endif
```
jrhea, line 148: Debug printouts should probably be removed

bindings/java/c/host.c
```diff
@@ -35,80 +35,80 @@ static jbyteArray CopyDataToJava(JNIEnv* jenv, const void* ptr, size_t size)
...
 {
+#ifdef DEBUG
+    printf("Entering account_exists_fn\n");
+#endif
     bool result = false;
     const char java_method_name[] = "account_exists";
     const char java_method_signature[] = "(I[B)I";
     assert(context != NULL);
     JNIEnv* jenv = attach();
-    if (jenv != NULL)
```
jrhea, line 43: Can you remind me why all the jenv null checks were removed?

jrhea: Nvmd.  I suppose the attach() method's assert handles the case where AttachCurrentThread fails.

atoulme: cmake fails to compile otherwise.

bindings/java/c/evmc-vm.c
atoulme: I put a `#ifdef DEBUG` as a precompile directive. Is that not proper?

atoulme:
you found that the call to ReleaseByteArrayElements() in get_balance_fn() segfaults. I would think that get_code_hash_fn() and copy_code_fn() would have a similar problem.
> Somehow that doesn't affect that code. Not sure why.

wdyt about opening an issue for better code coverage? if my initial unit tests were more complete, then 👆 would have been caught sooner, right?
> I have a second PR with all EVM reference tests.

small nit about removing DEBUG printouts (not really my call though)
> I can remove.

bindings/java/Makefile
atoulme: Indeed.

bindings/java/c/host.c
atoulme: Yeah, neither fail. The reference tests should cover them.

Hard to tell for me why it segfaults. The JVM exits with error code 134, which means exit code 6 + 128.
Exit code 6 is a SIGABRT.

atoulme:
debug prints are removed.

In the next PR I run all VM tests and remove other `ReleaseByteArrayElements` calls - speeding up the tests. No memory problems seen.

bindings/java/c/evmc-vm.c
atoulme: removed

axic:
I'm sorry folks it will take a bit longer for us to do a review, a bit busy on a different project.

bindings/java/c/host.c
chfast: What is the CMake issue?

bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
axic: It seems this is more like a bugfix, because `wrap:
```
The new buffer will be backed by the given byte array; that is, modifications to the buffer will cause the array to be modified and vice versa. The new buffer's capacity will be array.length, its position will be offset, its limit will be offset + length, and its mark will be undefined. Its backing array will be the given array, and its array offset will be zero.
```

And unless `getCode` returns a new copy, this would mean the underlying code would be modifiable. Thought the `wrap` code is definitely more efficient in the sense it does no copying. Really need to verify what `getCode` does.

Can you pull out this change into a PR of its own?

Ref:
- https://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html#wrap(byte[],%20int,%20int)
- https://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html#allocateDirect(int)
- https://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html#put(byte[],%20int,%20int)

chfast: The docs of `.getCode()` says

> The Client MUST copy the requested code...

So changing wrap (no-copy) to copy looks good to me.

I'm also fine with leaving the fix here because this PR is "Essential fixes for EVMC Java bindings".

atoulme: right - I found out in testing that if I wrap and send to native C we get a crash. By using a direct buffer, we fix the issue.

axic: I'd argue either it is split out or this PR is made into having separate functional changes in separate commits. It is hard to review in its current form, especially as it reformats the C file.

atoulme: Separated here: https://github.com/ethereum/evmc/pull/535

axic: Meged :wink:

bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
```diff
@@ -77,7 +77,7 @@ static ByteBuffer copy_code(int context_index, byte[] address, int code_offset)
             "HostContext does not exist for context_index: " + context_index);
     byte[] code = context.getCode(address).array();
 
-    if (code != null && code_offset > 0 && code_offset < code.length) {
```
axic, line 80: Is this change still relevant? It seems to have been missed in #535.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
```diff
@@ -7,37 +7,24 @@
...
+   * @param evmcPath the path to the evmc shared library
    * @param filename /path/filename of the evm shared object
    */
-  public static EvmcVm create(String filename) {
-    if (!EvmcVm.isEvmcLibraryLoaded) {
-      try {
-        // load so containing the jni bindings to evmc
-        System.load(System.getProperty("user.dir") + "/../c/build/lib/libevmc-java.so");
-        EvmcVm.isEvmcLibraryLoaded = true;
-      } catch (UnsatisfiedLinkError e) {
```
axic, line 32: Couldn't `System.load` still emit this exception? Shouldn't `create` specify `throws Exception`?

bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
atoulme: Right - I can't remember if that change is required. I left it out. It doesn't hurt to check for those.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
atoulme: System.load does not throw this exception explicitly, so it doesn't have to be caught there. It can be caught by the code creating the EVM to try different paths in succession. The current code would run `System.exit(1)` if the lib was not found, which is a big problem if you're trying different paths.

Note also System.load is idempotent, so there's no need to check if it's loaded or not.

bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
atoulme: OK I remember now. The code can never be null, since it's a byte array extracted from a ByteBuffer object. The `code_offset > 0` should be removed imo since none of the other host impls check for that.

atoulme:
Ready for another look.

bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
chfast: `code_offset` can be 0.

axic: Actually this entire `copy_code_fn` and it's Java counterpart is broken. It never copies anything back, that's also why the FIXME for `buffer_size` is there.

I suggest this is fixed independently.

axic: I suggest this is left to be done in #545.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
axic: It does seem to throw exceptions: https://docs.oracle.com/javase/10/docs/api/java/lang/System.html#load(java.lang.String)

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
```diff
@@ -7,37 +7,24 @@
...
-        System.err.println("Native code library failed to load.\n" + e);
-        System.exit(1);
-      }
-    }
-    if (Objects.isNull(evmcVm)) {
-      evmcVm = new EvmcVm(filename);
-    }
-    return evmcVm;
+  public static EvmcVm create(String evmcPath, String filename) {
```
axic, line 25: Why aren't we following the old pattern of calling load once statically?
```
public final class EvmcVm {
   static {
      System.loadLibrary("libevmc.so");
   }
```

bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
```diff
@@ -146,7 +146,8 @@ static synchronized int addContext(HostContext context) {
   }
 
   static void removeContext(int index) {
-    contextList.remove(index);
+    HostContext context = contextList.remove(index);
+    context.close();
```
axic, line 150: Why not mark `HostContext` as `AutoCloseable` and then this is a given?

atoulme: `AutoCloseable` allows you to write syntactic sugar in java with the `try-with-resources` approach:
https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

Doesn't really apply here. Also that `close` method of `AutoCloseable` throws an Exception in its signature, so you'd have to deal with a try/catch.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
atoulme: Say you do this - this means the static block runs at the time the class is loaded.
That means less control as to when you perform the load. If the static block throws, the class can never be loaded.
A behavior I have seen in Java is this:
https://www.adamh.cz/blog/2012/12/how-to-load-native-jni-library-from-jar/
Basically, your java jar ships with the so/dylib/dll file embedded. On startup, it copies the file out to a temp directory. Then it loads from it.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
atoulme: The exceptions are documented in the Javadoc but not in the signature of the method.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
axic: Exactly that seems to be the best approach. `libemvc-java.so` is not something the user should supply or load multiple times, it is the binding.

bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
axic: Actually after #548 this is not needed.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
```diff
@@ -7,37 +7,24 @@
...
+    System.load(evmcPath);
```
axic, line 26: I think it is a bad idea to keep reloading this. I'd imagine that messes up the bound addresses and may result in a crash.

PR #535, opened 2020-09-27
java: Use allocateDirect instead of wrap in copy_code

atoulme:
Peeling off a change from #532 to single it out. This fixes a bug in how Java interfaces with JNI. JNI requires a direct buffer allocation instead of a heap bytebuffer.


chfast (approved):
Can you rebase? This should fix the CI.

axic (approved):

PR #536, opened 2020-09-27
Add a method to provide a bytebuffer address

atoulme:
Adds a native JNI method to provide the address of a ByteBuffer. This is used by the code to call out to EVMC which asks for the memory address of the ByteBuffer rather than its contents.

### DIFF
--- a/bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
+++ b/bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
@@ -156,6 +156,13 @@ public final class EvmcVm implements AutoCloseable {
   native int get_result_size();
 
   /**
+   * Utility method to get a bytebuffer address
+   * @param buffer the byte buffer to consider
+   * @return the byte buffer address
+   */
+  native long address(ByteBuffer buffer);
+
+  /**
    * This method cleans up resources
    *
    * @throws Exception
--- a/bindings/java/c/evmc-vm.c
+++ b/bindings/java/c/evmc-vm.c
@@ -138,6 +138,15 @@ JNIEXPORT jint JNICALL Java_org_ethereum_evmc_EvmcVm_set_1option(JNIEnv* jenv,
     return (jint)option_result;
 }
 
+JNIEXPORT jlong JNICALL Java_org_ethereum_evmc_EvmcVm_address(JNIEnv* jenv,
+                                                              jclass jcls,
+                                                              jobject buf)
+{
+    (void)jcls;
+    void* p = (*jenv)->GetDirectBufferAddress(jenv, buf);
+    return (jlong)p;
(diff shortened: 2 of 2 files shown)


bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
```diff
@@ -155,6 +155,13 @@ public int set_option(String name, String value) {
   /** get size of result struct */
   native int get_result_size();
 
+  /**
+   * Utility method to get a bytebuffer address
+   * @param buffer the byte buffer to consider
+   * @return the byte buffer address
+   */
+  native long address(ByteBuffer buffer);
```
axic, line 163: So how exactly is this used? This seems to be leaking native pointers to Java.

atoulme: It's not leaking because the ByteBuffer object is created by Java in native JNI code. It therefore is managed by Java.

It is used in exactly one place: when we perform internal EVM calls, we have to pass around the address of the input data byte buffer. We want to keep the same contract for the initial external call.

So in my kotlin code down the line I do this for external calls:
```val result =
      executeInternal(
        sender,
        destination,
        value,
        code,
        vm().address(inputDataBuffer),
        inputData.size(),
        gas,
        callKind,
        revision,
        depth,
        hostContext
      )```

For internal calls, the EVMMessage struct passed back has the pointer to the input data byte buffer, so we do this:
```
val result = ethereumVirtualMachine.executeInternal(
      evmMessage.sender,
      evmMessage.destination,
      evmMessage.value,
      Bytes.EMPTY,
      evmMessage.inputData,
      evmMessage.inputDataSize,
      evmMessage.gas,
      depth = depth + 1,
      hostContext = this
    )```

axic: In `HostContext`:
```
   * This function supports EVM calls.
   *
   * @param msg The call parameters.
   * @return The result of the call.
   */
  ByteBuffer call(ByteBuffer msg);
```

Why is this exposing an underlying C structure here? This should just wrap proper Java types.

atoulme: Right, I am not good at C. I dealt with it this way. Better devs needed.
I made it work for me for now :shrug:

axic: I'd name this `get_bytebuffer_ptr()` to be similar to `get_result_size()`. Both of these should be removed and `call` changed to actually take a message object. This can be done outside of this PR as I understand this change is needed to actually use it.

axic:
If #555 is finished it would make this PR obsolete.

PR #537, opened 2020-09-27
java: Remove all the unnecessary JNIEnv checks

atoulme:
Remove all the checks for jenv in the bindings code. If jenv is not there, it's better to crash.

### DIFF
--- a/bindings/java/c/host.c
+++ b/bindings/java/c/host.c
@@ -21,6 +21,7 @@ static JNIEnv* attach()
     jint rs = (*jvm)->AttachCurrentThread(jvm, (void**)&jenv, NULL);
     (void)rs;
     assert(rs == JNI_OK);
+    assert(jenv != NULL);
     return jenv;
 }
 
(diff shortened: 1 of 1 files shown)


bindings/java/c/host.c
```diff
@@ -40,31 +40,28 @@ static bool account_exists_fn(struct evmc_host_context* context, const evmc_addr
     const char java_method_signature[] = "(I[B)I";
     assert(context != NULL);
     JNIEnv* jenv = attach();
-    if (jenv != NULL)
```
axic, line 43: ~~If this is removed, then I think you should add an `assert(jenv != NULL)` to all cases it is removed, just like there's an assert for other cases.~~

Actually just add an assert for non-null into `attach` itself.

I'm not sure in what case can `attach` return null?

atoulme: There's no case where attach can return null. I can add an assert for jenv if you like.

axic: These asserts are not (should not be) part of the release binary, but helps for finding bugs testing with debug builds.

chfast (approved):

bindings/java/c/host.c
```diff
@@ -331,32 +312,29 @@ static void selfdestruct_fn(struct evmc_host_context* context,
...
+
+    // set java method params
+    jcontext_index = context->index;
+    jaddress = CopyDataToJava(jenv, address, sizeof(struct evmc_address));
+    jbeneficiary = CopyDataToJava(jenv, beneficiary, sizeof(struct evmc_address));
+
+    // call java method
+    (*jenv)->CallStaticIntMethod(jenv, host_class, method, jcontext_index, jaddress, jbeneficiary);
+
     return;
```
axic, line 338: Useless statement.

bindings/java/c/host.c
```diff
@@ -492,43 +463,41 @@ static void emit_log_fn(struct evmc_host_context* context,
...
+        jbyteArray jtopic = CopyDataToJava(jenv, topics[i].bytes, sizeof(struct evmc_bytes32));
+        (*jenv)->SetObjectArrayElement(jenv, jtopics, (jsize)i, jtopic);
+        (*jenv)->DeleteLocalRef(jenv, jtopic);
     }
+
+    // call java method
+    (*jenv)->CallStaticIntMethod(jenv, host_class, method, jcontext_index, jaddress, jdata,
+                                 data_size, jtopics, topics_count);
+
     return;
```
axic, line 501: Same here.

axic (approved):

PR #541, opened 2020-09-28
java: tiny bits of clean up to the host implementation

axic:


axic:
I think all the cases where `CopyDataToJava` is used it needs to be confirmed whether the object must be freed or not. In some cases it is freed, in others it is not.

atoulme (approved):
LGTM

bindings/java/c/host.c
```diff
@@ -36,149 +36,122 @@ static jbyteArray CopyDataToJava(JNIEnv* jenv, const void* ptr, size_t size)
...
-    jaddress = CopyDataToJava(jenv, address, sizeof(struct evmc_address));
+    jbyteArray jaddress = CopyDataToJava(jenv, address, sizeof(struct evmc_address));
 
     // call java method
-    jint jresult = (*jenv)->CallStaticIntMethod(jenv, host_class, method, jcontext_index, jaddress);
-    result = !!jresult;
-
-    return result;
+    jint jresult = (*jenv)->CallStaticIntMethod(jenv, host_class, method, context->index, jaddress);
+    return jresult == 1;
```
chfast, line 59:
```suggestion
    return jresult != 0;
```
More classic "convert int to bool".

axic: Fixed.

chfast (approved):

PR #542, opened 2020-09-29
java: clean up memory handling

axic:
Apparently the bindings should be leaking a lot of memory because references are never released, see https://www.ibm.com/support/knowledgecenter/en/SSYKE2_8.0.0/com.ibm.java.vm.80.doc/docs/jni_refs.html and https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#Release_PrimitiveType_ArrayElements_routines

### DIFF
--- a/bindings/java/c/host.c
+++ b/bindings/java/c/host.c
@@ -71,7 +71,10 @@ static bool account_exists_fn(struct evmc_host_context* context, const evmc_addr
     // call java method
     jboolean jresult =
         (*jenv)->CallStaticBooleanMethod(jenv, host_class, method, context->index, jaddress);
-    // FIXME: release jaddress?
+    (*jenv)->ReleaseByteArrayElements(jenv, jaddress, (jbyte*)address, JNI_ABORT);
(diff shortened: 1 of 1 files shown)


atoulme:
I ran the VM reference tests against the EVM in a continuous loop for several hours and didn't notice a leak. It's cool to dispose of memory explicitly. I believe the JVM disposes of ByteBuffer objects it creates through garbage collection otherwise.

axic:
> I ran the VM reference tests against the EVM in a continuous loop for several hours and didn't notice a leak. It's cool to dispose of memory explicitly. I believe the JVM disposes of ByteBuffer objects it creates through garbage collection otherwise.

My impression from the documentation was that these objects are part of garbage collection, but owned by the JNI interface, hence they need to be freed. They are not large objects however, so I can imagine it can lasts days before OOM.

Can you point me to the documentation where this is clarified?

Also note some of the memory freeing is already part of the code, it is inconsistent.

atoulme:
OK, I think I'm wrong here. You're right, let's dispose of memory explicitly.

axic:

PR #543, opened 2020-09-30
Add installation instructions

alita-moore:


README.md
```diff
@@ -13,6 +13,14 @@ to access Ethereum environment and state.
 
 ## Usage
 
+### Install
```
chfast, line 16: This should be a section before `## Usage`: `## Install`, as specified in https://github.com/RichardLitt/standard-readme/blob/master/spec.md#install

Also add subsection `### C/C++` as other languages have different instructions.

README.md
```diff
@@ -13,6 +13,14 @@ to access Ethereum environment and state.
...
+
+```bash
+cmake .
```
chfast, line 19: It's better to build in separate dir.
```
mkdir build && cd build
cmake ..
cmake --build .
cmake --build . --target install
```

chfast (changes requested):

README.md
alita-moore: > This should be a section before `## Usage`: `## Install`, as specified in https://github.com/RichardLitt/standard-readme/blob/master/spec.md#install
> 
> Also add subsection `### C/C++` as other languages have different instructions.

Are there any required / unusual packages? For example CL11? 

I have had success in building with gcc < 6 but the docs suggest 6+. Should I add that as well? 

Is this meant to be used as an external C library or as a system package?

chfast: > Are there any required / unusual packages? For example CL11?

This is handled by CMake.

> I have had success in building with gcc < 6 but the docs suggest 6+. Should I add that as well?

I don't remember why GCC 6 is minimal, probably this was the minimal version we were testing with when the docs were added: https://github.com/ethereum/evmc/commit/35bf88d8acf77e371924a59e40c2afb405f702c6.

It looks like GCC 5 has full C++14 support, so should be fine too. What version did you use? If you think this is important, we can address the issue.

> Is this meant to be used as an external C library or as a system package?

I'm not sure I understand the issue. The instructions will install EVMC libraries, tools, and CMake package config on system level. But where it is installed can be controlled with `-DCMAKE_INSTALL_PREFIX` variable.

chfast (commented):
Looks good, but please squash commits into single one.

README.md
alita-moore: I don't think it's too important, so it's okay.

Primarily, my confusion is in how to most effectively use this library. It would be very convenient to execute bytecode from the console directly, if that were possible.

alita-moore:
I'm not certain if the PR was properly squashed because I couldn't also squash the updates. Regardless, 19ffe51 should be sufficient

README.md
chfast: This may not be what you are looking for. The EVMC project defines API for EVM implementations but does not provide one. It includes some additional testing tools and CLI EVM execution tool, but you need to get the EVM implementation from some other source. If you need full EVM implementation, please see [evmone](https://github.com/ethereum/evmone).

chfast (approved):

README.md
alita-moore: it is correct, I'm just not certain how to apply it. I'll try to read through the examples more, but if possible a simple step-by step tutorial on how to use evm1 with this would be really helpful :)

PR #545, opened 2020-09-30
java: properly implement code_copy_fn

axic:


bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
```diff
@@ -70,16 +70,20 @@ static ByteBuffer get_code_hash(int context_index, byte[] address) {
...
             getContext(context_index),
             "HostContext does not exist for context_index: " + context_index);
     byte[] code = context.getCode(address).array();
 
-    if (code != null && code_offset > 0 && code_offset < code.length) {
-      int length = code.length - code_offset;
-      return ByteBuffer.allocateDirect(length).put(code, code_offset, length);
+    /* Ensure code is returned and both offset/length are positive numbers. */
+    if (code != null && offset >= 0 && length >= 0 && offset <= code.length) {
+      int length = Math.min(length, code.length - offset);
```
axic, line 82: We can decide whether to do this slicing here or in C.

bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
```diff
@@ -70,16 +70,17 @@ static ByteBuffer get_code_hash(int context_index, byte[] address) {
...
             getContext(context_index),
             "HostContext does not exist for context_index: " + context_index);
     byte[] code = context.getCode(address).array();
 
-    if (code != null && code_offset > 0 && code_offset < code.length) {
-      int length = code.length - code_offset;
-      return ByteBuffer.allocateDirect(length).put(code, code_offset, length);
+    /* Ensure code is returned and both offset/length are positive numbers. */
+    if (code != null) {
+      // TODO: use wrap here, no need to duplicate this given it is copied again.
```
atoulme, line 82: Not sure I understand the TODO.

atoulme: To be less terse: this TODO says we can use `ByteBuffer.wrap`, am I reading this right? But for C code to interact with ByteBuffer, you need to use a direct memory allocation. I had panics when I tried to use heap-based bytebuffers.

atoulme, line 82: This TODO says we can use ByteBuffer.wrap, am I reading this right? But for C code to interact with ByteBuffer, you need to use a direct memory allocation. I had panics when I tried to use heap-based bytebuffers.

axic: I did not see this difference in the JNI docs, only whether it is copied or not. And here we do not need copying given the C counterpart does a copy anyway.

Can you point me to the documentation where this is mentioned?

axic:
@chfast which one do you prefer, the first or the second commit?

bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
atoulme: I think the code used to get the pointer of the bytebuffer, and that would crash the JVM. So I think you're saying that's all changed now - I'm good with a change there, and I'll make sure to give you a reproduction case if that happens again.

axic: I think you are right, documentation does seem to claim that `allocateDirect` will 100% be laid on a way it can be accessed without copying, while other may require copying. It is not clear however who does that copying, but it seems it is done by the JVM so it would mean a slowdown.

bindings/java/c/host.c
```diff
@@ -254,23 +258,28 @@ static size_t copy_code_fn(struct evmc_host_context* context,
...
-                                                      jaddress, jcode_offset);
+    jobject jresult =
+        (*jenv)->CallStaticObjectMethod(jenv, host_class, method, context->index, jaddress);
     assert(jresult != NULL);
 
     // copy jresult back to buffer_data
-    buffer_data = (uint8_t*)(*jenv)->GetDirectBufferAddress(jenv, jresult);
-    (void)buffer_data;
-    assert(buffer_data != NULL);
+    uint8_t* code = (uint8_t*)(*jenv)->GetDirectBufferAddress(jenv, jresult);
```
axic, line 268: This should use `GetDirectBuffer` from #552.

axic:
@chfast still needs your review

bindings/java/c/host.c
chfast: Looks unfinished.

axic: You can ignore that, still waiting for an answer on this: https://github.com/ethereum/evmc/pull/545#issuecomment-705864279

axic:
@chfast okay this is my final proposal.

bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
```diff
@@ -16,6 +16,13 @@
  * <p>The set of all callback functions expected by VM instances.
  */
 final class Host {
+  static private ByteBuffer ensureDirectBuffer(ByteBuffer input) {
+    // Reallocate if needed.
+    if (!input.isDirect())
```
chfast, line 21: No idea what it does. Is this even tested?

axic: Not all of them are, but `get/set_storage` is. This helper enforces that the `ByteBuffer` is in a format ("direct buffeR") which can be queried from JNI (e.g. the `GetDirectBuffer` helper in C won't assert).

chfast (approved):

PR #546, opened 2020-10-01
cmake: Add example end-to-end tests to CTest

chfast:
Include EVMC examples tests in CTest.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -66,10 +66,6 @@ commands:
           name: "Test"
           command: |
             cmake --build ~/build --target test -- ARGS="-j4 --schedule-random --output-on-failure"
-            # Test statically linked end-to-end example
-            ~/build/examples/evmc-example-static
-            # Test dynamically loaded end-to-end example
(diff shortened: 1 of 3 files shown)


gumb0 (approved):

circle.yml
```diff
@@ -66,10 +66,6 @@ commands:
           name: "Test"
           command: |
             cmake --build ~/build --target test -- ARGS="-j4 --schedule-random --output-on-failure"
-            # Test statically linked end-to-end example
-            ~/build/examples/evmc-example-static
-            # Test dynamically loaded end-to-end example
-            ~/build/examples/evmc-example ~/build/examples/example_vm/libexample-vm.so
```
axic, line 72: All of these are still built, right? I think these artifacts are used by some integration.

axic (approved):
Otherwise looks good.

circle.yml
chfast: Yes. Now CTest handles execution of these lines.

PR #549, opened 2020-10-09
java: clean up the EvmcVm class a bit

axic:
1. Make the native methods explicitly private (they are not supposed to be available due to requiring a VM pointer).
2. Properly return exception on loading failures.

Some references:
- https://developer.android.com/training/articles/perf-jni#java
- https://www.protechtraining.com/content/java_fundamentals_tutorial-_java_native_interface_jni
- https://rkennke.wordpress.com/2007/07/30/efficient-jni-programming-iv-wrapping-native-data-objects/


bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
```diff
@@ -92,14 +94,14 @@ public String version() {
...
-  native void destroy(ByteBuffer nativeVm);
+  private static native void destroy(ByteBuffer nativeVm);
 
   /**
    * Function to execute a code by the VM instance.
    *
    * <p>This is a mandatory method and MUST NOT be set to NULL.
    */
-  native void execute(
+  private static native void execute(
```
atoulme, line 104: Why is this static?

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcLoaderException.java
```diff
@@ -0,0 +1,10 @@
+// EVMC: Ethereum Client-VM Connector API.
+// Copyright 2019-2020 The EVMC Authors.
+// Licensed under the Apache License, Version 2.0.
+package org.ethereum.evmc;
+
+public class EvmcLoaderException extends Exception {
```
atoulme, line 6: Consider making it extend RuntimeException - this way it's an unchecked exception.

atoulme:
Please don't make the methods static - this would only allow one EVM in a JVM at a time. Having multiple EVMs is a good thing.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
axic: Because it only exists once, it is the native library which can only be loaded once.

axic:
> Please don't make the methods static - this would only allow one EVM in a JVM at a time. Having multiple EVMs is a good thing.

The JNI binding (libevmc-java.so) should be loaded once. Multiple EVMC VMs can be loaded and that is distinguished by the `nativeVM` object which the JNI side looks up and dispatches.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
atoulme: OK, if it's purely functional, that may work.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcLoaderException.java
axic: But perhaps I don't fully understand the distinction between `Exception` and `RuntimeException` -- could you explain please?

atoulme: Sure thing. In Java, you _must_ catch checked exceptions, or the code doesn't compile. You can let unchecked exceptions "unchecked", meaning you don't need to catch them explicitly. They still happen and bubble up.
The net difference is that sometimes your Java code ends up looking like a bunch of try/catch statements or throws.
This is a subject of hot debate and frankly, I'm fine either way. I just prefer `RuntimeException`s myself.
https://stackoverflow.com/questions/6115896/understanding-checked-vs-unchecked-exceptions-in-java

axic: Thanks! My argument for checked exception here is that this EVMC-specific loading error should be handled at the place this is integrated in, likely it would be translated into some local exception in the project?

Though I don't have a very strong opinion, fine either way.

atoulme: OK, let’s do it.

axic:
@atoulme okay to merge this?

axic:
@atoulme merging this now to progress with the refactoring but we can revert certain changes if it turns out to be bad.

atoulme:
All good. Sorry for the slow response.

PR #550, opened 2020-10-09
java: remove the extra wrapping of HostContext and pass it directly

axic:
Depends on #549.

### DIFF
--- a/bindings/java/c/host.h
+++ b/bindings/java/c/host.h
@@ -12,11 +12,6 @@
 extern "C" {
 #endif
 
-struct evmc_host_context
-{
-    int index;
-};
-
 int evmc_java_set_jvm(JNIEnv*);
 const struct evmc_host_interface* evmc_java_get_host_interface();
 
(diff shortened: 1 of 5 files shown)


bindings/java/c/host.h
```diff
@@ -11,11 +11,6 @@
 extern "C" {
 #endif
 
-struct evmc_host_context
```
axic, line 14: The whole point of `evmc_host_context` is to be an opaque object from the interface perspective, and since it only needs to be valid during the scope of `execute` I think this change is the correct one.

atoulme:
I think that's ok, not sure I grasp it entirely. I thought the point of having this organization was to pass the context when doing internal calls, but I think you're saying that's not necessary.

axic:
> I think that's ok, not sure I grasp it entirely. I thought the point of having this organization was to pass the context when doing internal calls, but I think you're saying that's not necessary.

The context is needed for all the host functions, just that we can opaquely pass it around I think. We can only make this simplification because the scope is limited.

chfast (commented):
Looks nice. I assume that an `HostContext` object is not going to be garbage collected while being the argument of `execute()`.

axic:
> Looks nice. I assume that an HostContext object is not going to be garbage collected while being the argument of execute().

That is my assumption. We are safe to do this because `execute` is only executed in a single place where the context's lifetime outlives that of `execute`, so it won't be collected.

chfast (approved):

PR #551, opened 2020-10-09
java: add CopyFromBuffer helper and ensure the properly sized ByteBuffer is used

axic:


bindings/java/c/evmc-vm.c
```diff
@@ -88,6 +88,10 @@ JNIEXPORT void JNICALL Java_org_ethereum_evmc_EvmcVm_execute(JNIEnv* jenv,
                                                              jint jcode_size,
                                                              jobject jresult)
 {
+    jint rs = evmc_java_set_jvm(jenv);
```
axic, line 91: This is part of another PR. We should actually pass down the `jenv` via the context.

axic:
@atoulme @jrhea this is likely the reason you got crashes. The code was doing an unchecked copying of whatever was passed.

atoulme (approved):
👍

PR #552, opened 2020-10-09
java: remove code_size argument from execute()

axic:


axic:
@atoulme @jrhea this is another potential place where buffer overread could happen, the unchecked copying prior to this PR.

bindings/java/c/evmc-vm.c
```diff
@@ -90,12 +90,12 @@ JNIEXPORT void JNICALL Java_org_ethereum_evmc_EvmcVm_execute(JNIEnv* jenv,
                                                              jint jrev,
                                                              jobject jmsg,
                                                              jobject jcode,
-                                                             jint jcode_size,
                                                              jobject jresult)
 {
     (void)jcls;
     struct evmc_message* msg = (struct evmc_message*)(*jenv)->GetDirectBufferAddress(jenv, jmsg);
     assert(msg != NULL);
+    size_t code_size = (size_t)(*jenv)->GetDirectBufferCapacity(jenv, jcode);
```
chfast, line 98: Add assert that `code_size` is not negative. Probably should be done before casting to `size_t`.

chfast (changes requested):

bindings/java/c/evmc-vm.c
axic: Done.

chfast (approved):

chfast (approved):

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
```diff
@@ -114,11 +113,11 @@ native void execute(
    * <p>This allows the context to managed in one method
    */
   public synchronized ByteBuffer execute(
-      HostContext context, int rev, ByteBuffer msg, ByteBuffer code, int size) {
```
axic, line 117: Note: this is a breaking change, but the Java binding we have is not stable enough and more changes are coming.

Everything merged so far were non-breaking changes.

Issue #548, opened 2020-10-08
Clean up the Java interface


axic:
Currently the Java interface exposes a lot of internal C structures to Java:
- `evmc_tx_context` in `HostContext.getTxContext`
- `evmc_message` in `HostContext.call` and `EvmcVm.execute`

Besides that none of the basic types are represented as objects, but as `byte[]`, i.e. address, balance, storage key, storage value, etc.

It would make sense to create Java appropriates for
- the basic types
- transaction context and message (and these make use of those basic types also).

Then there needs to be some translation in place which can serialise/deserialise these to the appropriate byte representation expected by the C interface.

PR #555, opened 2020-10-09
java: introduce proper types and minimise reliance on ByteBuffer/byte[]

axic:
Part of #548.

### DIFF
--- a/bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
+++ b/bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
@@ -3,6 +3,7 @@
 // Licensed under the Apache License, Version 2.0.
 package org.ethereum.evmc;
 
+import org.ethereum.evmc.types.*;
 import java.nio.ByteBuffer;
 
 /**
--- a/bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
+++ b/bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
@@ -66,8 +66,8 @@ final class Host {
   }
 
   /** Call callback function. */
-  static ByteBuffer call(HostContext context, ByteBuffer msg) {
-    return ensureDirectBuffer(context.call(msg));
+  static Result call(HostContext context, Message msg) {
+    return context.call(msg);
   }
 
   /** Get transaction context callback function. */
--- a/bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
+++ b/bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
@@ -22,41 +22,41 @@ final class Host {
 
   /** Check account existence callback function. */
   static boolean account_exists(HostContext context, byte[] address) {
-    return context.accountExists(address);
+    return context.accountExists(new Address(address));
   }
 
   /** Get storage callback function. */
   static ByteBuffer get_storage(HostContext context, byte[] address, byte[] key) {
-    return ensureDirectBuffer(context.getStorage(address, key));
+    return context.getStorage(new Address(address), new Bytes32(key)).getByteBuffer();
   }
 
   /** Set storage callback function. */
   static int set_storage(HostContext context, byte[] address, byte[] key, byte[] value) {
-    return context.setStorage(address, key, value);
+    return context.setStorage(new Address(address), new Bytes32(key), new Bytes32(value));
   }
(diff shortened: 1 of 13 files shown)


bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
```diff
@@ -31,7 +33,7 @@ static ByteBuffer get_storage(int context_index, byte[] address, byte[] key) {
         requireNonNull(
             getContext(context_index),
             "HostContext does not exist for context_index: " + context_index);
-    return context.getStorage(address, key);
+    return context.getStorage(new Addres(address), new Bytes32(key)).getByteBuffer();
```
atoulme, line 36: typo Addres

bindings/java/java/src/main/java/org/ethereum/evmc/HostContext.java
```diff
@@ -44,7 +45,7 @@
    * @param value The value to be stored.
    * @return The effect on the storage item.
    */
-  int setStorage(byte[] address, byte[] key, byte[] value);
+  int setStorage(Address address, bytes32 key, Bytes32 value);
```
atoulme, line 48: bytes32 should be Bytes32

bindings/java/java/src/main/java/org/ethereum/evmc/types/Message.java
```diff
@@ -0,0 +1,45 @@
+// EVMC: Ethereum Client-VM Connector API.
+// Copyright 2019-2020 The EVMC Authors.
+// Licensed under the Apache License, Version 2.0.
+package org.ethereum.evmc.types;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+public class TxContext {
+  private final bool is32bit() {
```
atoulme, line 10: bool -> boolean
is32bit is not used.

atoulme:
I really like this idea. Would remove all guessing. Feels like you have some coding issues.

Do you intend for implementors to create Address, Bytes32 Uint256 objects? If so, they'll need public constructors.

axic:
@atoulme thanks! This is still work in progress, hence the draft mode, but I think this would be the safest way to handle things. Only issue is `Message` which is a mess due to the pointer thingie.

bindings/java/java/src/main/java/org/ethereum/evmc/types/Message.java
```diff
@@ -0,0 +1,47 @@
...
+  public ByteBuffer getByteBuffer() {
+    // This is big endian by default, which is what we want.
+    return ByteBuffer.allocateDirect(expectedLength() + this.input.remaining())
+        .putInt(this.kind)
+        .putInt(this.flags)
+        .putInt(this.depth)
+        .putLong(this.gas)
+        .put(this.destination.getByteBuffer())
+        .put(this.sender.getByteBuffer())
+        // TODO: change this offset field based on endianness
```
axic, line 40: Actually all these helpers operate in **current endianess** (https://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html#putInt(int)) so all this should be fine.

bindings/java/java/src/main/java/org/ethereum/evmc/types/Message.java
```diff
@@ -0,0 +1,47 @@
...
+  public Address destination;
+  public Address sender;
+  public ByteBuffer input;
+  public Uint256 value;
+  public Bytes32 create2_salt;
+
+  public Message() {}
+
+  public ByteBuffer getByteBuffer() {
+    // This is big endian by default, which is what we want.
```
axic, line 32: ByteBuffer is set to big-endian by default, but we want to set it to native endian (helper to set: https://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html#order(java.nio.ByteOrder)).

We can use `ByteOrder` for that: https://docs.oracle.com/javase/6/docs/api/java/nio/ByteOrder.html#nativeOrder()

axic:
Need to add Result class, and also pass Result/Message in HostContext.call.

axic:
This currently contains the desired API from the Java perspective, yet to actually implement all of this. I'm leaning towards now that the C part would need to just receive an instance of Result/Message and manually read/write the fields and compose the C struct, as opposed to trying to construct the C struct from within Java.

axic:

PR #557, opened 2020-10-19
java: statically load JNI library

axic:
This should implement the pattern of storing the shared object in the JAR file, copying it to a temp folder, and loading it from there.

### DIFF
--- a/bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
+++ b/bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
@@ -3,8 +3,11 @@
 // Licensed under the Apache License, Version 2.0.
 package org.ethereum.evmc;
 
+import java.io.IOException;
 import java.nio.ByteBuffer;
-import java.util.Objects;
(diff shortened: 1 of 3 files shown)


bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
```diff
@@ -14,30 +14,20 @@
  * <p>Defines the Java methods capable of accessing the evm implementation.
  */
 public final class EvmcVm implements AutoCloseable {
-  private static EvmcVm evmcVm;
-  private static boolean isEvmcLibraryLoaded = false;
   private ByteBuffer nativeVm;
+
+  static {
+    System.loadLibrary("evmc-java");
```
axic: Instead of this, we need copy the shared library out of the JAR into some temporary directory and load from there. See https://github.com/ethereum/evmc/pull/532#discussion_r497421023

axic:
Pulled in all the changes from #559.

atoulme (approved):
lgtm

chfast (approved):

PR #559, opened 2020-10-22
Java loader loaded

atoulme:
Trying out lazy loader with temp file


atoulme:
This can replace #557.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
```diff
@@ -3,41 +3,79 @@
...
+          Path evmcLib = Files.createTempFile("libevmc-java", extension);
+          Files.copy(
+              EvmcVm.class.getResourceAsStream("/libevmc-java." + extension),
+              evmcLib,
+              StandardCopyOption.REPLACE_EXISTING);
+          evmcLib.toFile().deleteOnExit();
+          try {
+            System.load(evmcLib.toAbsolutePath().toString());
+          } catch (UnsatisfiedLinkError e1) {
+            error = e1;
```
axic, line 49: Why is there a need for a local `error` variable? Because `errorLoadingEvmc` can be set once in the function?

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
```diff
@@ -3,41 +3,79 @@
...
   private ByteBuffer nativeVm;
+
+  static {
+    Throwable error = null;
+    try {
+      // load so containing the jni bindings to evmc
+      System.loadLibrary("libevmc-java");
+    } catch (UnsatisfiedLinkError e) {
+      String extension = null;
+      String operSys = System.getProperty("os.name").toLowerCase();
```
axic, line 28:
```suggestion
      String os = System.getProperty("os.name").toLowerCase();
```

axic:
@atoulme is this working now properly?

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
atoulme: Yes, it’s final, and therefore immutable.

atoulme:
It does. I will change the variable name in a separate commit. If you approve I’ll merge.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
atoulme: Changed to `os`

atoulme:
@axic @chfast can I get a review please? Quite a bit waiting downstream from it.

bindings/java/java/src/test/java/org/ethereum/evmc/EvmcTest.java
```diff
@@ -3,14 +3,28 @@
...
 import org.junit.jupiter.api.Test;
 
 final class EvmcTest {
-  private static final String exampleVmPath =
-      System.getProperty("user.dir") + "/../c/build/lib/libexample-vm.so";
+  private static final String extension;
+  private static final String exampleVmPath;
+  static {
+    String os = System.getProperty("os.name", "generic").toLowerCase();
+    if (os.contains("mac") || os.contains("darwin")) {
```
chfast, line 17: This seems inconsistent with the other helper that figures out the DLL extension.

bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
```diff
@@ -3,23 +3,17 @@
...
  *
  * <p>The set of all callback functions expected by VM instances.
  */
 final class Host {
-  static private ByteBuffer ensureDirectBuffer(ByteBuffer input) {
+  private static ByteBuffer ensureDirectBuffer(ByteBuffer input) {
     // Reallocate if needed.
-    if (!input.isDirect())
-      return ByteBuffer.allocateDirect(input.remaining()).put(input);
+    if (!input.isDirect()) return ByteBuffer.allocateDirect(input.remaining()).put(input);
```
chfast, line 16: Are these stylistic changes done manually or part of some language linter?

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
```diff
@@ -3,41 +3,79 @@
...
-      try {
-        // load so containing the jni bindings to evmc
-        System.load(System.getProperty("user.dir") + "/../c/build/lib/libevmc-java.so");
-        EvmcVm.isEvmcLibraryLoaded = true;
-      } catch (UnsatisfiedLinkError e) {
-        System.err.println("Native code library failed to load.\n" + e);
-        System.exit(1);
-      }
+    if (!isAvailable()) {
+      throw new EvmcLoaderException("Cannot load evmc native library", errorLoadingEvmc);
```
chfast, line 76:
```suggestion
      throw new EvmcLoaderException("Cannot load EVMC native library", errorLoadingEvmc);
```

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
```diff
@@ -3,41 +3,79 @@
...
+    }
+    errorLoadingEvmc = error;
+  }
+
+  /**
+   * Returns true if the native library was loaded successfully and EVMC capabilities are available.
+   *
+   * @return true if the library is available
+   */
+  public static boolean isAvailable() {
```
chfast, line 64: Must this be public? Seems to be only used inside `create()`. Drop it if not needed.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
```diff
@@ -3,41 +3,79 @@
...
 public final class EvmcVm implements AutoCloseable {
-  private static EvmcVm evmcVm;
-  private static boolean isEvmcLibraryLoaded = false;
+  private static final Throwable errorLoadingEvmc;
   private ByteBuffer nativeVm;
+
+  static {
+    Throwable error = null;
+    try {
+      // load so containing the jni bindings to evmc
```
chfast, line 24:
```suggestion
      // Load DLL containing the JNI bindings to EVMC.
```

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
```diff
@@ -3,41 +3,79 @@
...
+      } else if (os.contains("mac")) {
+        extension = "dylib";
+      } else {
+        error = e;
+      }
+      if (extension != null) {
+        try {
+          Path evmcLib = Files.createTempFile("libevmc-java", extension);
+          Files.copy(
+              EvmcVm.class.getResourceAsStream("/libevmc-java." + extension),
```
chfast, line 42: How is it placed here?

chfast, line 42: I'm not sure what is the best practice for this, but can you use fixed extension for all OSes? E.g. put it in the JAR as `java/src/main/resources/libevmc-java.jni` and the load as `System.load("/tmp/libevmc-java.jni")`. You will not have to figure out the OS-specific DLL extension.

Did you test it on Windows? On Windows the DLLs don't have the "lib" prefix.

bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
atoulme: yes, it's using (spotless)[https://github.com/diffplug/spotless].

Might be better to use braces for if statements in java always.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
atoulme: Yes, otherwise another program located in a different java package cannot check if EVMC is available.

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
atoulme: You would want to package all native libraries in the Java jar. This means the jar contains a .so, a .dylib, and a .dll, at a minimum. You can always let folks place the native lib in the path if they want to load their custom library or use a 32 bit system.

I guess the best way to test on Windows would be to have a CircleCI Windows build, I don't have a Windows machine.

bindings/java/java/src/test/java/org/ethereum/evmc/EvmcTest.java
atoulme: Good catch, fixed.

atoulme:
Can you look at the cmake build failure in Windows? I probably miss something basic.

axic:
@atoulme so the code here and in #555 is identical, right? And this loader is what you prefer?

I am not sure why this was missed, but if you are happy with the logic here and it works with tuweni, I'll clean this PR up and try to get it merged.

bindings/java/java/src/main/java/org/ethereum/evmc/Host.java
axic: I thought we depend on this setting (https://github.com/ethereum/evmc/blob/master/bindings/java/build.gradle#L19-L31) to enforce out coding style.

Is CI not actually enforcing this? Are you using a different version of spotless?

axic: According to https://www.jvt.me/posts/2020/05/15/gradle-spotless/ we have to run `./gradlew clean spotlessApply test`, which we actually do: https://github.com/ethereum/evmc/blob/master/bindings/java/Makefile#L13

axic: Ah, I see we should be running `spotlessCheck` to find issues haha.

axic: Fixed in #581.

axic:
Rebased and pushed this entire branch over #557.

PR #560, opened 2020-11-02
C++: Optimize EVMC types comparison

chfast:
When comparing equal, byteswap is not necessary so don't suggest it to the compiler.

### DIFF
--- a/include/evmc/evmc.hpp
+++ b/include/evmc/evmc.hpp
@@ -53,7 +53,7 @@ struct address : evmc_address
     {}
 
     /// Explicit operator converting to bool.
-    constexpr inline explicit operator bool() const noexcept;
+    inline constexpr explicit operator bool() const noexcept;
 };
 
 /// The fixed size array of 32 bytes for storing 256-bit EVM values.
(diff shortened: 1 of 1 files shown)


include/evmc/evmc.hpp
```diff
@@ -159,10 +159,9 @@ inline constexpr uint64_t fnv1a_by64(uint64_t h, uint64_t x) noexcept
 /// The "equal to" comparison operator for the evmc::address type.
 inline constexpr bool operator==(const address& a, const address& b) noexcept
 {
-    // TODO: Report bug in clang keeping unnecessary bswap.
-    return load64be(&a.bytes[0]) == load64be(&b.bytes[0]) &&
-           load64be(&a.bytes[8]) == load64be(&b.bytes[8]) &&
-           load32be(&a.bytes[16]) == load32be(&b.bytes[16]);
+    return load64le(&a.bytes[0]) == load64le(&b.bytes[0]) &&
```
axic, line 162: Why is this faster? Is the `load32le` helper recognized to be a simple `mov`/`lea`?

chfast: This omits `bswap`.

axic: That's what I asked "recognized to be a simple mov".

axic (approved):

yperbasis (approved):

gumb0 (approved):

include/evmc/evmc.hpp
chfast: Yes. The `load64le` is rather `mov` instead of `mov` + `bswap` or `movbe`.

chfast: However, GCC can recognize that `bswap(a) == bswap(b)` is actually `a == b`. So this only helps Clang.

PR #562, opened 2020-11-02
Testable tool commands

chfast:
This move the implementation of "run" command from the `evmc` tool to separate library. This allows provide unit tests for this command.

### DIFF
--- a/include/evmc/evmc.hpp
+++ b/include/evmc/evmc.hpp
@@ -113,34 +113,34 @@ struct bytes32 : evmc_bytes32
 using uint256be = bytes32;
 
 
-/// Loads 64 bits / 8 bytes of data from the given @p bytes array in big-endian order.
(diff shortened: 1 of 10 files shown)


include/evmc/evmc.hpp
```diff
@@ -114,18 +114,18 @@ using uint256be = bytes32;
 
 
 /// Loads 64 bits / 8 bytes of data from the given @p bytes array in big-endian order.
-constexpr inline uint64_t load64be(const uint8_t* bytes) noexcept
+constexpr inline uint64_t load64be(const uint8_t* bs) noexcept
```
axic, line 117: Why not `data`? Also the comment still says `bytes`.

axic (approved):
Looks okay to me apart the naming question.

include/evmc/evmc.hpp
chfast: Changed to `data`.

PR #564, opened 2020-11-03
tools: Add --input option to `evmc run`

chfast:


tools/evmc/main.cpp
```diff
@@ -17,12 +17,14 @@ int main(int argc, const char** argv)
...
     int64_t gas = 1000000;
     auto rev = EVMC_ISTANBUL;
+    std::string input_hex;
 
     auto& run_cmd = *app.add_subcommand("run", "Execute EVM bytecode");
     run_cmd.add_option("code", code_hex, "Hex-encoded bytecode")->required();
     run_cmd.add_option("--vm", vm_config, "EVMC VM module")->required()->envname("EVMC_VM");
     run_cmd.add_option("--gas", gas, "Execution gas limit", true)->check(CLI::Range(0, 1000000000));
     run_cmd.add_option("--rev", rev, "EVM revision", true);
+    run_cmd.add_option("--input", input_hex, "Hex-encoded input bytes");
```
axic, line 27: I know it is common to name this `input`, but I think technically `--calldata` is a better choice as it reflects what it does.

axic: I'd go as far to propose we should rename it in the EVMC interface in with ABIv8.

chfast: I think this is fine to rename it, but I see two issues:
1. There is execution "output".
2. In EVMC this will be `msg.calldata_data`.

gumb0 (approved):

PR #573, opened 2021-03-08
Enhance hex utils

chfast:


test/unittests/utils_test.cpp
```diff
@@ -0,0 +1,49 @@
...
+    EXPECT_EQ(hex(0x55), "55");
+    EXPECT_EQ(hex(0x7f), "7f");
+    EXPECT_EQ(hex(0x80), "80");
+    EXPECT_EQ(hex(0xff), "ff");
+}
+
+TEST(utils, hex_of_bytes)
+{
+    uint8_t data[] = {0x00, 0x01, 0x00, 0xa0, 0xff, 0x00, 0x55, 0x00};
+    EXPECT_EQ(hex(data, sizeof(data)), "000100a0ff005500");
```
axic, line 23:
```suggestion
    EXPECT_EQ(hex(data, 0), "");
    EXPECT_EQ(hex(nullptr, 0), ""); // is this working?
```

tools/utils/utils.cpp
```diff
@@ -8,32 +8,43 @@
 
 namespace evmc
 {
-bytes from_hex(const std::string& hex)
+namespace
+{
+inline int from_hex_digit(char h)
```
axic, line 13: Can't you just copy all this from fizzy? Or is it a copy from fizzy?

test/unittests/tool_commands_test.cpp
```diff
@@ -36,6 +36,16 @@ TEST(tool_commands, run_empty_code)
     EXPECT_EQ(out.str(), out_pattern("Frontier", 1, "success", 0, ""));
 }
 
+TEST(tool_commands, run_oog)
```
axic, line 39: Is this really part of the "allow optional 0x prefix" commit?

axic (approved):
The comments were just about potential improvements, since the CLI exposed part seems to be fine it would okay to merge without addressing those.

test/unittests/tool_commands_test.cpp
chfast: Yes, because I use 0x in this test :) I also triggered some previously uncovered lines BTW.

tools/utils/utils.cpp
chfast: Fizzy has a copy from evmc, except ported to C++17. This is a new thing.

tools/utils/utils.cpp
```diff
@@ -8,32 +8,62 @@
...
+    else
+        throw std::out_of_range{"not a hex digit"};
+}
+
+template <typename OutputIt>
+inline void from_hex(const char* hex, size_t size, OutputIt result)
 {
-    if (hex.length() % 2 == 1)
+    // Omit the optional 0x prefix.
+    if (size >= 2 && hex[0] == '0' && hex[1] == 'x')
```
gumb0, line 29: You could do
```
size_t i = ((size >= 2 && hex[0] == '0' && hex[1] == 'x') ? 2 : 0);
```
and still pass `hex` as `const std::string&`

chfast: Done.

chfast:
Replaced with https://github.com/ethereum/evmc/pull/575.

Issue #569, opened 2021-01-16
Add stdin (or file input) support to evmc run


axic:
Currently it can be worked around with
```
evmc run `cat file`
```

It would be nice to perhaps support `-` as a well known indicator of "look at stdin". Alternatively just recognise that argument is not a hex string but a filename and try loading it.

chfast:
With #574 file names are now recognized for `code` and `--input` arguments.

PR #574, opened 2021-03-09
evmc tool CLI improvements

chfast:
For arguments `code` and `--input` we try to guess if the value is a file path or hex string. If it is a file then we load hex string from the file.

There are some gaps here still: we assume the file contains valid hex string. We could additionally inspect the file content but that requires custom `ExistingFile` validator.

Closes https://github.com/ethereum/evmc/issues/569.


tools/evmc/main.cpp
```diff
@@ -3,29 +3,84 @@
...
+        return evmc::from_hex(str);
+
+    // Must be a file path.
+    std::ifstream file{str, std::ios::in | std::ios::binary};
+    const auto content =
+        std::string(std::istreambuf_iterator<char>{file}, std::istreambuf_iterator<char>{});
+    return evmc::from_hex(content);
+}
+
+struct HexValidator : public CLI::Validator
```
axic, line 38: Nice!

tools/evmc/main.cpp
```diff
@@ -3,29 +3,84 @@
...
+    auto create = false;
 
     CLI::App app{"EVMC tool"};
     const auto& version_flag = *app.add_flag("--version", "Print version information and exit");
     const auto& vm_option =
         *app.add_option("--vm", vm_config, "EVMC VM module")->envname("EVMC_VM");
 
     auto& run_cmd = *app.add_subcommand("run", "Execute EVM bytecode");
-    run_cmd.add_option("code", code_hex, "Hex-encoded bytecode")->required();
+    run_cmd.add_option("code", code_hex, "Bytecode")->required()->check(Hex | CLI::ExistingFile);
```
axic, line 77: Cool!

tools/evmc/main.cpp
```diff
@@ -23,10 +43,10 @@ int main(int argc, const char** argv)
         *app.add_option("--vm", vm_config, "EVMC VM module")->envname("EVMC_VM");
 
     auto& run_cmd = *app.add_subcommand("run", "Execute EVM bytecode");
-    run_cmd.add_option("code", code_hex, "Hex-encoded bytecode")->required();
+    run_cmd.add_option("code", code_hex, "Bytecode")->required()->check(Hex);
```
axic: So this replaces `required`, does that mean HexValidator should fail if the input is shorter than 2 nibbles?

axic: This also raises the question, is there a CLI test for calling this without input bytecode?

tools/evmc/main.cpp
```diff
@@ -6,9 +6,24 @@
 #include <CLI/CLI.hpp>
 #include <evmc/hex.hpp>
 #include <evmc/loader.h>
+#include <fstream>
 
 namespace
 {
+/// Returns the input str if already valid hex string. Otherwise, interprets the str as a file
+/// name and loads the hex string from this file.
+/// @todo The file content is not validated.
```
axic, line 15: So does the file needs to be a hex string, or it is a binary file?

tools/evmc/main.cpp
chfast: The "required" mean the argument is required. Values are always validated. Any order works.

tools/evmc/main.cpp
chfast: hex string

tools/evmc/main.cpp
axic: Oh, for some reason the split diff didn't show it to be already chained. So just ignore this question then.

tools/evmc/main.cpp
axic: Can it be validated? Even if just validating it here with `abort`, I know it is not nice, but reduces confusion should someone misuse it.

chfast: It still throws "invalid hex digit" exception or similar. This is caught and reported although it is not clear where the error comes from. I added test for this case.

axic: Isn't it coming from somewhere inside `cmd::run`?

chfast: It is. But I mean user will see the error about invalid hex without information about the source. He has to guess it is code and/or input what is invalid.

axic: I think it comes from there as it uses `load_hex` which throws the error and `main` catches that. Can you leave this as a comment here (that it is caught later in `evmc::run`), for now?

test/tools/CMakeLists.txt
```diff
@@ -39,5 +39,35 @@ add_evmc_tool_test(
...
 
+add_evmc_tool_test(
+    invalid_hex_code
+    "--vm $<TARGET_FILE:evmc::example-vm> run 0x600"
+    "code: \\(incomplete hex byte pair\\) OR \\(File does not exist: 0x600\\)"
+)
+
+add_evmc_tool_test(
+    invalid_hex_input
+    "--vm $<TARGET_FILE:evmc::example-vm> run 0x --input aa0y"
```
axic, line 50: Hmm, we accept this weird Ethereum concept of `0x` for empty input? Does it accept `run ''` or `run ""` ?

chfast: All should work

axic: Could you add one more test with that? I hope we can remove the `run 0x` case at some point in the future :)

chfast: https://github.com/ethereum/evmc/blob/master/test/unittests/hex_test.cpp#L71

test/tools/CMakeLists.txt
```diff
@@ -39,5 +39,35 @@ add_evmc_tool_test(
     "Result: +success[\r\n]+Gas used: +6[\r\n]+Output: +02[\r\n]"
 )
 
+add_evmc_tool_test(
+    invalid_hex_code
+    "--vm $<TARGET_FILE:evmc::example-vm> run 0x600"
+    "code: \\(incomplete hex byte pair\\) OR \\(File does not exist: 0x600\\)"
```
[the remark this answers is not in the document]
chfast: This is CLI11 default but probably can be changed if you re-implement many things. I'm not going to do it now.

gumb0 (approved):

test/tools/CMakeLists.txt
axic: I meant adding a test for `run ""`.

And yes, I know it currently accepts `0x` as empty and not asking to remove it right now :)

chfast: I tried 30 mins, but I don't know how to do this in CMake.

chfast: Added.

axic: Hmm, I thought `"--vm $<TARGET_FILE:evmc::example-vm> run '' --input aa0y"` would have been enough?

axic (approved):

PR #575, opened 2021-03-15
Create separate evmc::hex library for hex encoding/decoding

chfast:


lib/hex/hex.cpp
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    if (b != empty_byte_mark)
+        throw hex_error{hex_errc::incomplete_hex_byte_pair};
+}
+}  // namespace
+
+std::error_code make_error_code(hex_errc errc) noexcept
+{
+    static hex_error_category category;
```
gumb0, line 78:
```suggestion
    static const hex_error_category category;
```

gumb0:
I'm reading intro to `std::error_code` at http://blog.think-async.com/2010/04/system-error-support-in-c0x-part-4.html and I think public function `const std::error_category& hex_category()` is going to be needed, for the user to be able to check `e.code().category() == evmc::hex_category()`.

include/evmc/hex.hpp
```diff
@@ -0,0 +1,60 @@
...
+    return {hex_chars[b >> 4], hex_chars[b & 0xf]};
+}
+
+/// Validates hex encoded string.
+std::error_code validate_hex(const std::string& hex) noexcept;
+
+/// Decodes hex encoded string to bytes.
+///
+/// Exceptions:
+/// - std::length_error when the input has invalid length (must be even).
```
gumb0, line 47: Outdated.

chfast: Fixed.

include/evmc/hex.hpp
```diff
@@ -0,0 +1,60 @@
...
+bytes from_hex(const std::string& hex);
+
+/// Encodes bytes as hex string.
+std::string hex(const uint8_t* data, size_t size);
+}  // namespace evmc
+
+namespace std
+{
+template <>
+struct is_error_code_enum<evmc::hex_errc> : true_type
```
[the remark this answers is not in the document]
chfast: Done.

lib/hex/hex.cpp
chfast: Move to `hex_category()` with `static const`.

lib/hex/hex.cpp
```diff
@@ -0,0 +1,119 @@
...
+    // Omit the optional 0x prefix.
+    const auto hex_begin =
+        (hex.size() >= 2 && hex[0] == '0' && hex[1] == 'x') ? hex.begin() + 2 : hex.begin();
+
+    constexpr int empty_byte_mark = -1;
+    int b = empty_byte_mark;
+    for (auto it = hex_begin; it != hex.end(); ++it)
+    {
+        const auto h = *it;
+        if (std::isspace(h))
```
gumb0, line 38: Hm why do you want to allow whitespace? If this is to skip extra whitespace from command line, it's better to handle it in command line code.

Also currently it looks to allow `"0x12"` and `"  12"` (also `"1 2"`), but not `"   0x12"`

chfast: Yes, this is what I want for simplicity. This is useful for loading hex from files to get rid of line breaks.

gumb0: Just `hex.erase(std::remove_if(hex.begin(), hex.end()), std::isspace), hex.end())` them!

lib/hex/hex.cpp
```diff
@@ -0,0 +1,122 @@
...
+            return "unknown error";
+        }
+    }
+};
+}  // namespace
+
+const std::error_category& hex_category() noexcept
+{
+    // Create static category object. This involves mutex-protected dynamic initialization.
+    // Because of the C++ defect 253, the {} syntax is used.
```
gumb0, line 79: "valarray helper functions are almost entirely useless" ? http://www.open-std.org/jtc1/sc22/wg21/docs/lwg-defects.html#253

chfast: http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#253

lib/hex/hex.cpp
chfast: This will not work with input iterators (planned change). In the end we can have `hex_decoding_iterator` and `strip_whitespace_iterator` but I think having just one is simpler.

PR #579, opened 2021-03-22
java: EvmcVm.close does not throw exception

axic:
Pulled out of #555.


atoulme (approved):
sure, important if implementing Closeable

chfast (approved):

PR #594, opened 2021-04-30
mocked_host: create account in set_storage if non-existent

axic:


axic (commented):
Looks good to me. We should extend `cmd::run` with an ether transfer option and have tests that a) account is preserved with balance on create; b) on balance + storage setting.

This can be done in a further PR since it requires more changes.

chfast (approved):

Issue #175, opened 2019-01-14
Rename EVMC_LATEST_REVISION to EVMC_MAX_REVISION


axic:
See https://github.com/ethereum/evmc/pull/174#discussion_r247493343

chfast:
Closed by #184 .

PR #595, opened 2021-05-02
Add EVMC_LONDON revision

yperbasis:
Add `evmc_revision` value for the upcoming [London fork](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/mainnet-upgrades/london.md).

### DIFF
--- a/bindings/go/evmc/evmc.go
+++ b/bindings/go/evmc/evmc.go
@@ -147,6 +147,8 @@ const (
 	Constantinople   Revision = C.EVMC_CONSTANTINOPLE
 	Petersburg       Revision = C.EVMC_PETERSBURG
 	Istanbul         Revision = C.EVMC_ISTANBUL
+	Berlin           Revision = C.EVMC_BERLIN
+	London           Revision = C.EVMC_LONDON
 )
 
 type VM struct {
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -813,19 +813,26 @@ enum evmc_revision
     /**
      * The Istanbul revision.
      *
-     * The spec draft: https://eips.ethereum.org/EIPS/eip-1679.
+     * https://eips.ethereum.org/EIPS/eip-1679
      */
     EVMC_ISTANBUL = 7,
 
     /**
      * The Berlin revision.
      *
-     * The spec draft: https://eips.ethereum.org/EIPS/eip-2070.
+     * https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/mainnet-upgrades/berlin.md
      */
     EVMC_BERLIN = 8,
 
+    /**
+     * The London revision.
+     *
+     * https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/mainnet-upgrades/london.md
+     */
+    EVMC_LONDON = 9,
+
     /** The maximum EVM revision supported. */
-    EVMC_MAX_REVISION = EVMC_BERLIN
+    EVMC_MAX_REVISION = EVMC_LONDON
 };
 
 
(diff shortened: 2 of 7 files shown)


tools/evmc/main.cpp
```diff
@@ -48,7 +48,7 @@ int main(int argc, const char** argv)
     std::string vm_config;
     std::string code_arg;
     int64_t gas = 1000000;
-    auto rev = EVMC_ISTANBUL;
+    auto rev = EVMC_MAX_REVISION;
```
chfast, line 51: Here we rather want to go with stable Berlin for now. If we want to minimize number of changes here, we may want to introduce `EVMC_LATEST_STABLE_REVISON` (something we considered before). Comments @axic ?

include/evmc/evmc.h
```diff
@@ -813,19 +813,26 @@ enum evmc_revision
...
-     * The spec draft: https://eips.ethereum.org/EIPS/eip-1679.
+     * https://eips.ethereum.org/EIPS/eip-1679
      */
     EVMC_ISTANBUL = 7,
 
     /**
      * The Berlin revision.
      *
-     * The spec draft: https://eips.ethereum.org/EIPS/eip-2070.
+     * https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/mainnet-upgrades/berlin.md
```
axic, line 823: Still don't get why the EIPs are deprecated and now there's a volatile place for the spec, i.e. this link has changed twice already since it has been added to the EIP 😬

test/unittests/instructions_test.cpp
```diff
@@ -327,3 +327,17 @@ TEST(instructions, berlin_hard_fork)
...
 }
+
+TEST(instructions, london_hard_fork)
+{
+    const auto l = evmc_get_instruction_metrics_table(EVMC_LONDON);
+    const auto b = evmc_get_instruction_metrics_table(EVMC_BERLIN);
+    const auto ln = evmc_get_instruction_names_table(EVMC_LONDON);
+    const auto bn = evmc_get_instruction_names_table(EVMC_BERLIN);
+
+    for (int op{OP_STOP}; op <= OP_SELFDESTRUCT; ++op)
```
axic, line 338: I know this isn't introduced by this PR as a new thing, but why is this going from STOP->SELFDESTRUCT and not `0` to `255` or some `min`/`max` based on the enum type? I know it is identical but still feels weird basing it on the instruction name.

tools/evmc/main.cpp
axic: Took a bit of digging, but this was the issue: #175. And this was discussion sparking it: https://github.com/ethereum/evmc/pull/174#discussion_r245318043

I think what we discussed there still stands.

include/evmc/evmc.h
chfast: This looks bad to me too. I think we should ask for shorter permanent links.

test/unittests/instructions_test.cpp
chfast: Make sense, I will change that later.

chfast (approved):

bindings/go/evmc/evmc.go
```diff
@@ -147,6 +147,8 @@ const (
 	Constantinople   Revision = C.EVMC_CONSTANTINOPLE
 	Petersburg       Revision = C.EVMC_PETERSBURG
 	Istanbul         Revision = C.EVMC_ISTANBUL
+	Berlin           Revision = C.EVMC_BERLIN
```
axic, line 150: Hmm, isn't this a bugfix which should be released as 8.0.1?

chfast: It is, but we don't need to fix it proactively.

gumb0 (approved):

PR #596, opened 2021-05-04
cpp: Add support for _uin256be literals

axic:
Only extending the literal tests, because the other tests are covered
via bytes32 tests given uint256be is an alias.

### DIFF
--- a/include/evmc/evmc.hpp
+++ b/include/evmc/evmc.hpp
@@ -332,6 +332,13 @@ constexpr bytes32 operator""_bytes32() noexcept
 {
     return internal::from_literal<bytes32, c...>();
 }
+
+/// Literal for evmc::uint256be.
+template <char... c>
+constexpr uint256be operator""_uint256be() noexcept
+{
+    return internal::from_literal<uint256be, c...>();
+}
 }  // namespace literals
 
 using namespace literals;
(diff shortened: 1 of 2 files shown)


chfast (approved):
Are you going to use it?

chfast:
What will be confusing here is that these literals are required to have exactly 64 hex digits.

axic:
> Are you going to use it?

I was adding `--value` support to the evmc tool, that's how I found this, but in the end it is actually not that important.

> What will be confusing here is that these literals are required to have exactly 64 hex digits.

Yeah I was wondering if it would make sense make this work on decimal instead. Probably better holding the merge back a bit.

chfast:
> Yeah I was wondering if it would make sense make this work on decimal instead. Probably better holding the merge back a bit.

Unless there exist some simplified algorithm for decimal literals I'm not aware of, this requires implementing division by 10. https://github.com/chfast/intx/blob/master/include/intx/int128.hpp#L924

axic:
@chfast leaning towards that even with base 16 this may be useful

axic:
Rebased. I think this is still useful for unit tests, based on `ethereum/tests`. I think I needed/wanted it for soltests at the time.

PR #599, opened 2021-05-20
Move output stream operators to evmc::evmc_cpp

chfast:
This removes internal `evmc::tool-utils` library which only implements 2 `operator<<` for EVMC types and moves these functions to C and C++ public API. This simplifies dependencies and allows to use these functions in EVM tracing in evmone.


axic (approved):

axic (approved):
Nice that it is documented now.

PR #603, opened 2021-05-26
Support London EIP-3198: BASEFEE opcode

yperbasis:
Necessary facilities for [EIP-3198](https://eips.ethereum.org/EIPS/eip-3198), which is included into [London hard fork](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/mainnet-upgrades/london.md).


bindings/go/evmc/host.go
```diff
@@ -76,6 +76,7 @@ type TxContext struct {
 	GasLimit   int64
 	Difficulty Hash
 	ChainID    Hash
+	BaseFee    Hash // EIP-1559, EIP-3198
```
axic, line 79: Do we need to keep these comments here? If so, perhaps for the older ones (such as ChainID) too?

chfast (approved):

PR #605, opened 2021-06-28
Add EVMC_LATEST_STABLE_REVISON

chfast:
```c
    /**
     * The latest known EVM revision with finalized specification.
     *
     * This is handy for EVM tools to always use the latest revision available.
     */
    EVMC_LATEST_STABLE_REVISION = EVMC_LONDON
```

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -840,7 +840,14 @@ enum evmc_revision
     EVMC_SHANGHAI = 10,
 
     /** The maximum EVM revision supported. */
-    EVMC_MAX_REVISION = EVMC_SHANGHAI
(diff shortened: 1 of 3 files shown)


axic (approved):

yperbasis (approved):
I'd add something like
```
_Static_assert(EVMC_LATEST_STABLE_REVISION <= EVMC_MAX_REVISION);
```

chfast:
> I'd add something like
> 
> ```
> _Static_assert(EVMC_LATEST_STABLE_REVISION <= EVMC_MAX_REVISION);
> ```

Make sense. Added in #606 but in C++. I don't want to place it in C as that requires C11.

PR #608, opened 2021-06-29
go: Add missing Shanghai revision constant

chfast:
Added a test that fails in case of `C.EVMC_MAX_REVISION` is increased as a remainder to add new Revision constants to Go.
> evmc_test.go:65: missing constant for revision 10

The missing London Revision is added.

### DIFF
--- a/bindings/go/evmc/evmc.go
+++ b/bindings/go/evmc/evmc.go
@@ -139,16 +139,18 @@ const (
 type Revision int32
 
 const (
-	Frontier         Revision = C.EVMC_FRONTIER
-	Homestead        Revision = C.EVMC_HOMESTEAD
-	TangerineWhistle Revision = C.EVMC_TANGERINE_WHISTLE
-	SpuriousDragon   Revision = C.EVMC_SPURIOUS_DRAGON
-	Byzantium        Revision = C.EVMC_BYZANTIUM
-	Constantinople   Revision = C.EVMC_CONSTANTINOPLE
-	Petersburg       Revision = C.EVMC_PETERSBURG
-	Istanbul         Revision = C.EVMC_ISTANBUL
-	Berlin           Revision = C.EVMC_BERLIN
-	London           Revision = C.EVMC_LONDON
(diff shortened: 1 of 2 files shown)


bindings/go/evmc/evmc.go
```diff
@@ -149,6 +149,8 @@ const (
 	Istanbul         Revision = C.EVMC_ISTANBUL
 	Berlin           Revision = C.EVMC_BERLIN
 	London           Revision = C.EVMC_LONDON
+	Shanghai         Revision = C.EVMC_SHANGHAI
+	maxRevision      Revision = C.EVMC_MAX_REVISION
```
axic, line 153: Why not also add latest stable?

chfast: `maxRevision` is private and cannot be used outside of the `evmc` package.

axic: How is it private? Because lowercase?

bindings/go/evmc/evmc.go
```diff
@@ -149,6 +149,8 @@ const (
...
 )
```
axic, line 155: Unrelated, but can't comment on the line: can `func (err Error) Error() string` be replaced by the new string helpers?

bindings/go/evmc/evmc.go
chfast: Yes. But made it public now together with `LatestStableRevision`. I don't want to overthink it.

bindings/go/evmc/evmc_test.go
```diff
@@ -59,3 +59,9 @@ func TestExecuteEmptyCode(t *testing.T) {
 		t.Errorf("execution returned unexpected error: %v", err)
 	}
 }
+
+func TestRevision(t *testing.T) {
+	if MaxRevision != London {
```
axic: Should there also be a test for `LatestStableRevision` here?

chfast: Added.

Issue #335, opened 2019-06-27
Add information about current address for DELEGATE call


chfast:
In the DELEGATECALL (and legacy CALLCODE) there are 3 addresses involved:
- the destination address where the code is (`msg.destination`),
- the current address where execution should happen,
- the sender address (`msg.sender`).

The current address is not provided in the `msg` so the Host has to store it on its own.

jlokier:
Related, I don't see where it is documented that  `msg.destination` passed to `call` must be replaced by a different value in the host before it passes the (copied) message on to `execute`, when doing `DELEGATECALL` or `CALLCODE`.  (In `call`, the host needs to know the code address from the VM, but in `execute` the child VM needs to know the account address for `get_storage` etc.)

It would be helpful if the comment for `evmc_message::destination` in `evmc.h` mentioned this dual role.  As it is, I had to read some implementions to confirm this is wha the host is expected to do.

Because the host has to keep track of the current account address for `DELEGATECALL` anyway, this makes the address parameter in `get_storage`, `set_storage`, `selfdestruct` and `access_storage` redundant.  At least it's cheap to pass, being a pointer.

Knowing the only permitted value supports an optimisation in the host: It can ignore the address parameter and keep an account object handy instead of looking it up on each call.

chfast:
Done in #611.

PR #611, opened 2021-07-13
Add code_address to evmc_message

yperbasis:
Without code address `evmc_message` is not enough to implement `evmc_call_fn` for the DELEGATECALL case. For more detail please see https://github.com/torquem-ch/silkworm/pull/286.

**_Warning:_ This is an API breaking change and requires a major version bump!**

Related PRs: https://github.com/ethereum/evmone/pull/360 and https://github.com/torquem-ch/silkworm/pull/288.

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -109,7 +109,7 @@ struct evmc_message
     /** The amount of gas for message execution. */
     int64_t gas;
 
-    /** The destination of the message. */
+    /** The destination (recipient) of the message. */
     evmc_address destination;
 
     /** The sender of the message. */
@@ -141,6 +141,18 @@ struct evmc_message
      * Ignored in evmc_execute_fn().
      */
     evmc_bytes32 create2_salt;
+
+    /**
+     * The address of the code to be executed.
+     *
+     * May be different from the evmc_message::destination (recipient) in case of
+     * ::EVMC_CALLCODE or ::EVMC_DELEGATECALL.
+     * See Section 8 "Message Call" of the Yellow Paper for detail.
+     *
+     * Not required when invoking evmc_execute_fn(), only when invoking evmc_call_fn().
+     * Ignored if kind is ::EVMC_CREATE or ::EVMC_CREATE2.
+     */
+    evmc_address code_address;
 };
 
 
--- a/bindings/go/evmc/evmc.go
+++ b/bindings/go/evmc/evmc.go
@@ -43,6 +43,7 @@ static struct evmc_result execute_wrapper(struct evmc_vm* vm,
 		input_size,
 		*value,
 		{{0}}, // create2_salt: not required for execution
+		{{0}}, // code_address: not required for execution
 	};
 
 	struct evmc_host_context* context = (struct evmc_host_context*)context_index;
--- a/bindings/go/evmc/host.go
+++ b/bindings/go/evmc/host.go
@@ -93,7 +93,7 @@ type HostContext interface {
 	EmitLog(addr Address, topics []Hash, data []byte)
 	Call(kind CallKind,
 		destination Address, sender Address, value Hash, input []byte, gas int64, depth int,
-		static bool, salt Hash) (output []byte, gasLeft int64, createAddr Address, err error)
+		static bool, salt Hash, codeAddress Address) (output []byte, gasLeft int64, createAddr Address, err error)
 	AccessAccount(addr Address) AccessStatus
 	AccessStorage(addr Address, key Hash) AccessStatus
 }
@@ -208,7 +208,8 @@ func call(pCtx unsafe.Pointer, msg *C.struct_evmc_message) C.struct_evmc_result 
 
 	kind := CallKind(msg.kind)
 	output, gasLeft, createAddr, err := ctx.Call(kind, goAddress(msg.destination), goAddress(msg.sender), goHash(msg.value),
-		goByteSlice(msg.input_data, msg.input_size), int64(msg.gas), int(msg.depth), msg.flags != 0, goHash(msg.create2_salt))
+		goByteSlice(msg.input_data, msg.input_size), int64(msg.gas), int(msg.depth), msg.flags != 0, goHash(msg.create2_salt),
+		goAddress(msg.code_address))
 
 	statusCode := C.enum_evmc_status_code(0)
 	if err != nil {
--- a/bindings/go/evmc/host_test.go
+++ b/bindings/go/evmc/host_test.go
@@ -57,7 +57,7 @@ func (host *testHostContext) EmitLog(addr Address, topics []Hash, data []byte) {
 
 func (host *testHostContext) Call(kind CallKind,
 	destination Address, sender Address, value Hash, input []byte, gas int64, depth int,
-	static bool, salt Hash) (output []byte, gasLeft int64, createAddr Address, err error) {
+	static bool, salt Hash, codeAddress Address) (output []byte, gasLeft int64, createAddr Address, err error) {
 	output = []byte("output from testHostContext.Call()")
 	return output, gas, Address{}, nil
 }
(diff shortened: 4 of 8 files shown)


chfast:
What was the problem placing it next to `destination` and `sander`?

yperbasis:
> What was the problem placing it next to `destination` and `sander`?

I was thinking about better backwards compatibility. But then again, `code_address` is mandatory for correct calls, and logically it belongs next to `destination` and `sender`. I can put it back there once PR #612 is merged.

yperbasis:
@chfast Then again, one argument for adding `code_address` at the end of `evmc_message` rather than next to  `destination` is that it's only required for `evmc_call_fn`, not `evmc_execute_fn`.

Also, an alternative is to remove `code_address` & `create2_salt` from `evmc_message` and instead pass them as arguments to `evmc_call_fn`.

chfast:
> Also, an alternative is to remove `code_address` & `create2_salt` from `evmc_message` and instead pass them as arguments to `evmc_call_fn`.

This is an indicator that using the same `evmc_message` for both execution and calls is not good idea. We may see this problem from this perspective and plan something for the future.

I will review this PR soon.

chfast (approved):

gumb0 (approved):

jlokier:
A related issue that this maybe closes is #335.

include/evmc/evmc.h
```diff
@@ -141,6 +141,18 @@ struct evmc_message
      * Ignored in evmc_execute_fn().
      */
     evmc_bytes32 create2_salt;
+
+    /**
+     * The address of the code to be executed.
+     *
+     * May be different from the evmc_message::destination (recipient) in case of
+     * ::EVMC_CALLCODE or ::EVMC_DELEGATECALL.
```
jlokier, line 149: @yperbasis, When working on EVMC compatibility in [Nimbus-eth1](https://github.com/status-im/nimbus-eth1), I was surprised to find that correct handling of `msg.destination` received via `evmc_call_fn` required checking whether `msg.destination` corresponds to a valid precompile in the current block's fork.  This is because `DELEGATECALL` and `CALLCODE` lose their special behaviour when calling a precompile.

Perhaps the EVMC API comment should mention that it depends on whether the destination is a precompile?

chfast: I don't believe this is the case. What is so special about precompiles in this context?

yperbasis: Precompiles are handled on the Host side, not VM side (in EVMC terms). After this change Hosts should use `msg.code_address` rather than `msg.destination` to determine which smart contract or precompile to execute. On the VM side DELEGATECALL and CALLCODE shouldn't have any special behaviour depending on whether the code address is a precompile or not. At least in EVMC terms, where VM is not responsible for things like maintaining EIP-2929 `accessed_addresses`.

For example, that's how the Host side looks in Silkworm after this change: https://github.com/torquem-ch/silkworm/blob/master/core/silkworm/execution/evm.cpp#L156

jlokier: @chfast, before this change,`msg.destination` given to `evmc_call_fn` always contained the address for looking up the contract bytecode for all types of call (obviously not for precompiles), and yes of course that's a host side operation.

Then for `DELEGATECALL` and `CALLCODE`, `msg.destination` had to be replaced with the recipient address before calling `evmc_execute_fn`.

Silkworm calculated the recipient [here](https://github.com/torquem-ch/silkworm/blob/27a26b062652af38cd35c77ba3c86ed8077d01e4/core/silkworm/execution/evm.cpp#L227), and then overwrotes `msg.destination` [here](https://github.com/torquem-ch/silkworm/blob/27a26b062652af38cd35c77ba3c86ed8077d01e4/core/silkworm/execution/evm.cpp#L212) *conditional on the `evmc_execute_fn` not being a precompile call*.

In Nimbus I simplified that to just overwrite `msg.destination` with the equivalent of `address_stack_.top()` if doing `DELEGATECALL` or `CALLCODE` (to a non-precompile) before calling `evmc_execute_fn`, which works out the same as Silkworm (because it is equal to `msg.sender` in the `CALLCODE` case).

Now, after this change, `msg.destination` no longer needs overwriting by the host, so the meaning of `msg.destination` is *no longer* dependent on whether it's a fork-specific precompile address.

So I retract my suggestion because I'd mistakenly thought the only change was adding "missing but calculatable" field `msg.code_address`.  I hadn't realised the meaning of `msg.destination` has also been changed.

ps. It is neat that `msg` can now be passed along unchanged by nested calls Until now I thought it had been an intentional design hack to overload `msg.destination` in this clever way to save a field. That proved quite handy to make me think harder about the information flows in Nimbus and how the EVM isn't allowed to know the code address (ironically, in practice it will now be passed the code address, though it's not supposed to use it).

jlokier: (I've edited my comment above; if you read the old version and ignored it, please read again, there remains an unresolved issue with regard to the new `msg.code_address` change and `EVMC_CAPABILITY_PRECOMPILES`.

jlokier: @yperbasis,
> Precompiles are handled on the Host side, not VM side (in EVMC terms).

Nimbus has called precompiles via EVMC since it first supported any kind of EVMC a couple of years ago, and the flag`EVMC_CAPABILITY_PRECOMPILES` was added since then, but we didn't bother using it until recently.

If precompiles are _required_ to be handled on the host side, that's surprising news.

chfast: Yes, I can see the problem now.
I need to track Host implementations if they change `msg.destination` on the way. Even if they do, that's probably not recommended thing to do and EVMC precompiles implementation should just only consider `msg.code_address` for the precompile selection. Is this a problematic change?

What I plan to do as a follow up is to clarify the documentation and change EVMC examples. Might be also nice to change `destination` to `recipient`. How about we discuss the problem there?

jlokier: I don't mind where the discussion takes place (though I tend to miss some GitHub notications without a poke).

Documentation on the flow from EVM caller to nested EVM callee would be a nice improvement.  It might be enough just to expand the comments a bit.  There were quite a few things I had to figure out.  I don't need that documentation any more, but it would make EVMC clearer for the next perosn.

I think all host implementations must have changed `msg.destination` (before `msg.code_address` was added), for regular, non-precompiles, to get the correct answers from `DELEGATECALL` and `CALLCODE`.  And since `msg.code_address` they must stop changing `msg.destination` to get the correct answers. So perhaps there's no need to track the implementations. If they run consensus tests, they'll notice.

That leaves only precompiles. I don't know if any implementations call precompiles via EVMC, or if any loadable EVM provides precompiles (please let me know if you find one). Google returns no interesting matches for `EVMC_CAPABILITY_PRECOMPILES`. Nimbus is on track to use it, though.

This change isn't a problem. As EVMC precompile calls are Nimbus to Nimbus only as far as I know, no problem changing it to `msg.code_address`. It would be good to have a clear resolution on what the API-correct thing is if only for "someday" compatibility with others. If nothing is really correct, we can switch to a vendor extension if that helps. The comment at `EVMC_CAPABILITY_PRECOMPILES` about EIP-1352 is wrong anyway, as the host must use the fork-specific threshold due to some consensus tests using the low address range.

jlokier: > If they run consensus tests, they'll notice.

No, wait, who am I kidding. If both the host and EVM are doing the same wrong thing with `msg.code_address` added, they won't notice, and I'd guess cross-testing between different hosts and EVMs using the full consensus suite is not done much.

yperbasis: > Nimbus has called precompiles via EVMC since it first supported any kind of EVMC a couple of years ago, and the flag`EVMC_CAPABILITY_PRECOMPILES` was added since then, but we didn't bother using it until recently.
> 
> If precompiles are _required_ to be handled on the host side, that's surprising news.

My bad, I forgot about `EVMC_CAPABILITY_PRECOMPILES`.

chfast: See https://github.com/ethereum/evmc/pull/615.

PR #612, opened 2021-07-16
go: create2_salt is not required for execution

chfast:
Pulled from #611.

### DIFF
--- a/bindings/go/evmc/evmc.go
+++ b/bindings/go/evmc/evmc.go
@@ -30,7 +30,7 @@ static struct evmc_result execute_wrapper(struct evmc_vm* vm,
 	enum evmc_call_kind kind, uint32_t flags, int32_t depth, int64_t gas,
 	const evmc_address* destination, const evmc_address* sender,
 	const uint8_t* input_data, size_t input_size, const evmc_uint256be* value,
-	const uint8_t* code, size_t code_size, const evmc_bytes32* create2_salt)
+	const uint8_t* code, size_t code_size)
 {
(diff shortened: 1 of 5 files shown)


bindings/go/evmc/evmc.go
```diff
@@ -42,7 +42,7 @@ static struct evmc_result execute_wrapper(struct evmc_vm* vm,
 		input_data,
 		input_size,
 		*value,
-		*create2_salt,
+		{{0}}, // create2_salt: not required for execution
```
chfast, line 45: I changed that from the original commit because `= {}` is not valid ISO C syntax.

yperbasis (approved):

include/evmc/evmc.h
```diff
@@ -137,7 +137,7 @@ struct evmc_message
     /**
      * The optional value used in new contract address construction.
      *
-     * Ignored unless kind is EVMC_CREATE2.
+     * Ignored unless kind is EVMC_CREATE2. Not required when invoking evmc_execute_fn().
```
gumb0, line 140: Minor suggestion to reword, but it's up to you.
```suggestion
     * Needed only for host to calculate created address when kind is EVMC_CREATE2. Ignored in evmc_execute_fn().
```

gumb0 (approved):

include/evmc/evmc.h
chfast: Thanks.

Mohsenghasemi12:
خوبه

PR #613, opened 2021-07-27
rust: Fixes suggested by clippy

vorot93:


bindings/rust/evmc-sys/Cargo.toml
```diff
@@ -13,4 +13,4 @@ categories = ["external-ffi-bindings"]
 edition = "2018"
 
 [build-dependencies]
-bindgen = "0.54.0"
+bindgen = "0.59"
```
axic, line 16: This bumps the minimum Rust requirement from 1.37 to 1.47, which is fine but the README must be updated accordingly.

axic: Moved this change to #638.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -341,8 +335,8 @@ impl<'a> ExecutionContext<'a> {
             gas: message.gas(),
             recipient: *message.recipient(),
             sender: *message.sender(),
-            input_data: input_data,
-            input_size: input_size,
+            input_data,
+            input_size,
```
axic, line 339: I do not much like this style, but it is prevalent in many languages now.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -524,10 +518,10 @@ fn from_buf_raw<T>(ptr: *const T, size: usize) -> Vec<T> {
     // Pre-allocate a vector.
     let mut buf = Vec::with_capacity(size);
     unsafe {
-        // Set the len of the vec manually.
-        buf.set_len(size);
         // Copy from the C buffer to the vec's buffer.
         std::ptr::copy(ptr, buf.as_mut_ptr(), size);
+        // Set the len of the vec manually.
+        buf.set_len(size);
```
axic, line 524: The reordering here is in line with the std: https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.set_len

Clippy was complaining that the data may be uninitialised, but since we set the length after copying, the warning is gone. The tests actually validate the `input` for different conversions.

bindings/rust/evmc-vm/src/lib.rs
```diff
@@ -452,9 +446,9 @@ unsafe fn deallocate_output_data(ptr: *const u8, size: usize) {
 }
 
 /// Returns a pointer to a heap-allocated evmc_result.
-impl Into<*const ffi::evmc_result> for ExecutionResult {
-    fn into(self) -> *const ffi::evmc_result {
-        let mut result: ffi::evmc_result = self.into();
+impl From<ExecutionResult> for *const ffi::evmc_result {
```
axic, line 449: These changes were recommended by clippy and defining `From` gives both `From` and `Into` traits.

PR #615, opened 2021-09-07
Clarify evmc_message::code_address usage

chfast:
- Clarifies the documentation around `evmc_message::code_address` and other `evmc_message` fields.
- Fixes the usage of `code_address` vs `destination` in the "precompiles vm" example.
- Resolves the issue reported in https://github.com/ethereum/evmc/pull/611#discussion_r702855921.


yperbasis (approved):

chfast:
@jlokier, please check if this clarifies the situation around precompiles implementation.

yperbasis (commented):
It also makes sense to change the description of `EVMC_CAPABILITY_PRECOMPILES` to say 
"...  for the range of code addresses" instead of "...  for the range of destination addresses".

PR #617, opened 2021-12-20
loader: Introduce error code for initialization

chfast:
Introduce special loader error code EVMC_LOADER_UNSPECIFIED_ERROR
for initializing evmc_loader_error_code objects. This allows defensive
usage pattern where user init error code to a value other than
EVMC_LOADER_SUCCESS what protects against a misbehaving loader
implementation.

Go and Java bindings use new value. Rust bindings do not have EVMC loader support.

### DIFF
--- a/include/evmc/loader.h
+++ b/include/evmc/loader.h
@@ -21,7 +21,10 @@ extern "C" {
 /** The function pointer type for EVMC create functions. */
 typedef struct evmc_vm* (*evmc_create_fn)(void);
 
-/** Error codes for the EVMC loader. */
+/// Error codes for the EVMC loader.
+///
+/// Objects of this type SHOULD be initialized with ::EVMC_LOADER_UNSPECIFIED_ERROR
+/// before passing to the EVMC loader.
 enum evmc_loader_error_code
 {
     /** The loader succeeded. */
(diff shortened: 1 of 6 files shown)


yperbasis (approved):

include/evmc/loader.h
```diff
@@ -21,7 +21,10 @@ extern "C" {
 /** The function pointer type for EVMC create functions. */
 typedef struct evmc_vm* (*evmc_create_fn)(void);
 
-/** Error codes for the EVMC loader. */
+/// Error codes for the EVMC loader.
```
gumb0, line 24: `///` for comments are not used anywhere else in this file

gumb0:
> This allows defensive usage pattern where user init error code to a value other than EVMC_LOADER_SUCCESS what protects against a misbehaving loader implementation.

So as I understand this allows to distinguish "loader did nothing" from "loader returned success".

include/evmc/loader.h
chfast: I noticed, but I wanted to use `//` everywhere since we enabled C99.

chfast:
> So as I understand this allows to distinguish "loader did nothing" from "loader returned success".

Yes. This seems better than `ec = EVMC_SUCCESS`.

gumb0 (approved):

PR #620, opened 2021-12-27
loader: Do not repeat strlen(options) condition

chfast:
This at least confuses static analyzers.

### DIFF
--- a/lib/loader/loader.c
+++ b/lib/loader/loader.c
@@ -267,16 +267,15 @@ struct evmc_vm* evmc_load_and_configure(const char* config, enum evmc_loader_err
     if (!vm)
         return NULL;
 
-    if (vm->set_option == NULL && strlen(options) != 0)
-    {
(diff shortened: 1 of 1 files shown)


lib/loader/loader.c
```diff
@@ -267,16 +267,15 @@ struct evmc_vm* evmc_load_and_configure(const char* config, enum evmc_loader_err
...
-    }
-
-
     while (strlen(options) != 0)
     {
+        if (vm->set_option == NULL)
+        {
+            ec = set_error(EVMC_LOADER_INVALID_OPTION_NAME, "%s (%s) does not support any options",
+                           vm->name, path);
+            goto exit;
```
gumb0, line 276: Could be `break;` now, but maybe keeping `goto` is fine for consistency.

gumb0 (approved):

lib/loader/loader.c
chfast: I use `goto exit` because the later `switch` is also using it in case of error.

PR #624, opened 2022-02-09
Split block and tx context, use out parameter

chfast:
This split transaction and block context into separate structs. They have different lifetimes so the Host can keep the same block context while processing multiple transactions.

We also return these by out-parameter because otherwise additional copy is needed in VM.

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -214,13 +214,13 @@ struct evmc_host_context;
 /**
  * Get transaction context callback function.
  *
- *  This callback function is used by an EVM to retrieve the transaction and
- *  block context.
+ * This callback function is used by an EVM to retrieve the transaction and block context.
  *
- *  @param      context  The pointer to the Host execution context.
- *  @return              The transaction context.
+ * @param[out] tx_context  The pointer to the transaction context struct to be filled.
+ * @param      host        The pointer to the Host execution context.
  */
-typedef struct evmc_tx_context (*evmc_get_tx_context_fn)(struct evmc_host_context* context);
+typedef void (*evmc_get_tx_context_fn)(struct evmc_tx_context* tx_context,
+                                       struct evmc_host_context* host);
 
 /**
  * Get block hash callback function.
--- a/examples/example_vm/example_vm.cpp
+++ b/examples/example_vm/example_vm.cpp
@@ -224,8 +224,9 @@ evmc_result execute(evmc_vm* instance,
 
         case OP_NUMBER:
         {
-            evmc_uint256be value =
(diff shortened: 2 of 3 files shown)


include/evmc/evmc.h
```diff
@@ -190,18 +190,23 @@ struct evmc_message
...
+
+/// The additional information about the block in which execution happens.
+struct evmc_block_context
+{
+    int64_t number;             ///< The block number.
+    int64_t timestamp;          ///< The block timestamp.
+    int64_t gas_limit;          ///< The block gas limit.
+    evmc_uint256be chain_id;    ///< The blockchain's ChainID.
+    evmc_uint256be base_fee;    ///< The block base fee per gas (EIP-1559, EIP-3198).
+    evmc_uint256be difficulty;  ///< The block difficulty.
```
yperbasis, line 208: I'd rename `difficulty` to `random` for compatibility with the upcoming [EIP-4399](https://eips.ethereum.org/EIPS/eip-4399).

bindings/go/evmc/host.go
```diff
@@ -1,14 +1,14 @@
...
 // Licensed under the Apache License, Version 2.0.
 
 package evmc
 
 /*
 #cgo CFLAGS: -I${SRCDIR}/../../../include -Wall -Wextra -Wno-unused-parameter
 
-#include <evmc/evmc.h>
-#include <evmc/helpers.h>
+#include </home/chfast/Projects/ethereum/evmc/include/evmc/evmc.h>
```
yperbasis, line 10: Probably unintentional change

chfast:
This can be made to work as desired without without ABI change using C `restrict`. But only in to-be-released clang 14.
```patch
diff --git a/include/evmc/evmc.hpp b/include/evmc/evmc.hpp
index 2007b94..8655531 100644
--- a/include/evmc/evmc.hpp
+++ b/include/evmc/evmc.hpp
@@ -563,6 +563,11 @@ public:
         return result{host->call(context, &message)};
     }

+    void helper(evmc_tx_context& __restrict__ tx) const noexcept
+    {
+        tx = host->get_tx_context(context);
+    }
+
     /// @copydoc HostInterface::get_tx_context()
     ///
     /// The implementation caches the received transaction context
@@ -572,7 +577,7 @@ public:
     evmc_tx_context get_tx_context() const noexcept final
     {
         if (tx_context.block_timestamp == 0)
-            tx_context = host->get_tx_context(context);
+            helper(tx_context);
         return tx_context;
     }
```

So I plan to do the ABI change but only in the aspect of parameter passing. The split will not be done until I collect stats about transaction access usage.

chfast:
For now #631 should be enough.

PR #636, opened 2022-05-06
C++: Conversion {address, bytes32} → bytes_view

chfast:
This change allows passing `address` and `bytes32` objects to functions taking `bytes_view` as argument.


axic (changes requested):
Can you add a test?

chfast:
> Can you add a test?

Added.

axic (approved):

PR #643, opened 2022-05-12
Convert evmc::hex library into single hex.hpp header

chfast:


include/evmc/hex.hpp
```diff
@@ -17,55 +17,105 @@ using bytes = std::basic_string<uint8_t>;
...
+        str += hex(b);
+    return str;
+}
 
-/// Creates error_code object out of a hex error code value.
-inline std::error_code make_error_code(hex_errc errc) noexcept
+namespace internal_hex
+{
+/// Extracts the nibble value out of a hex digit.
+/// Returns -1 in case of invalid hex degit.
```
gumb0, line 41:
```suggestion
/// Returns -1 in case of invalid hex digit.
```

include/evmc/hex.hpp
```diff
@@ -109,27 +110,33 @@ inline constexpr int from_hex_digit(char h)
...
-        throw hex_error{hex_errc::invalid_hex_digit};
+        return -1;
 }
 
-template <typename OutputIt>
-inline constexpr void from_hex(std::string_view hex, OutputIt result)
+inline constexpr bool isspace(char ch) noexcept
 {
-    // TODO: This can be implemented with hex_decode_iterator and std::copy.
+    // Implementation taken from LLVM's libc.
```
gumb0, line 56: Why not use standard one itself?

include/evmc/hex.hpp
```diff
@@ -17,55 +17,105 @@ using bytes = std::basic_string<uint8_t>;
...
-/// Encodes bytes as hex string.
-std::string hex(bytes_view bs);
-}  // namespace evmc
+    return internal_hex::from_hex(hex, noop_output_iterator{});
+}
 
-namespace std
+/// Decodes hex encoded string to bytes.
+///
+/// In case the input is invalid the returned value is null. This can happen if a non-hex digit
```
gumb0, line 110: Nit
```suggestion
/// In case the input is invalid the returned value is std::nullopt. This can happen if a non-hex digit
```

lib/tooling/run.cpp
```diff
@@ -71,8 +71,13 @@ int run(evmc::VM& vm,
...
         << " gas limit\n";
 
-    const auto code = from_hex(code_hex);
-    const auto input = from_hex(input_hex);
+    auto opt_code = from_hex(code_hex);
+    auto opt_input = from_hex(input_hex);
+    if (!opt_code || !opt_input)
+        throw std::invalid_argument{"invalid hex"};
+
+    const auto code = *std::move(opt_code);
```
gumb0, line 79: Not sure if this moves value out of optional or moves entire optional to new optional variable...

Maybe just 
```suggestion
    const auto& code = *opt_code;
```

include/evmc/hex.hpp
```diff
@@ -174,7 +174,9 @@ inline bytes from_hex(std::string_view hex)
 {
     bytes bs;
     bs.reserve(hex.size() / 2);
-    internal_hex::from_hex(hex, std::back_inserter(bs));
+    const auto ec = internal_hex::from_hex(hex, std::back_inserter(bs));
+    if (static_cast<int>(ec) != 0)
```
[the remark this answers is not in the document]
gumb0: Haven't seen later comment removing the codes. Weird that github doesn't show this as outdated.

include/evmc/hex.hpp
```diff
@@ -17,55 +17,105 @@ using bytes = std::basic_string<uint8_t>;
...
-    return {hex_chars[b >> 4], hex_chars[b & 0xf]};
+    // Omit the optional 0x prefix.
+    if (hex.size() >= 2 && hex[0] == '0' && hex[1] == 'x')
+        hex.remove_prefix(2);
+
+    constexpr int empty_byte_mark = -1;
+    int b = empty_byte_mark;
+    for (const auto h : hex)
+    {
+        if (isspace(h))
```
axic, line 71: Why allow space here? Why not allow space in the prefix?

include/evmc/hex.hpp
chfast: Not constexpr.

axic: Should have a test for it then 😬

lib/tooling/run.cpp
chfast: This is to call `operator*()&&` of `std::optional`. But your suggestion is better.

gumb0 (approved):

include/evmc/hex.hpp
```diff
@@ -17,55 +17,105 @@ using bytes = std::basic_string<uint8_t>;
...
+inline constexpr bool from_hex(std::string_view hex, OutputIt result) noexcept
 {
-    static constexpr auto hex_chars = "0123456789abcdef";
-    return {hex_chars[b >> 4], hex_chars[b & 0xf]};
+    // Omit the optional 0x prefix.
+    if (hex.size() >= 2 && hex[0] == '0' && hex[1] == 'x')
+        hex.remove_prefix(2);
+
+    constexpr int empty_byte_mark = -1;
+    int b = empty_byte_mark;
```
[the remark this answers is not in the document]
chfast: Done.

test/unittests/hex_test.cpp
```diff
@@ -86,34 +73,33 @@ TEST(hex, from_hex_skip_whitespace)
...
-    std::error_code ec1 = hex_errc::invalid_hex_digit;
-    EXPECT_EQ(ec1.category(), hex_category());
+    // Test internal isspace() compliance with std::isspace().
+    // The https://en.cppreference.com/w/cpp/string/byte/isspace has the list of "space" characters.
 
-    std::error_code ec2 = hex_errc::incomplete_hex_byte_pair;
-    EXPECT_EQ(ec2.category(), hex_category());
+    for (int i = int{std::numeric_limits<char>::min()}; i <= std::numeric_limits<char>::max(); ++i)
+    {
+        switch (const auto c = static_cast<char>(i); c)
```
axic, line 90: Maybe add `std::isspace(c) == evmc::internal_hex::isspace(c)` too? Probably not needed.

include/evmc/hex.hpp
axic: Yes.

include/evmc/hex.hpp
```diff
@@ -26,9 +27,47 @@ enum class hex_errc
     /// Input contains incomplete hex byte (length is odd).
     incomplete_hex_byte_pair = 2,
 };
+}  // namespace evmc
+
+namespace std
+{
+/// Template specialization of std::is_error_code_enum for evmc::hex_errc.
+/// This enabled implicit conversions from evmc::hex_errc to std::error_code.
```
axic:
```suggestion
/// This enables implicit conversions from evmc::hex_errc to std::error_code.
```

test/unittests/hex_test.cpp
chfast: Ah, that's actually much better idea than my test.

test/unittests/hex_test.cpp
```diff
@@ -86,34 +74,35 @@ TEST(hex, from_hex_skip_whitespace)
...
+        const auto c = static_cast<char>(i);
+        EXPECT_EQ(evmc::internal_hex::isspace(c), (std::isspace(c) != 0));
+        switch (c)
+        {
+        case ' ':
+        case '\f':
+        case '\n':
+        case '\r':
+        case '\t':
+        case '\v':
```
axic, line 100: I actually wonder if `isspace` should be used in the hex parser or just checking for space would be enough?

chfast: I need at least ` `, `\r`, `\n`.

PR #645, opened 2022-05-16
test: Test compilation without exceptions

chfast:


axic (commented):
Why does this compile before merging the hex noexcept changes?

chfast:
> Why does this compile before merging the hex noexcept changes?

Ha, good question. It is because the `evmc::hex` is still implemented in side `hex.cpp` file.

I'm not sure this is any good test. I was using it while reworking hex do I decided to drop it anyway. The full no-exception build should be better.

chfast:
Now it makes more sense.

axic (approved):
Though it was never verified why this worked before merging #643 ?

PR #647, opened 2022-05-23
tool: Use @file syntax to loading data from a file

chfast:
For `--input` and `code` arguments allow loading data from a file by preceding the file path with @ character. E.g. `--input @sha1.in`.

Change the type of code and input parameters of `tooling::run()` from hex-encoded string to bytes. The hex decoding must happen outside. This moves the hex decoding errors out of the function concern.

### DIFF
--- a/tools/evmc/main.cpp
+++ b/tools/evmc/main.cpp
@@ -10,25 +10,27 @@
 
 namespace
 {
-/// Returns the input str if already valid hex string. Otherwise, interprets the str as a file
-/// name and loads the file content.
+/// If the argument starts with @ returns the content of the file at the path following @.
+/// Otherwise, returns the argument.
 /// @todo The file content is expected to be a hex string but not validated.
 std::string load_hex(const std::string& str)
 {
-    if (evmc::validate_hex(str))
-        return str;
+    if (str[0] == '@')  // The argument is file path.
+    {
+        std::ifstream file{str.c_str() + 1};
+        return {std::istreambuf_iterator<char>{file}, std::istreambuf_iterator<char>{}};
+    }
 
-    // Must be a file path.
-    std::ifstream file{str};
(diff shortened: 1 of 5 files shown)


tools/evmc/main.cpp
```diff
@@ -10,25 +10,27 @@
...
-struct HexValidator : public CLI::Validator
+struct HexOrFileValidator : public CLI::Validator
 {
-    HexValidator() : CLI::Validator{"HEX"}
+    HexOrFileValidator() : CLI::Validator{"HEX|@FILE"}
     {
-        name_ = "HEX";
         func_ = [](const std::string& str) -> std::string {
+            if (!str.empty() && str[0] == '@')
+                return CLI::ExistingFile(str.substr(1));
```
axic, line 37: Shouldn't this try to load & verify the content?

tools/evmc/main.cpp
```diff
@@ -10,25 +10,27 @@
...
+/// If the argument starts with @ returns the content of the file at the path following @.
+/// Otherwise, returns the argument.
 /// @todo The file content is expected to be a hex string but not validated.
 std::string load_hex(const std::string& str)
 {
-    if (evmc::validate_hex(str))
-        return str;
+    if (str[0] == '@')  // The argument is file path.
+    {
+        std::ifstream file{str.c_str() + 1};
```
axic, line 20: Why use `c_str() + 1` here and `substr(1)` below?

tools/evmc/main.cpp
chfast: It could but it feels too much. I don't know any CLI library doing so. And you would need to load the file twice.

axic: Since the files are usually small, the overhead may be worth it here for easy-of-use (and not relying on exceptions).

tools/evmc/main.cpp
chfast: I actually changed this to add more info to the error.

PR #648, opened 2022-05-26
Refactor hex decoding utilities

chfast:
Generalize the `core from_hex()` to take the input as pair in input iterators. The optional 0x prefix is still omitted but the "skip space" feature has been dropped.

Add `skip_space_iterator`. This restores the option to filter-out whitespace before parsing hex. This is useful e.g. when loading hex from a file.

### DIFF
--- a/test/unittests/filter_iterator_test.cpp
+++ b/test/unittests/filter_iterator_test.cpp
@@ -1,0 +1,107 @@
+// EVMC: Ethereum Client-VM Connector API.
+// Copyright 2022 The EVMC Authors.
+// Licensed under the Apache License, Version 2.0.
+
+#include <tools/evmc/filter_iterator.hpp>
+#include <gtest/gtest.h>
+#include <cctype>
+
+using evmc::skip_space_iterator;
+
+namespace
+{
+std::string remove_space(std::string_view in)
+{
+    // Copy input to additional buffer. This helps with out-of-buffer reads detection by sanitizers.
+    const std::vector<char> in_buffer(in.begin(), in.end());
+
+    // Filter the input.
+    std::string out;
+    std::copy(skip_space_iterator{in_buffer.begin(), in_buffer.end()},
+              skip_space_iterator{in_buffer.end(), in_buffer.end()}, std::back_inserter(out));
+    return out;
+}
+
+bool is_positive(int x) noexcept
+{
+    return x > 0;
+}
+}  // namespace
+
+
+TEST(filter_iterator, filter_positive_integers)
+{
+    std::vector<int> in{1, 0, 0, 2, -3, 3, 4, 5, 0, 6, 7, -1, -2, 0, 8, 9, -10};
+    std::vector<int> out;
+
+    using iter = evmc::filter_iterator<std::vector<int>::const_iterator, is_positive>;
+    std::copy(iter{in.begin(), in.end()}, iter{in.end(), in.end()}, std::back_inserter(out));
+    ASSERT_EQ(out.size(), 9u);
+    EXPECT_EQ(out[0], 1);
+    EXPECT_EQ(out[1], 2);
+    EXPECT_EQ(out[2], 3);
+    EXPECT_EQ(out[3], 4);
+    EXPECT_EQ(out[4], 5);
+    EXPECT_EQ(out[5], 6);
+    EXPECT_EQ(out[6], 7);
+    EXPECT_EQ(out[7], 8);
+    EXPECT_EQ(out[8], 9);
+}
+
+
+TEST(skip_space_iterator, empty)
+{
+    EXPECT_EQ(remove_space(""), "");
+    EXPECT_EQ(remove_space(" "), "");
+    EXPECT_EQ(remove_space("  "), "");
+}
+
+TEST(skip_space_iterator, filter_middle)
+{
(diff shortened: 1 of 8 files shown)


axic (approved):
Looks okay to me (at least the `from_hex` part). Would be nice to get a check from @gumb0.

test/unittests/tooling_test.cpp
```diff
@@ -160,7 +159,7 @@ TEST(tool_commands, bench_inconsistent_output)
     auto vm = evmc::VM{evmc_create_example_vm()};
     std::ostringstream out;
 
-    const auto code = *from_hex("6000 54 6001 6000 55 6000 52 6001 601f f3");
+    const auto code = *from_hex("60005460016000556000526001601ff3");
```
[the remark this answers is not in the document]
chfast: No. We apply filter only when loading from a file.

tools/evmc/skip_space_iterator.hpp
```diff
@@ -0,0 +1,60 @@
...
+/// The constexpr variant of std::isspace().
+inline constexpr bool isspace(char ch) noexcept
+{
+    // Implementation taken from LLVM's libc.
+    return ch == ' ' || (static_cast<unsigned>(ch) - '\t') < 5;
+}
+
+/// The input filter iterator which skips whitespace characters from the base input iterator.
+template <typename BaseIterator>
+struct skip_space_iterator
```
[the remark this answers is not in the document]
chfast: We extract the filter function (`isspace`) to get generic filter iterator (like `boost::filter_iterator`). This can be done later to filter out literal separator (`'`).

test/unittests/skip_space_iterator_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include <cctype>
+
+using evmc::skip_space_iterator;
+
+namespace
+{
+std::string remove_space(std::string_view in)
+{
+    // Copy input to additional buffer. This helps with out-of-buffer reads detection by sanitizers.
+    const auto in_buffer = std::make_unique<char[]>(in.size());
```
gumb0, line 16: Maybe for a test just simpler to use `std::string in_buffer(in)`

tools/evmc/skip_space_iterator.hpp
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+public:
+    constexpr skip_space_iterator(BaseIterator it, BaseIterator end) noexcept
+      : base{it}, base_end{end}
+    {
+        forward_to_next_value();
+    }
+
+    constexpr auto operator*() { return value; }
```
gumb0, line 49: Why keep `value` as a member and not just
```suggestion
    constexpr auto operator*() { return *base; }
```

You version will return last valid value even if iterator already points to an end.

test/unittests/skip_space_iterator_test.cpp
gumb0: Could be `vector<char>` maybe. Just doing this in 4 lines feels like unnecessary complication.

tools/evmc/skip_space_iterator.hpp
chfast: You should not read from an input iterator twice. So the only read is in `forward_to_next_value()` and we need to cache the value.

test/unittests/filter_iterator_test.cpp
```diff
@@ -0,0 +1,107 @@
...
+{
+std::string remove_space(std::string_view in)
+{
+    // Copy input to additional buffer. This helps with out-of-buffer reads detection by sanitizers.
+    const std::vector<char> in_buffer(in.begin(), in.end());
+
+    // Filter the input.
+    std::string out;
+    std::copy(skip_space_iterator{in_buffer.begin(), in_buffer.end()},
+              skip_space_iterator{in_buffer.end(), in_buffer.end()}, std::back_inserter(out));
```
axic, line 21: What is the point of `end,end`?

chfast: For filter iterator you need iterator pair so for the end filter iterator double end is provided. I considered some other options (default argument value, sentinel) but `boost::filter_iterator` does the same so I left it as is.

axic: But couldn't you just pass `end()` as the iterator?

gumb0 (approved):

test/unittests/filter_iterator_test.cpp
axic:
```
std::copy(skip_space_iterator{in_buffer.begin(), in_buffer.end()}, in_buffer.end(), std::back_inserter(out));
```

Isn't `end` an iterator itself?

chfast: Yes, this is also "workable" but the iterators would have different types.
1. Additional overloads for comparison operators are needed (so we can compare `filter_iterator` with `BaseIterator`).
2. The `std::copy()` and many algorithms cannot be used because they expect `InputIterator`s to be of the same type. https://en.cppreference.com/w/cpp/algorithm/copy.

I expect much better user experience with `std::range`, but I also have 0 experience with these so far.

axic: Alright, I was wondering as it looked weird. Fine this way to avoid more code.

axic (approved):

PR #649, opened 2022-05-26
hex: Add from_hex<T> for user types

chfast:
Depends on #648.

### DIFF
--- a/include/evmc/hex.hpp
+++ b/include/evmc/hex.hpp
@@ -113,4 +113,26 @@ inline std::optional<bytes> from_hex(std::string_view hex)
         return {};
     return bs;
 }
+
+/// Decodes hex-encoded string into custom type T with .bytes array of uint8_t.
+///
+/// When the input is smaller than the result type, the result is padded with zeros on the left
+/// (the result bytes of lowest indices are filled with zeros).
+/// TODO: Support optional left alignment.
+template <typename T>
+constexpr std::optional<T> from_hex(std::string_view s) noexcept
+{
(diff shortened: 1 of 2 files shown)


include/evmc/hex.hpp
```diff
@@ -113,4 +113,24 @@ inline std::optional<bytes> from_hex(std::string_view hex)
         return {};
     return bs;
 }
+
+/// Decodes hex-encoded string into custom type T with .bytes array of uint8_t.
+///
+/// The decoded bytes are right aligned in case the input is smaller than the result type.
```
axic, line 119: Why?

chfast: Because they imitate BE numbers.

chfast: Yes, but left/right does not have clear meaning.

chfast: > When the input is smaller than the result type, the result is padded with zeros on the left (the result bytes of lowest indices are filled with zeros).

?

chfast: Yes, an option can be added if needed, but so far this is only used for literals and loading storage values from JSON.

gumb0 (approved):

include/evmc/hex.hpp
axic: Perhaps add `// TODO: support left alignment too with a flag`.

axic (approved):

axic (commented):
Perhaps add `// TODO: support left alignment too with a flag`.

PR #651, opened 2022-05-30
Fix java bindings

vvviiimmm:
1. `vm_execute` input data was not handled properly. An additional `inputData` parameter has been added to `EvmcVm`'s `execute` function to avoid pointer arithmetics in java
2. `ReleaseByteArrayElements` was called on a pointer that is not derived from `GetByteArrayElements` causing segfaults
3. `get_block_hash_fn` typo

Also, it was quite challenging to encode `evmc_message` into ByteBuffer and then decode the result. I'd suggest capturing that logic into java classes so the callers won't have to work with memory buffers directly.

### DIFF
--- a/bindings/java/c/evmc-vm.c
+++ b/bindings/java/c/evmc-vm.c
@@ -97,7 +97,8 @@ JNIEXPORT jobject JNICALL Java_org_ethereum_evmc_EvmcVm_execute(JNIEnv* jenv,
                                                                 jobject jcontext,
                                                                 jint jrev,
                                                                 jobject jmsg,
-                                                                jobject jcode)
+                                                                jobject jcode,
+                                                                jobject jinput_data)
 {
     (void)jcls;
     struct evmc_message* msg = (struct evmc_message*)(*jenv)->GetDirectBufferAddress(jenv, jmsg);
@@ -112,6 +113,7 @@ JNIEXPORT jobject JNICALL Java_org_ethereum_evmc_EvmcVm_execute(JNIEnv* jenv,
     struct evmc_result* result =
         (struct evmc_result*)(*jenv)->GetDirectBufferAddress(jenv, jresult);
     assert(result != NULL);
+    msg->input_data = GetDirectBuffer(jenv, jinput_data, &msg->input_size);
     *result = evmc_execute(evm, host, (struct evmc_host_context*)jcontext, (enum evmc_revision)jrev,
                            msg, code, code_size);
     return jresult;
--- a/bindings/java/c/host.c
+++ b/bindings/java/c/host.c
@@ -166,8 +166,6 @@ static evmc_uint256be get_balance_fn(struct evmc_host_context* context, const ev
     evmc_uint256be result;
     CopyFromByteBuffer(jenv, jresult, &result, sizeof(evmc_uint256be));
 
-    (*jenv)->ReleaseByteArrayElements(jenv, jaddress, (jbyte*)address, 0);
-
     return result;
 }
 
@@ -225,8 +223,6 @@ static evmc_bytes32 get_code_hash_fn(struct evmc_host_context* context, const ev
     evmc_bytes32 result;
     CopyFromByteBuffer(jenv, jresult, &result, sizeof(evmc_bytes32));
 
-    (*jenv)->ReleaseByteArrayElements(jenv, jaddress, (jbyte*)address, 0);
-
     return result;
 }
 
@@ -275,8 +271,6 @@ static size_t copy_code_fn(struct evmc_host_context* context,
         if (length > 0)
             memcpy(buffer_data, code + code_offset, length);
     }
-
-    (*jenv)->ReleaseByteArrayElements(jenv, jaddress, (jbyte*)address, 0);
 
     return length;
 }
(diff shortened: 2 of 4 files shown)


axic:
Thanks! It seems to fail on CI, can you please check?

> Also, it was quite challenging to encode evmc_message into ByteBuffer and then decode the result. I'd suggest capturing that logic into java classes so the callers won't have to work with memory buffers directly.

Absolutely, we did some work on it but didn't had the time to finish. See #542 and #555.

bindings/java/c/host.c
```diff
@@ -369,7 +363,7 @@ static struct evmc_tx_context get_tx_context_fn(struct evmc_host_context* contex
 
 static evmc_bytes32 get_block_hash_fn(struct evmc_host_context* context, int64_t number)
 {
-    char java_method_name[] = "get_code_hash";
+    char java_method_name[] = "get_block_hash_fn";
```
axic, line 366: Oops. Shows the lack of units tests for Java.

bindings/java/c/evmc-vm.c
```diff
@@ -112,6 +113,10 @@ JNIEXPORT jobject JNICALL Java_org_ethereum_evmc_EvmcVm_execute(JNIEnv* jenv,
     struct evmc_result* result =
         (struct evmc_result*)(*jenv)->GetDirectBufferAddress(jenv, jresult);
     assert(result != NULL);
+    size_t input_data_size;
+    const uint8_t* input_data = GetDirectBuffer(jenv, jinputdata, &input_data_size);
+    char* input_data_addr = (char*)msg + offsetof(struct evmc_message, input_data);
+    memcpy(input_data_addr, &input_data, sizeof(uint8_t*));
```
axic, line 119: Why does it need to use this convoluted approach of memory copying, as opposed to just `msg.input_data = GetDirectBuffer(jenv, jinputdata, &input_data_size);` ?

bindings/java/c/host.c
```diff
@@ -276,8 +272,6 @@ static size_t copy_code_fn(struct evmc_host_context* context,
             memcpy(buffer_data, code + code_offset, length);
     }
 
-    (*jenv)->ReleaseByteArrayElements(jenv, jaddress, (jbyte*)address, 0);
```
axic, line 279: > ReleaseByteArrayElements was called on a pointer that is not derived from GetByteArrayElements causing segfaults

These helpers rely on `CopyDataToJava` which uses `NewByteArray/SetByteArrayRegion`. Are you saying those cannot be freed with it?

bindings/java/c/host.c
axic: Extracted this into #653.

bindings/java/c/evmc-vm.c
axic: Actually should be `msg.input_data = GetDirectBuffer(jenv, jinputdata, &msg.input_size);` as the size setting is missing right now.

bindings/java/c/evmc-vm.c
```diff
@@ -97,7 +97,8 @@ JNIEXPORT jobject JNICALL Java_org_ethereum_evmc_EvmcVm_execute(JNIEnv* jenv,
                                                                 jobject jcontext,
                                                                 jint jrev,
                                                                 jobject jmsg,
-                                                                jobject jcode)
+                                                                jobject jcode,
+                                                                jobject jinputdata)
```
axic, line 101:
```suggestion
                                                                jobject jinput)
```

Would name it like this if there are no symbol clashes.

bindings/java/c/host.c
vvviiimmm: Can't say for sure, but isn't `NewByteArray` creating a local ref that will be garbage collected? https://docs.oracle.com/javase/10/docs/specs/jni/functions.html#local-references

bindings/java/java/src/main/java/org/ethereum/evmc/EvmcVm.java
```diff
@@ -144,16 +144,21 @@ public String version() {
    * <p>This is a mandatory method and MUST NOT be set to NULL.
    */
   private static native ByteBuffer execute(
-      ByteBuffer nativeVm, HostContext context, int rev, ByteBuffer msg, ByteBuffer code);
+      ByteBuffer nativeVm,
+      HostContext context,
+      int rev,
+      ByteBuffer msg,
+      ByteBuffer code,
+      ByteBuffer inputdata);
```
axic, line 152: Same here.

bindings/java/java/src/test/java/org/ethereum/evmc/EvmcTest.java
```diff
@@ -230,16 +255,21 @@ void testExecute_EVMC_CREATE() throws Exception {
...
-      char[] inputData = "hello w\0".toCharArray();
+      char[] inputDataPlaceholder = {};
+      byte[] inputDataBytes = {0x68, 0x65, 0x6c, 0x6c, 0x6f};
+      ByteBuffer inputDataByteBuffer =
+          ByteBuffer.wrap(inputDataBytes).order(ByteOrder.nativeOrder());
       long gas = 200000;
       int depth = 0;
       ByteBuffer msg =
-          new TestMessage(kind, sender, recipient, value, inputData, gas, depth).toByteBuffer();
+          new TestMessage(kind, sender, recipient, value, inputDataPlaceholder, gas, depth)
```
axic, line 265: I was wondering what this `TestMessage` is and why it needs the `inputDataPlaceholder`.

It turns out `vm.execute` still gets this entire data structure with `input_data`/`input_length`.

This is a bad idea having shadow data which is not checked. Probably the better option is removing those from the `msg` structure, which also means it needs to be reassambled in the binding.

chfast:
@vvviiimmm Are you using EVMC in Java?

axic:

PR #654, opened 2022-06-03
hex: Add from_spaced_hex()

chfast:
The `from_spaced_hex()` is `from_hex()` extension where the whitespace in the input is ignored.


include/evmc/filter_iterator.hpp
```diff
@@ -32,10 +33,19 @@ template <typename BaseIterator,
           bool predicate(typename std::iterator_traits<BaseIterator>::value_type) noexcept>
 struct filter_iterator
 {
+    /// The iterators' difference type.
```
gumb0, line 36: either `iterator's` or just `iterator`, make it consistent with comments below

include/evmc/filter_iterator.hpp
```diff
@@ -54,26 +64,32 @@ struct filter_iterator
...
 
+    /// The increment operator.
     constexpr void operator++() noexcept
     {
         ++base;
         forward_to_next_value();
     }
 
-    constexpr bool operator!=(const filter_iterator& o) const noexcept { return base != o.base; }
+    /// The equal operator.
```
gumb0, line 88: I think it should be "equality"/"inequality" operator 
https://docs.microsoft.com/en-us/cpp/cpp/equality-operators-equal-equal-and-exclpt-equal?view=msvc-170

include/evmc/filter_iterator.hpp
chfast: This was suppose to be "two iterators' difference type".

include/evmc/filter_iterator.hpp
```diff
@@ -83,6 +99,7 @@ struct skip_space_iterator : filter_iterator<BaseIterator, is_not_space>
     using filter_iterator<BaseIterator, is_not_space>::filter_iterator;
 };
 
+/// CTAD.
```
gumb0, line 102: a bit cryptic
```suggestion
/// Class template argument deduction guide.
```

include/evmc/filter_iterator.hpp
gumb0: Fine with changing to "Two"

gumb0 (approved):
Looks good, only minor suggestions in comments.

gumb0 (approved):

PR #661, opened 2022-08-05
Extend storage status

chfast:
Expand the number of storage statuses to cover all possible EVM
behaviors. These provide enough information to VM to compute not only
gas costs but also refunds. Including legacy SSTORE costs before net
gas metering.

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -498,40 +498,97 @@ typedef evmc_bytes32 (*evmc_get_storage_fn)(struct evmc_host_context* context,
 /**
  * The effect of an attempt to modify a contract storage item.
  *
+ * See @ref storagestatus for additional information about design of this enum
+ * and analysis of the specification.
+ *
  * For the purpose of explaining the meaning of each element, the following
  * notation is used:
  * - 0 is zero value,
  * - X != 0 (X is any value other than 0),
- * - Y != X, Y != 0 (Y is any value other than X and 0),
- * - Z != Y (Z is any value other than Y),
- * - the "->" means the change from one value to another.
+ * - Y != 0, Y != X,  (Y is any value other than X and 0),
+ * - Z != 0, Z != X, Z != X (Z is any value other than Y and X and 0),
+ * - the "o -> c -> v" triple describes the change status in the context of:
+ *   - o: original value (cold value before a transaction started),
+ *   - c: current storage value,
+ *   - v: new storage value to be set.
+ *
+ * The order of elements follows EIPs introducing net storage gas costs:
+ * - EIP-2200: https://eips.ethereum.org/EIPS/eip-2200,
+ * - EIP-1283: https://eips.ethereum.org/EIPS/eip-1283.
  */
 enum evmc_storage_status
 {
     /**
-     * The value of a storage item has been left unchanged: 0 -> 0 and X -> X.
+     * The new/same value is assigned to the storage item without affecting the cost structure.
+     *
+     * The storage value item is either:
+     * - left unchanged (c == v) or
+     * - the dirty value (o != c) is modified again (c != v).
+     * This is the group of cases related to minimal gas cost of only accessing warm storage.
+     * 0|X   -> 0 -> 0 (current value unchanged)
+     * 0|X|Y -> Y -> Y (current value unchanged)
+     * 0|X   -> Y -> Z (modified previously added/modified value)
+     *
+     * This is "catch all remaining" status. I.e. if all other statuses are correctly matched
+     * this status should be assigned to all remaining cases.
      */
-    EVMC_STORAGE_UNCHANGED = 0,
+    EVMC_STORAGE_ASSIGNED = 0,
 
     /**
-     * The value of a storage item has been modified: X -> Y.
+     * A new storage item is added by changing
+     * the current clean zero to a nonzero value.
+     * 0 -> 0 -> Z
      */
-    EVMC_STORAGE_MODIFIED = 1,
+    EVMC_STORAGE_ADDED = 1,
 
     /**
-     * A storage item has been modified after being modified before: X -> Y -> Z.
+     * A storage item is deleted by changing
+     * the current clean nonzero to the zero value.
+     * X -> X -> 0
      */
-    EVMC_STORAGE_MODIFIED_AGAIN = 2,
+    EVMC_STORAGE_DELETED = 2,
 
     /**
-     * A new storage item has been added: 0 -> X.
+     * A storage item is modified by changing
+     * the current clean nonzero to other nonzero value.
+     * X -> X -> Z
      */
-    EVMC_STORAGE_ADDED = 3,
+    EVMC_STORAGE_MODIFIED = 3,
 
     /**
-     * A storage item has been deleted: X -> 0.
+     * A storage item is added by changing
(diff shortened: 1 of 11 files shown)


chfast:
I have Client-side implementation of this using only 4 checks:
- `original != current` (dirty)
- `original == new_value` (restored)
- `current != 0`
- `new_value != 0`

https://github.com/ethereum/evmone/blob/state/test/state/host.cpp#L99-L162 (unstable link).

Then you can get the gas cost and refund per revision from a lookup table.

include/evmc/evmc.h
```diff
@@ -503,36 +503,111 @@ typedef evmc_bytes32 (*evmc_get_storage_fn)(struct evmc_host_context* context,
...
+     * in the transaction context. This is the group of "no-op" cases related to
+     * minimal gas cost.
+     * 0|X   -> 0 -> 0 (current value unchanged)
+     * 0|X|Y -> Y -> Y (current value unchanged)
+     * 0|X   -> Y -> Z (modified previously added/modified value)
+     *
+     * This corresponds to the EIP-2200 rule 2 where:
+     * c == v.
+     */
+    EVMC_STORAGE_MODIFIED_AGAIN = 0,
```
yperbasis, line 530: Why `EVMC_STORAGE_MODIFIED_AGAIN` rather than `EVMC_STORAGE_UNCHANGED`?

chfast: It felt more correct to me: in general we modify a value, just gas cost does not care. The "unchanged" strongly suggests that `current == value` what is not always the case for this group. But I can change it to `EVMC_STORAGE_UNCHANGED` if people think this is fine.

include/evmc/evmc.h
```diff
@@ -503,36 +503,111 @@ typedef evmc_bytes32 (*evmc_get_storage_fn)(struct evmc_host_context* context,
...
+ *
+ * The order of elements follows EIPs introducing net storage gas costs:
+ * - EIP-2200: https://eips.ethereum.org/EIPS/eip-2200,
+ * - EIP-1283: https://eips.ethereum.org/EIPS/eip-1283.
  */
 enum evmc_storage_status
 {
     /**
-     * The value of a storage item has been left unchanged: 0 -> 0 and X -> X.
+     * The value of a storage item has been left unchanged has been modified again
```
chfast, line 520: "or" missing.

include/evmc/evmc.h
yperbasis: What is this rule 2 of EIP-2200 referenced in the doc? "If _current value_ equals _new value_ (this is a no-op), `SLOAD_GAS` is deducted."? Or something else?

chfast: Oh, this is a mistake. This a group of this rule (2) and 3.2 without matching any subrules of 3.2. If you can parse my DSL, this can be e.g. `X → Y → Z`. Because these have the same cost, I decided to use single enum value for it.

If you think this is too confusing, I can assign separate values to these. But this suggests these have different behavior what is not the case.

yperbasis: I'd prefer separate `EVMC_STORAGE_UNCHANGED` and `EVMC_STORAGE_MODIFIED_AGAIN` for extra clarity, even if they are priced the same.

chfast: I reviewed my work on this and distinguishing these two requires additional check `current == new` (not needed to distinguish all other cases). Moreover, this would be inconvenient to test: e.g. state tests are not able to spot a mistake.

So I'm against.

yperbasis: OK, then please amend the comment to say that it covers both rule 2 as well as 3.2 without any of the sub-rules.

chfast: Please see the changed version.

yperbasis: What about calling this guy `EVMC_STORAGE_NO_OP`?

include/evmc/evmc.h
```diff
@@ -502,36 +502,114 @@ typedef evmc_bytes32 (*evmc_get_storage_fn)(struct evmc_host_context* context,
...
+     * minimal gas cost.
+     * 0|X   -> 0 -> 0 (current value unchanged)
+     * 0|X|Y -> Y -> Y (current value unchanged)
+     * 0|X   -> Y -> Z (modified previously added/modified value)
+     *
+     * This is "catch all remaining" status. I.e. if all other statuses are correctly matched
+     * this status should be assigned to all remaining cases.
+     *
+     * This corresponds to the EIP-2200 rule 2 or rule 3.2 without sub-rules, where:
+     * c == v or (c != v, o != c, (o == 0 or (o != 0, c != 0, v != 0)), o != v)
```
gumb0, line 530: This is hard to parse, and rule numbers like `3.2.1.1` are not useful because don't exist in EIP.

I perfectly understand notation like `X -> Y -> 0`, why do you need this new one?

chfast: This is redundant. I try to explain what EIP rules it matches with the ordered list on condition to reach the particular rules in the spec tree in the EIP.

Any ideas how to improve this?

include/evmc/evmc.h
chfast: I think `NO_OP` may be ok, however the spec uses this term in `current == value` context only.

Another possible name can be `UNAFFECTED`.

yperbasis: `UNAFFECTED` sounds good.

chfast: `REASSIGNED`?

Time for twitter pool.

include/evmc/evmc.h
chfast: Do you think it is better to remove it? I can have separate doxygen chapter (md document) that describes all matching details using big tables.

include/evmc/evmc.h
yperbasis: `REASSIGNED` is fine with me.

docs/storage_change_status.md
```diff
@@ -0,0 +1,320 @@
+# Storage change status
+
+> Analysis of possible storage change statuses in [EIP-2200].
+
+## Specification
+
+> This is the [EIP-2200] specification with modifications:
+> - the clause tree has been converted to ordered lists to be referencable,
```
gumb0, line 8:
```suggestion
> - the clause tree has been converted to ordered lists to be referenceable,
```

chfast: CLion says otherwise.

include/evmc/evmc.h
chfast: Based on the comments from @gumb0 and my preferences, can we leave it as `EVMC_STATUS_MODIFIED_AGAIN`? Also documentation is updated.

chfast:
Documentation rendered:
- https://output.circle-artifacts.com/output/job/df5edcb0-85f0-4745-87b8-a464456a4c34/artifacts/0/docs/storagestatus.html
- https://output.circle-artifacts.com/output/job/df5edcb0-85f0-4745-87b8-a464456a4c34/artifacts/0/docs/group__EVMC.html#gae012fd6b8e5c23806b507c2d3e9fb1aa

yperbasis (approved):

include/evmc/evmc.h
chfast: Changed to `EVMC_STORAGE_ASSIGNED`.

PR #665, opened 2022-08-10
C++: Rename evmc::result → evmc::Result

chfast:
Rename the `evmc::result` to `evmc::Result` for consistency with C++ types of similar kind.

### DIFF
--- a/include/evmc/evmc.hpp
+++ b/include/evmc/evmc.hpp
@@ -328,7 +328,7 @@ constexpr auto make_result = evmc_make_result;
 ///
 /// This is a RAII wrapper for evmc_result and objects of this type
 /// automatically release attached resources.
-class result : private evmc_result
+class Result : private evmc_result
 {
 public:
     using evmc_result::create_address;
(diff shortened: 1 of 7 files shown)


include/evmc/evmc.hpp
```diff
@@ -328,7 +328,7 @@ constexpr auto make_result = evmc_make_result;
 ///
 /// This is a RAII wrapper for evmc_result and objects of this type
 /// automatically release attached resources.
-class result : private evmc_result
+class Result : private evmc_result
```
axic, line 331: What is the reasoning behind this? Complex data types are upper case, simple data types (bytes32 for example) are lowercase?

chfast: For consistency with `evmc::VM`, `evmc::Host`, `evmc::HostInterface`, etc. Not that we have perfect consistency but upper case names are easier to use in C++. @yperbasis was in favor.

gumb0 (approved):

PR #666, opened 2022-08-12
Add gas refund to evmc_result

### LINKED ISSUE #413: Calculate refunds on VM side

chfast:
Add `gas_refund` field to `evmc_result` and `evmc::Result`.

Closes #413.
Replaces #417

### DIFF
--- a/include/evmc/evmc.hpp
+++ b/include/evmc/evmc.hpp
@@ -345,33 +345,39 @@ public:
     ///
     /// @param _status_code  The status code.
     /// @param _gas_left     The amount of gas left.
+    /// @param _gas_refund   The amount of refunded gas.
     /// @param _output_data  The pointer to the output.
     /// @param _output_size  The output size.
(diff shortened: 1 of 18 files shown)


include/evmc/evmc.hpp
```diff
@@ -345,33 +345,39 @@ class Result : private evmc_result
     ///
     /// @param _status_code  The status code.
     /// @param _gas_left     The amount of gas left.
+    /// @param _gas_refund   The amount of refunded gas.
```
axic, line 348: Does `gas_left` includes the `gas_refund` value?

chfast: No. It is mentioned somewhere that gas refund is uncapped. I can try to improve the documentation.

This is implementable, but I decided to leave the transaction-specific logic outside of EVMC yet.

PR #673, opened 2022-11-09
Add wording about output_data in case of output_size=0

axic:


include/evmc/evmc.h
```diff
@@ -419,6 +419,8 @@ struct evmc_result
      *  freed with evmc_result::release().
      *
      *  This MAY be NULL.
```
axic, line 421: Why is this, and some other places double indented, but not every block?

axic:
Oh no, the doxygen link in CI is 404, hence the failing test.

include/evmc/evmc.h
chfast: Because of migration from `///` comments to `/**` long time ago.

include/evmc/evmc.h
```diff
@@ -419,6 +419,8 @@ struct evmc_result
...
+     *
+     * If output_size is NULL this pointer MAY NOT be dereferenced.
```
chfast, line 423:
```suggestion
     * If result::output_size is 0 this pointer MUST NOT be dereferenced.
```
and combine with the paragraph above.

include/evmc/evmc.h
axic: Fixed.

chfast (approved):

axic:
New build here: https://output.circle-artifacts.com/output/job/49b296b4-94bd-44c8-a556-f51800483554/artifacts/0/docs/index.html

Screenshot

PR #674, opened 2022-11-09
ci: Update doxygen to 1.9.5 and pin MinGW

axic:


axic:
This should fix the `lint` job, but the `go-windows` job still fails. The last successful one used MinGW 11.2, and current 12.2 is installed. It seems to be a recent problem (see [this](https://forum.golangbridge.org/t/running-gcc-failed-exit-status-1/29465)). We may want to pin an older one or wait for some upstream fixes?

chfast (approved):
CI generated docs: https://output.circle-artifacts.com/output/job/7ea229c4-6b92-429d-aacb-2aca28f2359b/artifacts/0/docs/index.html

PR #677, opened 2022-12-13
fix rust bindings compile failed on aarch64 linux

bxq2011hust:


bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -353,8 +353,8 @@ fn build_create_fn(names: &VMNameSet) -> proc_macro2::TokenStream {
                 execute: Some(__evmc_execute),
                 get_capabilities: Some(__evmc_get_capabilities),
                 set_option: Some(__evmc_set_option),
-                name: unsafe { ::std::ffi::CStr::from_bytes_with_nul_unchecked(#static_name_ident.as_bytes()).as_ptr() as *const i8 },
-                version: unsafe { ::std::ffi::CStr::from_bytes_with_nul_unchecked(#static_version_ident.as_bytes()).as_ptr() as *const i8 },
+                name: unsafe { ::std::ffi::CStr::from_bytes_with_nul_unchecked(#static_name_ident.as_bytes()).as_ptr() as *const ::libc::c_char },
+                version: unsafe { ::std::ffi::CStr::from_bytes_with_nul_unchecked(#static_version_ident.as_bytes()).as_ptr() as *const ::libc::c_char },
```
axic, line 357: Looking at bindgen's output of `evmc_vm`:
```
pub struct evmc_vm {
    #[doc = " EVMC ABI version implemented by the VM instance."]
    #[doc = ""]
    #[doc = " Can be used to detect ABI incompatibilities."]
    #[doc = " The EVMC ABI version represented by this file is in ::EVMC_ABI_VERSION."]
    pub abi_version: ::std::os::raw::c_int,
    #[doc = " The name of the EVMC VM implementation."]
    #[doc = ""]
    #[doc = " It MUST be a NULL-terminated not empty string."]
    #[doc = " The content MUST be UTF-8 encoded (this implies ASCII encoding is also allowed)."]
    pub name: *const ::std::os::raw::c_char,
    #[doc = " The version of the EVMC VM implementation, e.g. \"1.2.3b4\"."]
    #[doc = ""]
    #[doc = " It MUST be a NULL-terminated not empty string."]
    #[doc = " The content MUST be UTF-8 encoded (this implies ASCII encoding is also allowed)."]
    pub version: *const ::std::os::raw::c_char,
    ...
```

Not sure why we were casting to `i8` in the first place, maybe during a recent `bindgen` bump this changed but we didn't notice.

Also in [`std:ffi`](https://doc.rust-lang.org/nightly/core/ffi/struct.CStr.html#method.as_ptr) we use `as_ptr()` already returns `i8` so no need to cast to that.

In [`core:ffi`](https://doc.rust-lang.org/nightly/std/ffi/struct.CStr.html#method.as_ptr) `as_ptr()` already returns a `c_char`, no need to cast it to that. I'll just create a PR for this one.

axic: Can you check if #678 fixes the problem for you?

chfast: I think this is famous C thing: `char` is signed on x86/linux but unsigned on arm/linux. https://godbolt.org/z/vx5YGKqnW
We can add arm64 CI build if it is really needed...

bxq2011hust: > Can you check if #678 fixes the problem for you?

Ok, I will check it later

bxq2011hust: > I think this is famous C thing: `char` is signed on x86/linux but unsigned on arm/linux. https://godbolt.org/z/vx5YGKqnW We can add arm64 CI build if it is really needed...

Yes, it is the signed and unsigned problem

bxq2011hust: > Can you check if #678 fixes the problem for you?

yes, it fixed

PR #679, opened 2023-01-03
Migrate Windows CI to CircleCI

### LINKED ISSUE #510: Migrate Appveyor to CircleCI (Windows)
See https://circleci.com/build-environments/windows/

axic:
Closes #510.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -300,6 +300,74 @@ jobs:
       - build
       - test
 
+  build-windows:
+    executor: win/server-2022
+    environment:
+      CMAKE_BUILD_TYPE: Release
+    steps:
+      - checkout
+      - run:
+          name: "Setup environment (bash)"
+          shell: bash
+          command: |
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -1,39 +0,0 @@
-version: "{build}"
```
chfast, line 1: Don't delete it.

circle.yml
```diff
@@ -426,6 +460,7 @@ workflows:
       - build-gcc-32bit
       - build-cmake-min
       - build-macos
+      - build-windows
```
chfast, line 463: The AppVeyor CI had 4 windows builds. We don't have to keep all of them, but having 32-bit build may be useful.

chfast (dismissed):

circle.yml
axic: Yeah, the 32-bit can be easily added back I suppose with replacing `-Arch amd64` with `-Arch x86`.

Not sure if circleci has images with the 3 older visual studio versions.

chfast (approved):

PR #681, opened 2023-02-06
Add CREATE3 message support

gumb0:


include/evmc/evmc.h
```diff
@@ -187,6 +188,8 @@ struct evmc_message
      * Defined as `c` in the Yellow Paper.
      */
     evmc_address code_address;
+
+    uint8_t initcontainer_index;
```
axic, line 192: Lacks documentation.

include/evmc/evmc.h
```diff
@@ -187,6 +188,20 @@ struct evmc_message
      * Defined as `c` in the Yellow Paper.
      */
     evmc_address code_address;
+
+    /**
+     * The init code for creation message.
+     *
+     * Ignored unless kind is ::EVMC_CREATE3.
+     */
+    const uint8_t* init_code;
```
gumb0, line 197: EOF creation requires both `init_code` and `input_data` for execution.

Alternative approach to this could be new method in Host, that will take `evmc_message` and `init_code` as parameters.

gumb0: Yet another option could be passing initcontainer index instead of full init_code, but then Host would need to get parent container from state and parse it to get initcontainer from it. Seems inefficient.

include/evmc/evmc.h
```diff
@@ -187,6 +188,16 @@ struct evmc_message
...
      */
     evmc_address code_address;
+
+    /**
+     * The code to be executed.
+     */
+    const uint8_t* code;
+
+    /**
+     * The code to be executed.
```
chfast, line 198:
```suggestion
     * The length of the code to be executed.
```

chfast (approved):

PR #682, opened 2023-02-13
C++: Support zero literal (0x0_address)

axic:


include/evmc/evmc.hpp
```diff
@@ -289,7 +289,8 @@ namespace literals
 template <typename T>
 constexpr T parse(std::string_view s) noexcept
 {
-    return from_hex<T>(s).value();
+    auto const v = from_hex<T>(s);
```
axic, line 292: The failure case is also used for parsing errors, including odd nibbles.

Need to rethink.

test/unittests/cpp_test.cpp
```diff
@@ -149,6 +149,12 @@ TEST(cpp, std_hash)
     std::fill_n(eb.bytes, sizeof(eb), uint8_t{0xee});
     EXPECT_EQ(std::hash<evmc::bytes32>{}(eb), static_cast<size_t>(0xbb14e5c56b477375));
 
+    const auto zero_address = 0x0_address;
+    EXPECT_EQ(zero_address, evmc::address{});
```
gumb0, line 153: These can be static_asserts I think.

PR #685, opened 2023-02-24
cmake: Update compiler settings

chfast:


axic (approved):
It's draft but seems to pass tests?

chfast:
> It's draft but seems to pass tests?

Warnings are not treated as errors by default any more. So I will need to change this. The PR was to check if Cable update is not completely broken.

PR #688, opened 2023-04-19
Update CHANGELOG and README

chfast:


yperbasis (approved):

CHANGELOG.md
```diff
@@ -5,12 +5,24 @@ Documentation of all notable changes to the **EVMC** project.
...
 - The Prague EVM revision (anticipated after Cancun)
   [#683](https://github.com/ethereum/evmc/pull/683)
+- C++: Make `evmc::Result` accessible via `evmc_result` reference
+  [#686](https://github.com/ethereum/evmc/pull/686)
+
+### Changed
+
+- Set Shanghai as the latest stable revision `EVMC_LATEST_STABLE_REVISION`.
+  [#687](https://github.com/ethereum/evmc/pull/687)
+- Bump minimum supported compiler versions:
```
axic, line 21: I wonder actually because of these two changes (default + compiler reqs) this may be a breaking release :(

chfast: I don't think so. The ABI is unchanged.

axic: But behaviour is different to what the user expects. Upgrading the non-breaking version may break compilation, and may break execution if they rely on `LATEST_STABLE`.

chfast: Previous compilers (except CMake bump) work it is just impractical to test them.

The opposite is worse: e.g. solidity will need to upgrade EVMC to be able to use evmone with EVMC 11 although there is no ABI break in EVMC 11.

chfast: From evmc.ethereum.org:

> The releases with MINOR version change allow adding new API features and modifying the language bindings API. Backward incompatible API changes are allowed but should be avoided if possible.

axic: I'm not sure what is the best, but the orthodox view is changing `LATEST_STABLE` is breaking.

Very likely no real user of EVMC depends on that, Solidity doesn't, @yperbasis does silkworm depend on it?

Assuming Silkworm is fine, I'm fine going with non-breaking.

yperbasis: Silkworm does not depend on `LATEST_STABLE`.

PR #690, opened 2023-05-05
Add refund to execution result in Go binding

HerbertJordan:
This PR modifies the `Execute()` function from the Go bindings to return a `Result` struct instead of a list of result values to support the retrieval of the amount of refunded gas. This information is available in [evmc.h](https://github.com/ethereum/evmc/blob/v10.1.0/include/evmc/evmc.h)'s [evmc_result](https://github.com/ethereum/evmc/blob/v10.1.0/include/evmc/evmc.h#L404-L410), but was not exposed in the Go binding so far.

### DIFF
--- a/bindings/go/evmc/evmc.go
+++ b/bindings/go/evmc/evmc.go
@@ -195,12 +195,18 @@ func (vm *VM) SetOption(name string, value string) (err error) {
 	return err
 }
 
-func (vm *VM) Execute(ctx HostContext, rev Revision,
-	kind CallKind, static bool, depth int, gas int64,
-	recipient Address, sender Address, input []byte, value Hash,
-	code []byte) (output []byte, gasLeft int64, err error) {
-	result, err := vm.Execute2(ctx, rev, kind, static, depth, gas, recipient, sender, input, value, code)
-	return result.Output, result.GasLeft, err
+type Parameters struct {
+	Context   HostContext
+	Revision  Revision
+	Kind      CallKind
+	Static    bool
+	Depth     int
+	Gas       int64
+	Recipient Address
+	Sender    Address
+	Input     []byte
+	Value     Hash
+	Code      []byte
 }
 
 type Result struct {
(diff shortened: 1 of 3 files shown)


chfast (commented):
Hi @HerbertJordan, thanks for the contribution.

I'd rather change the existing API of `Execute()` to return `Result` as you proposed. What do you think?

HerbertJordan:
Hi @chfast, thanks for reviewing this PR.

If backward compatibility is not an issue, we should indeed update the existing `Execute()` function. In that case, may I also suggest replacing the parameter list with a parameter struct for the same reasons as for the result (see updated PR)?

bindings/go/evmc/evmc.go
```diff
@@ -195,29 +195,48 @@ func (vm *VM) SetOption(name string, value string) (err error) {
...
+	Code      []byte
+}
+
+type Result struct {
+	Output    []byte
+	GasLeft   int64
+	GasRefund int64
+}
+
+func (vm *VM) Execute(params Parameters) (res Result, err error) {
```
chfast, line 218: Although it may be more readable to use additional `Parameters` type, I'd restrain from adding it in this PR. Binding for EVMC are not so much about having beautiful Go code.

HerbertJordan: The main motivation for this change was less the readability but to make the interface future-proof for further modifications. Using a struct as a parameter list would enable additional (optional) parameters to be added in the future without breaking compatibility with existing code using the function -- analog to the argument why the result type is a struct.

chfast: Ok, I get it. Although I'm not sure this is good idea. There is a chance then that a modified API will compile for users but behave differently or incorrectly. Wouldn't you prefer a compiler error so that you can check what has changed in EVMC?

I'm not saying no to this, but it should go to separate PR anyway :)

HerbertJordan: It would only work for optional parameters with sensible defaults -- things like execution hints. Maybe only those should be grouped into a parameter struct.

For now, I removed the parameter struct again.

chfast (approved):

PR #695, opened 2023-09-02
cmake: Fix evmc_add_vm_test() not working in cross-compilation

chfast:
In add_test() use the `evmc::evmc-vmtester` CMake target name directly to make it work with `CROSSCOMPILING_EMULATOR`.


chfast:
Not sure if I should release this as EVMC 10.2. The `master` is already 11.x so we cannot use it in evmone without breaking ABI.

rodiazet (dismissed):

PR #699, opened 2023-11-06
Lookup CREATE4 tx initcodes by hash

pdobacz:
All changes under `bindings` still todo

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -170,8 +170,8 @@ struct evmc_message
     /**
      * The optional value used in new contract address construction.
      *
-     * Needed only for a Host to calculate created address when kind is ::EVMC_CREATE2.
-     * Ignored in evmc_execute_fn().
+     * Needed only for a Host to calculate created address when kind is ::EVMC_CREATE2,
+     * ::EVMC_CREATE3 or ::EVMC_CREATE4. Ignored in evmc_execute_fn().
      */
     evmc_bytes32 create2_salt;
 
@@ -181,7 +181,7 @@ struct evmc_message
      * For ::EVMC_CALLCODE or ::EVMC_DELEGATECALL this may be different from
      * the evmc_message::recipient.
      * Not required when invoking evmc_execute_fn(), only when invoking evmc_call_fn().
-     * Ignored if kind is ::EVMC_CREATE or ::EVMC_CREATE2.
+     * Ignored if kind is ::EVMC_CREATE, ::EVMC_CREATE2, ::EVMC_CREATE3 or ::EVMC_CREATE4.
      *
      * In case of ::EVMC_CAPABILITY_PRECOMPILES implementation, this fields should be inspected
      * to identify the requested precompile.
@@ -193,14 +193,14 @@ struct evmc_message
     /**
      * The init code for creation message.
      *
-     * Ignored unless kind is ::EVMC_CREATE3.
+     * Ignored unless kind is ::EVMC_CREATE3 or ::EVMC_CREATE4.
      */
     const uint8_t* init_code;
 
     /**
      * The size of the init code for creation message.
      *
-     * Ignored unless kind is ::EVMC_CREATE3.
+     * Ignored unless kind is ::EVMC_CREATE3 or ::EVMC_CREATE4.
      */
     size_t init_code_size;
 };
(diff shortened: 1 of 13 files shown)


include/evmc/evmc.h
```diff
@@ -170,8 +170,8 @@ struct evmc_message
     /**
      * The optional value used in new contract address construction.
      *
-     * Needed only for a Host to calculate created address when kind is ::EVMC_CREATE2.
-     * Ignored in evmc_execute_fn().
+     * Needed only for a Host to calculate created address when kind is ::EVMC_CREATE2,
+     * ::EVMC_CREATE3 or ::EVMC_CREATE4. Ignored in evmc_execute_fn().
```
gumb0, line 174: From the point of view of EVM-Host interaction CREATE3 and CREATE4 executions are the same, so I thought to cover both with a singe kind `EVMC_CREATE3`

pdobacz: OK, finally got to this. `EVMC_CREATE4` was introduced earlier, when `CREATE4` recursive validation was turned on.

At the moment we do hard failure (consume all gas) on eof validation failure (tentatively to be changed to light failure, but this is still pending discussion and decision). Hard failure is done more naturally in the host and requires `EVMC_CREATE4`. We need to finally decide on the light failure change and then I'll revert this addition.

But it can be done regardless of the lookup by hash change in the target `create4` branch.

pdobacz:
The CI failures are no worse than on the target branch. The bindings work is also not complete on the target branch so I take it it is ok to go there as is.

include/evmc/evmc.h
```diff
@@ -221,9 +221,13 @@ struct evmc_tx_context
...
     size_t blob_hashes_count;         /**< The number of blob hashes (EIP-4844). */
-    const uint8_t* const* initcodes;  /**< The array of initcode data (CREATE4). */
-    const size_t* initcode_sizes;     /**< The array of initcode sizes (CREATE4). */
-    size_t initcodes_count;           /**< The number of initcodes (CREATE4). */
+};
+
+/** Return value of the get initcode by hash query (CREATE4). */
+struct evmc_tx_initcode
+{
+    const uint8_t* data; /**< Initcode. Nullptr means no initcode with such hash given. */
```
gumb0, line 229:
```suggestion
    const uint8_t* data; /**< Initcode. NULL means no initcode with such hash given. */
```

include/evmc/evmc.h
```diff
@@ -244,6 +248,20 @@ struct evmc_host_context;
  */
 typedef struct evmc_tx_context (*evmc_get_tx_context_fn)(struct evmc_host_context* context);
 
+/**
+ * Get transaction initcode by hash callback function.
+ *
+ *  This callback function is used by an EVM to retrieve an initcode to
+ *  be supplied to a CREATE4 function. Should return result with nullptr in case there is no
```
gumb0, line 255:
```suggestion
 *  be supplied to a CREATE4 function. Should return result with NULL in case there is no
```

include/evmc/evmc.h
```diff
@@ -221,9 +221,13 @@ struct evmc_tx_context
     evmc_uint256be blob_base_fee;     /**< The blob base fee (EIP-7516). */
     const evmc_bytes32* blob_hashes;  /**< The array of blob hashes (EIP-4844). */
     size_t blob_hashes_count;         /**< The number of blob hashes (EIP-4844). */
-    const uint8_t* const* initcodes;  /**< The array of initcode data (CREATE4). */
-    const size_t* initcode_sizes;     /**< The array of initcode sizes (CREATE4). */
-    size_t initcodes_count;           /**< The number of initcodes (CREATE4). */
+};
+
+/** Return value of the get initcode by hash query (CREATE4). */
+struct evmc_tx_initcode
```
gumb0, line 227: I wonder if it makes sense to have this as more generic-named `evmc_bytes`. But fine to have it like this for now.

gumb0:
I'm fine with merging it into `create4` if needed. If you do, then make a PR for `create4` branch.

I think if we get rid of `EVMC_CREATE4` it all could just go into `create3` branch mayve.

gumb0:
Alternative design idea could be: new `initcode_hash` member in `evmc_message` that would be used only with `EVMC_CREATE4` instead of `evmc_message::init_code`. Then the hash is passed in `execute()` and no need to have `evmc_get_tx_initcode_by_hash()`, and `Host` can skip hashing the initcode one more time when calculating the address (in `compute_new_account_create3_address()`).

I'm not sure yet if it's worth it.

pdobacz:
> Alternative design idea could be: new `initcode_hash` member in `evmc_message` that would be used only with `EVMC_CREATE4` instead of `evmc_message::init_code`. Then the hash is passed in `execute()` and no need to have `evmc_get_tx_initcode_by_hash()`, and `Host` can skip hashing the initcode one more time when calculating the address (in `compute_new_account_create3_address()`).
> 
> I'm not sure yet if it's worth it.

I don't see how would this remove the need for `get_tx_initcode_by_hash`. It is called in `create_eof_impl` to single out a particular initcode based on a stack argument.

I agree that `initcode_hash` can be added to `evmc_message` to skip hashing in `compute_new_account_create3_address`. The downside is bloating `evmc_message` and coupling of the two hashing rules (bad in case we want to have 2 different ones for these 2 purposes). Hard to decide.

Does `CREATE4` gas cover for the second hashing (being the hashing of all initcodes delivered)? Because CREATE3/4 cover for the "inner" hashing done to compute the new address in their own `initcode` cost 8 gas/word, correct?

pdobacz:
> I don't see how would this remove the need for get_tx_initcode_by_hash. It is called in create_eof_impl to single out a particular initcode based on a stack argument.

ah ok, I see now, host has access to the entire tx, so it can pick the initcode out of it by itself. Hm... the failure modes would get trickier then, no? since this change would cause us to make a full failure on an non-existent tx_initcode hash?

PR #700, opened 2023-12-11
Simplify CREATE4, tidy failure conditions

pdobacz:
Removes the distinction between `EVMC_CREATE3` and `4`, as it is not needed anymore (see `evmone` PR referenced below). Name of the message kind pending.

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -80,8 +80,7 @@ enum evmc_call_kind
     EVMC_CALLCODE = 2,     /**< Request CALLCODE. */
     EVMC_CREATE = 3,       /**< Request CREATE. */
     EVMC_CREATE2 = 4,      /**< Request CREATE2. Valid since Constantinople.*/
-    EVMC_CREATE3 = 5,      /**< Request CREATE3. Valid since Prague.*/
-    EVMC_CREATE4 = 6       /**< Request CREATE4. Valid since Prague.*/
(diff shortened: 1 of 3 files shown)


chfast (commented):
LGTM just not 100% sure about the name.

pdobacz:
> LGTM just not 100% sure about the name.

yea, the name is a bit goofy, and won't last if we rename CREATE3/4 to something sensible (if it is at this point at all possible). Any ideas for alternatives?

chfast:
I'm fine with keeping it as is because not for `master` yet.

include/evmc/evmc.h
```diff
@@ -80,8 +80,7 @@ enum evmc_call_kind
     EVMC_CALLCODE = 2,     /**< Request CALLCODE. */
     EVMC_CREATE = 3,       /**< Request CREATE. */
     EVMC_CREATE2 = 4,      /**< Request CREATE2. Valid since Constantinople.*/
-    EVMC_CREATE3 = 5,      /**< Request CREATE3. Valid since Prague.*/
-    EVMC_CREATE4 = 6       /**< Request CREATE4. Valid since Prague.*/
+    EVMC_CREATE34 = 5      /**< Request CREATE3 or CREATE4. Valid since Prague.*/
```
gumb0, line 83: Maybe `EVMC_CREATE3_4`?

pdobacz: indeed, it is a bit better. If no one objects, I'll change to that

PR #702, opened 2024-01-11
Initcode transaction (TXCREATE) support

gumb0:
Requires https://github.com/ethereum/evmc/pull/681

### DIFF
--- a/examples/example_host.cpp
+++ b/examples/example_host.cpp
@@ -4,6 +4,7 @@
 
 /// @file
 /// Example implementation of an EVMC Host.
+/// TODO: Doesn't support TXCREATE transaction initcodes.
 
 #include "example_host.h"
 
@@ -139,6 +140,11 @@ public:
     }
 
     evmc_tx_context get_tx_context() const noexcept final { return tx_context; }
+
+    evmc_tx_initcode get_tx_initcode_by_hash(const evmc_bytes32& /* hash */) const noexcept final
+    {
+        return {nullptr, 0};
+    }
 
     // NOLINTNEXTLINE(bugprone-exception-escape)
     evmc::bytes32 get_block_hash(int64_t number) const noexcept final
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -169,8 +169,8 @@ struct evmc_message
     /**
      * The optional value used in new contract address construction.
      *
-     * Needed only for a Host to calculate created address when kind is ::EVMC_CREATE2.
-     * Ignored in evmc_execute_fn().
+     * Needed only for a Host to calculate created address when kind is ::EVMC_CREATE2,
+     * ::EVMC_EOFCREATE. Ignored in evmc_execute_fn().
      */
     evmc_bytes32 create2_salt;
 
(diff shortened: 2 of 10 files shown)


include/evmc/evmc.h
```diff
@@ -880,6 +901,9 @@ struct evmc_host_interface
 
     /** Set transient storage callback function. */
     evmc_set_transient_storage_fn set_transient_storage;
+
+    /** Get transaction initcode by hash callback function. */
+    evmc_get_tx_initcode_by_hash_fn get_tx_initcode_by_hash;
```
gumb0, line 906: Does it make sense to put it close to `get_tx_context` ?

pdobacz: as in, move it up/down in the file? I think so. I can't remember exactly why I positioned stuff where I did anymore.

examples/example_host.cpp
```diff
@@ -4,6 +4,7 @@
 
 /// @file
 /// Example implementation of an EVMC Host.
+/// TODO: Doesn't support TXCREATE transaction initcodes.
```
gumb0, line 7: Not sure it makes sense to support it, `example.c` only sends `EVMC_CALL` message anyway.

pdobacz: feel free to drop the comment like this, if you feel like nothing needs to be done. I was very generous when adding them in

pdobacz (approved):

chfast (commented):
I don't want to spoil the party, but I think the simpler API would be to extend the `evmc_tx_context` with

```cpp
hashed_initcode* initcodes;
size_t initcodes_count;
```

where 

```cpp
struct hashed_initcode {
    evmc_bytes32 hash;
    const uint8_t* code;
    size_t code_size;
};
```

pdobacz:
> I don't want to spoil the party, but I think the simpler API would be to extend the `evmc_tx_context` with

Indeed simpler but makes problematic:
1. Optimizing initcode lookup (unless we require the initcodes are sorted I suppose)
2. Any usecases where the initcode is fetched not from the TX, for which the fetch-by-hash was introduced in the first place

Neither is maybe big deal but sth to keep in mind.

gumb0:
> Indeed simpler but makes problematic:
> 
> 1. Optimizing initcode lookup (unless we require the initcodes are sorted I suppose)

I don't think this variant changes much, it only changes where lookup by hash is happening.
-  With current API, Host gets initcodes array from transaction and hashes them and puts them into `unordered_map` to later serve `get_tx_initcode_by_hash()` requests. 
- With this new API, Host will get initcode array from transaction, hash them once and serve as array in `tx_context`, then EVM will get them and put into `unordered_map` on EVM side (in evmone it will be in `ExecutionState`).

> 2. Any usecases where the initcode is fetched not from the TX, for which the fetch-by-hash was introduced in the first place

For something other than transaction, it will be some other different API? But I think fetch-by-hash is mainly for easier aggregating several transactions (or AA-transactions) into one batch transaction (merging their initcodes without changing TXCREATE instances).

gumb0:
Anyway I plan to try this in a different PR.

pdobacz:
Makes sense!

chfast:
The other approach can be implemented as a "flat map": vector of already sorted (hash, initcode) pairs. EVM only need to read it and can lookup values by binary search.

But I'm not insisting on sorting values right now. Linear search should be fine as the start.

gumb0:
> The other approach can be implemented as a "flat map": vector of already sorted (hash, initcode) pairs. EVM only need to read it and can lookup values by binary search.

https://github.com/ethereum/evmone/pull/702/commits/55a7102c2f4f13ccf372b597ad4b3f032002ef86
EVM puts them into `unordered_map`.

gumb0:
Replaced by #709

PR #705, opened 2024-02-29
Move code reference to message

chfast:


include/evmc/evmc.hpp
```diff
@@ -745,13 +739,9 @@ class VM
     /// but without providing the Host context and interface.
```
gumb0, line 739: Update comment above

bindings/rust/evmc-declare/src/lib.rs
```diff
@@ -395,52 +395,34 @@ fn build_execute_fn(names: &VMNameSet) -> proc_macro2::TokenStream {
...
             revision: ::evmc_vm::ffi::evmc_revision,
-            msg: *const ::evmc_vm::ffi::evmc_message,
-            code: *const u8,
-            code_size: usize
+            msg: *const ::evmc_vm::ffi::evmc_message
         ) -> ::evmc_vm::ffi::evmc_result
         {
             use evmc_vm::EvmcVm;
 
-            // TODO: context is optional in case of the "precompiles" capability
```
gumb0, line 405: why is this removed?

chfast: I wasn't able to translate the check easily and also not sure if this check isn't to heavy to be included in the bindings (unless this helps Rust translation somehow).

gumb0: Maybe `TODO` is worth to leave? It seems to point to something that is not yet implemented in bindings, but might be useful.

axic: I don't think the check for instance/msg null should be removed. This is the place it ensures it won't be null.

PR #707, opened 2024-03-02
Fix rust bindings generation

iamyulong:
# Description

This PR fixes two issues with the rust bindings:
- Constant `EMVC_ABI_VERSION` is not exported, caused by an outdated `bindgen`
- Enums are not properly rustified, due to invalid configuration.

# Testing

I've run `cargo test` for the whole workspace and `cargo build` for the `example-rust-vm`.


chfast:
Oh great. This explains why I'm also seeing this problems locally...

chfast (approved):

PR #709, opened 2024-03-18
Add TXCREATE transaction initcodes support - new approach

gumb0:
Replaces #702

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -169,7 +169,8 @@ struct evmc_message
     /**
      * The optional value used in new contract address construction.
      *
-     * Needed only for a Host to calculate created address when kind is ::EVMC_CREATE2.
+     * Needed only for a Host to calculate created address when kind is ::EVMC_CREATE2 or
+     * ::EVMC_EOFCREATE.
      * Ignored in evmc_execute_fn().
      */
     evmc_bytes32 create2_salt;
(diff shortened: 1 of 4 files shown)


bindings/go/evmc/evmc.go
```diff
@@ -288,3 +288,22 @@ func bytesPtr(bytes []byte) *C.uint8_t {
...
+        return nil
+    }
+
+	out := make([]C.evmc_tx_initcode, len(initcodes))
+	for i := 0; i < len(initcodes); i++ {
+		out[i].hash = evmcBytes32(initcodes[i].hash)
+		out[i].code = bytesPtr(initcodes[i].code)
+		out[i].code_size = C.size_t(len(initcodes[i].code))
+	}
+	return &out[0]
```
chfast, line 308: Are you sure you can pass this GC pointer to C?

And the formatting is wrong.

chfast: I think it make sense to add a Go unit test where we actually pass some initcodes. Go has a runtime inspection of passed pointers and will likely report errors if we are doing something wrong.

gumb0: I think it's definitely wrong, but digging deeper into this I am now confused, because when returning output data from Go host `call` to C side it does something similar:
https://github.com/ethereum/evmc/blob/954944a4286a9e3758c80c05acba7f6a6122b4bb/bindings/go/evmc/host.go#L233
Go byte array is converted to C pointer and returned in `evmc_result`. This seems wrong, because Go array is going to be garbage-collected, and then C has a dangling poiner.

chfast: There the `C.evmc_make_result` makes a copy.

gumb0: Ah I see, missed that, thanks.

chfast: In general, it looks you should pin the array `make([]C.evmc_tx_initcode, len(initcodes))` somewhere. Maybe add a private field `cInitcodes` to `TxContext`.

include/evmc/evmc.h
```diff
@@ -169,7 +169,8 @@ struct evmc_message
     /**
      * The optional value used in new contract address construction.
      *
-     * Needed only for a Host to calculate created address when kind is ::EVMC_CREATE2.
+     * Needed only for a Host to calculate created address when kind is ::EVMC_CREATE2,
```
chfast, line 172:
```suggestion
     * Needed only for a Host to calculate created address when kind is ::EVMC_CREATE2 or
```

chfast (approved):
Add commit description and mention EOF.

PR #713, opened 2024-07-25
Fix support for EOF creation tx in evmc run

pdobacz:
This PR will be accompanied by a PR in `evmone`, fixing the handling of top level creation msgs when validation EOF.

Here, we only replicate the behavior of creation txs for EOF - if the code begins with EOF magic it will be interpreted as an EOF creation transaction and the msg kind will be set appropriately, in order to be consumed by the `vm.execute`

### DIFF
--- a/lib/tooling/run.cpp
+++ b/lib/tooling/run.cpp
@@ -19,6 +19,9 @@ constexpr auto create_address = 0xc9ea7ed000000000000000000000000000000001_addre
 /// The gas limit for contract creation.
 constexpr auto create_gas = 10'000'000;
 
+/// MAGIC bytes denoting an EOF container.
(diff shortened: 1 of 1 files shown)


pdobacz:
I needed to do two tweaks to the CI for this to work:

- bump `build-macos` - not sure what I'm doing here, just best-guess, deprecation announcement [here](https://discuss.circleci.com/t/macos-resource-deprecation-update/46891)
- downgrade VS 2022 on appveyor - MSVC 1940 (17.40.x) is failing for a reason beyond my comprehension, using "Previous..." image as I got a hint from [some old post](https://www.appveyor.com/updates/2023/04/24/) puts it back at 17.39.x which worked. I can put in an issue to un-downgrade it lest we forget to

chfast (approved):

PR #728, opened 2025-02-07
Add EXPERIMENTAL revision

chfast:
Add new `EVMC_EXPERIMENTAL` to `evmc_revision` to allow EVM implementations to expose experimental features.

### DIFF
--- a/include/evmc/evmc.h
+++ b/include/evmc/evmc.h
@@ -1034,21 +1034,27 @@ enum evmc_revision
     EVMC_CANCUN = 12,
 
     /**
-     * The Prague revision.
+     * The Prague / Pectra revision.
      *
-     * The future next revision after Cancun.
+     * https://eips.ethereum.org/EIPS/eip-7600
      */
     EVMC_PRAGUE = 13,
 
     /**
-     * The Osaka revision.
+     * The Osaka / Fusaka revision.
      *
-     * The future next revision after Prague.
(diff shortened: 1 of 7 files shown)


include/evmc/evmc.h
```diff
@@ -1034,21 +1034,27 @@ enum evmc_revision
...
-     * The future next revision after Prague.
+     * https://eips.ethereum.org/EIPS/eip-7607
      */
     EVMC_OSAKA = 14,
 
+    /**
+     * The unspecified EVM revision used for EVM implementations to expose
+     * experimental features.
+     */
+    EVMC_EXPERIMENTAL = 15,
```
gumb0, line 1054: So the number is going to change with every added revision? Then if somebody has this hardcoded, e.g. in the command line to evmc-tool, it will break... Maybe not a huge concern.

gumb0: Yeah I guess it's mostly in tests, and it's not clear whether we would or would not want to include Experimental revision in them. Maybe rather we do want that, because experimental EVMs shouldn't break any existing functionality.

gumb0 (approved):

include/evmc/evmc.h
chfast: Maybe if we put the `EXPERIMENTAL` outside of the `MAX` this would work somehow... but let's see first how this version works.

